PR #22, opened 2020-11-11
REALMC-7148: Buildout auth commands

makesitgo:
A lot left to do here, but I want to get early eyes and hopefully and early merge here (with follow-up tickets) so that we can unblock a slew of work for others hopping on here


cmd/root.go
```diff
@@ -0,0 +1,67 @@
...
+		return cli.BaseCommand{}, cfgErr
+	}
+
+	return cli.BaseCommand{
+		Config:      cfg,
+		RealmClient: realm.NewClient(),
+	}, nil
+}
+
+func handleErr(err error) {
```
makesitgo, line 62: will be expanding on this through https://jira.mongodb.org/browse/REALMC-7340

internal/cli/login_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/10gen/realm-cli/internal/config"
+	"github.com/10gen/realm-cli/internal/mock"
+	"github.com/10gen/realm-cli/internal/realm"
+	u "github.com/10gen/realm-cli/internal/utils/test"
+	"github.com/10gen/realm-cli/internal/utils/test/it"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+func TestLogin(t *testing.T) {
+	t.Run("Setup should TODO", func(t *testing.T) {
```
makesitgo, line 18: going to do this here (soon), no JIRA needed for now

internal/cli/err/errors.go
```diff
@@ -64,7 +64,7 @@ func (err PrivilegedErr) Error() string {
 
 func findRootCause(err error) error {
```
tsedgwick, line 65: I believe this can be replaced with `errors.As` (https://golang.org/pkg/errors/#As)

internal/cli/err/errors.go
```diff
@@ -1,34 +1,34 @@
-package cli
+package clierr
```
tsedgwick, line 1: I know this is a common pattern in the `baas` repo, but it becomes hard to follow since the package normally is the name of the package. Any reason we can't have this in the `cli` package itself?

internal/config/config.go
```diff
@@ -0,0 +1,96 @@
+package config
+
+import (
+	"fmt"
+
+	clierr "github.com/10gen/realm-cli/internal/cli/err"
```
tsedgwick, line 6: to my earlier comment about this package - having to prefix your imports is typically antipattern in go and would like to avoid it if possible in the cli.

internal/cli/cli.go
```diff
@@ -0,0 +1,88 @@
...
+	if c, ok := cmd.(CommandPreparer); ok {
+		preRun = c.Setup
+	}
+
+	var postRun cmdFn
+	if c, ok := cmd.(CommandResponder); ok {
+		postRun = c.Feedback
+	}
+
+	return &cobra.Command{
```
tsedgwick, line 53: [mega-nit] would just create the top level cobra.Command then add to it so we can avoid nil checks further down:
```	
c := cobra.Command{
		Use:     use,
		Short:   short,
		RunE:    cobraFnE(cmd.Handler),
}

	if cp, ok := cmd.(CommandPreparer); ok {
		c.PreRunE: cobraFnE(cp.Setup),
	}

	if cr, ok := cmd.(CommandResponder); ok {
		c.PostRun = cr.Feedback
	}
return &c
```

cmd/root.go
```diff
@@ -0,0 +1,60 @@
...
+
+	rootCmd.PersistentFlags().StringVarP(&baseCmd.Config.Name, flags.Profile, flags.ProfileShort, config.DefaultName, usage.Profile)
+	rootCmd.PersistentFlags().BoolVar(&baseCmd.UI.DisableColors, flags.DisableColors, false, usage.DisableColors)
+	rootCmd.PersistentFlags().StringVar(&baseCmd.UI.OutputFormat, flags.OutputFormat, ui.DefaultOutputFormat, usage.OutputFormat)
+	rootCmd.PersistentFlags().StringVar(&baseCmd.UI.OutputTarget, flags.OutputTarget, "", usage.OutputTarget)
+	rootCmd.PersistentFlags().StringVar(&baseCmd.RealmConfig.BaseURL, flags.RealmBaseURL, realm.DefaultBaseURL, usage.RealmBaseURL)
+
+	rootCmd.AddCommand(cli.LoginCommand(baseCmd))
+	rootCmd.AddCommand(cli.LogoutCommand(baseCmd))
+	rootCmd.AddCommand(cli.WhoamiCommand(baseCmd))
```
tsedgwick, line 44: in my head, I had these flipped where baseCmd would wrap the underneath commands. That way, the underlying commands wouldn't need to reference the baseCmd at all as it would be a "middleware" of sorts.

tsedgwick (commented):
few callouts but nothing major. looking good so far.

internal/cli/cli.go
```diff
@@ -0,0 +1,88 @@
...
+		Use:     use,
+		Short:   short,
+		PreRunE: cobraFnE(preRun),
+		RunE:    cobraFnE(cmd.Handler),
+		PostRun: cobraFn(postRun),
```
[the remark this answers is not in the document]
makesitgo: So I went this route to intentionally limit the Feedback phase to be "error-free" (I don't want to report a "failed command" just because printing results failed, the server-side-effects likely still occurred).

I feel like that's something still worth preserving, over the benefit of simplicity/consistency here.  Thoughts?

internal/cli/err/errors.go
makesitgo: it couldn't be in the `cli` package due to import cycles... but I can move things around if this pattern is critical to remove.  but basically I wanted `Config.Load() error` to be able to return a `cli.Error`, and by forcing `Feedback` methods to be "error-free" (see comment above), the `ui` package wants access to `clieerr.HandleErr` to catch anything issues there

internal/cli/err/errors.go
```diff
@@ -1,34 +1,34 @@
...
+func NewPrivileged(message string, err error) PrivilegedErr {
+	return PrivilegedErr{NewWrapped(message, err)}
 }
 
-// WrappedErr is an error that can be unwrapped
-type WrappedErr interface {
+// ErrorWrapper is an error that can be unwrapped
+type ErrorWrapper interface {
 	Unwrap() error
 }
```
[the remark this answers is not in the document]
makesitgo: this interface _was_ necessary for its purpose... identifying a "cli error".  if/when we see one of these, the "root cause" of was it intended to be the "first non-cli error found in the wrapped chain".

considering we're working to get rid of the various wrapped errors, this likely goes away anyways.  but just wanted to call out the intended difference here b/w how the underlying go framework works

internal/cli/err/errors.go
makesitgo: again, there was an intentional nuance with this logic that the underlying go framework didn't outright support, and again, probably going away all together

internal/config/config.go
makesitgo: thoughts on a good name for a shared errors package then?  `commons.Error`?  I don't want to re-use `errors` with it in conflict with go's package.  open to ideas here, but I explained above the desire to keep it out of the `cli` package (so that `ui` can leverage it)

internal/cli/cli.go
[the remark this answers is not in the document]
makesitgo: Edit: using the `PostRunE` hook, and proper error handling for the feedback stage will be addressed in [REALMC-7340](https://jira.mongodb.org/browse/REALMC-7340)

internal/cli/cli.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+}
+
+// BaseCommand is the base command
```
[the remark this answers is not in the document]
makesitgo: This has now gone away :)

internal/cli/err/errors.go
makesitgo: CLI errors have been moved back to the `cli` package :)

internal/config/config.go
makesitgo: `clierr` is no more :)

cmd/root.go
makesitgo: we now have the `CommandFactory` building these, which inverts the control of a "base command".  I think we're good here now

internal/cli/cli.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	if responder, ok := cmdDef.Command.(CommandResponder); ok {
+		cmd.PostRunE = func(c *cobra.Command, a []string) error {
+			err := responder.Feedback(
+				factory.Profile,
+				newCobraUI(factory.Config.UI, c),
+			)
+			if err != nil {
+				return err // TODO(REALMC-XXXX): handle this error gracefully
```
tsedgwick, line 69: real ticket? 😬

tsedgwick (approved):
this is looking good 😎

internal/cli/cli.go
makesitgo: there were a few of these... I'm going to go through now and create them.  Just held off until your review, to make sure none of these TODOs were actual blockers for this merge

go.mod
```diff
@@ -3,10 +3,37 @@ module github.com/10gen/realm-cli
 go 1.14
 
 require (
+	4d63.com/gochecknoinits v0.0.0-20200108094044-eb73b47b9fc4 // indirect
+	dmitri.shuralyov.com/go/generated v0.0.0-20170818220700-b1254a446363 // indirect
+	github.com/AlecAivazis/survey/v2 v2.2.2
+	github.com/Netflix/go-expect v0.0.0-20180615182759-c93bf25de8e8
+	github.com/alecthomas/gocyclo v0.0.0-20150208221726-aa8f8b160214 // indirect
+	github.com/alexkohler/nakedret v1.0.0 // indirect
 	github.com/edaniels/golinters v0.0.3
```
kpatel71716, line 12: This all looks great, I am a fan of the framework built out here! One small thing we can do is update this dependency to use https://github.com/mongodb-forks/golinters . Before Eric left, he migrated his repos to this org

makesitgo: Nice catch!  Filed that to https://jira.mongodb.org/browse/REALMC-7430 for the time being

PR #24, opened 2020-11-21
REALMC-7341: Create integrated Realm client

makesitgo:


makesitgo:
Ignore the `Add MustBeNil helper` commit changes, I'll be dropping those in favor of #28

internal/cloud/realm/auth.go
```diff
@@ -0,0 +1,53 @@
...
+
+	var auth AuthResponse
+	var authErr error
+
+	dec := json.NewDecoder(res.Body)
+	defer func() {
+		if err := res.Body.Close(); err != nil {
+			authErr = err
+		}
+	}()
```
tsedgwick, line 46: should this really be an authErr? I would just `defer res.Body.Close()`

internal/cloud/realm/auth.go
```diff
@@ -0,0 +1,53 @@
...
+	if resErr != nil {
+		return AuthResponse{}, resErr
+	}
+
+	if res.StatusCode != http.StatusOK {
+		return AuthResponse{}, UnmarshalServerError(res)
+	}
+
+	var auth AuthResponse
+	var authErr error
```
tsedgwick, line 39: if we remove that defer func() - we can just inline the err creation

tsedgwick (approved):
lgtm

internal/cloud/realm/auth.go
makesitgo: result of a bad linter setting, we're squared away here now

PR #26, opened 2020-11-23
REALMC-7340: Handle CLI errors

makesitgo:


etc/testdata/baas_server_test_config.json
```diff
@@ -0,0 +1,144 @@
+{
```
tsedgwick, line 1: was this existing? not sure we want all of these vars/pws in our repo.

internal/cloud/realm/error.go
```diff
@@ -0,0 +1,36 @@
...
+// UnmarshalServerError attempts to unmarshal a server error from the *http.Response
+func UnmarshalServerError(res *http.Response) error {
+	buf := new(bytes.Buffer)
+	if _, err := buf.ReadFrom(res.Body); err != nil {
+		return err
+	}
+
+	payload := buf.String()
+	if payload == "" {
+		return ServerError{Message: res.Status}
```
tsedgwick, line 28: should this be `Code`?

tsedgwick (approved):
lgtm - just a few things

etc/testdata/baas_server_test_config.json
makesitgo: this server config exists in the old `realm-cli`, and its used to spin up a baas server in evergreen.  All the details here are more or less the same as the details you'd find in `etc/configs/local_config.json` (and similar files) that are committed in the `baas` repo.  Caveat being that I copy-pasted this, but I don't _think_ any of these values are truly secret (or meaningful) in any way

tsedgwick: my only hesitation / concern is baas is internal to us and this is public facing.

internal/cloud/realm/error.go
makesitgo: `Code` is meant to be the `baas` concept of "error code" (e.g. `"InvalidSession"`).  `Message` is ultimately what gets propagated out as the error string, so I want `Message` set here with the response's status text (should no payload be present).

Note: we're not capturing the response's _status code_ at all in this unmarshal process (i.e. whether it's a `401` or `500` or w/e)

etc/testdata/baas_server_test_config.json
makesitgo: a fair hesitation!  this has always been a bit of an issue though: the fact that realm-cli is open and baas isn't.

a benefit of doing things this way is it allows us to test against a custom, deterministic Realm server (one spun up locally alongside tests), and not have to write tests against `realm-dev.mongodb.com` (and subsequently worry about creating garbage test-data in those envs)

tsedgwick: although not as fluent as this solution, can we add an env var or a flag to pass in to point to the baas repo's local config? if we find it's too inconvenient, we can add a Makefile that is git ignored with the intent of local engineers updating it with their location. Thoughts on that?

makesitgo: I don't love the idea of having to reach into another repo for this, but don't have many great alternatives for keeping the structure of this JSON document secret.  its cumbersome for sure, but it's also only used/referenced in the `.evg.yml` script (side question: why do we commit `.evg.yml` in public repos, that seems to also expose a lot of info/details?).

the usage is here:
```
          cat > atlas_config.json <<EOF
          {
            "admin": {
              "authProviders": [
                {
                  "name": "mongodb-cloud",
                  "type": "mongodb-cloud",
                  "config": {
                      "baseUrl": "https://cloud-dev.mongodb.com",
                      "adminUsername": "${baas_mongodb_cloud_atlas_admin_username}",
                      "adminAPIKey": "${baas_mongodb_cloud_atlas_admin_api_key}"
                    },
                    "metadataFields": [{ "name": "name", "required": false}]
                }
              ]
            }
          }
          EOF
          go run cmd/server/main.go --configFile $ROOT_DIR/realm-cli/etc/testdata/baas_server_test_config.json --configFile ./atlas_config.json
```

at that point in the evergreen script, we've already cloned the `baas` repo, so if we copied over this config (im not positive its actually a dupe of an existing one yet, but can figure that out), that last command _should_ be able to be `--configFile $ROOT_DIR/baas/etc/configs/cli_test_config.json --configFile ./atlas_config.json`.

honestly, its still a bit of a hack job when running this locally, to this point I've been running the server with `local_cloud_dev_config.json` and then executing tests with:

```cmd
REALM_SERVER_BASE_URL=http://localhost:8080 REALM_MONGODB_CLOUD_ADMIN_USERNAME=tezzrgzr REALM_MONGODB_CLOUD_ADMIN_API_KEY=********-****-****-****-0923ac86a00a go test -v -tags debug github.com/10gen/realm-cli/internal/cloud/realm -run 'Test'
```

(the reason being b/c I can't that evg approach with merging an in-place config for configuring the `mongodb-cloud` auth provider without the evg env vars for username/password)

path of least resistance here seems to be porting this config over to `baas` as `cli_test_config.json` (or finding its duplicate), merging that in, then updating this script to point to that file and removing it here.  thoughts?

makesitgo: FYI: will follow up on the rest of this with this ticket: https://jira.mongodb.org/browse/REALMC-7449

cmd/root.go
```diff
@@ -47,5 +48,12 @@ func init() {
 	rootCmd.AddCommand(factory.Build(cli.WhoamiCommand))
 
 	// pass a list of functions to be run before each command's Execute function
-	cobra.OnInitialize(func() { cli.HandleErr(profile.Load()) })
+	cobra.OnInitialize(func() { handleErr(profile.Load()) })
+}
+
+func handleErr(err error) {
```
makesitgo, line 54: @tsedgwick ... here's the change that works on this `log.Fatal` handling for the CLI (which you brought up [here](https://github.com/10gen/realm-cli/pull/28#discussion_r529654776)).

figured if you still want to talk about it and/or suggest some changes, here would be the place

makesitgo: follow-up: removed this entirely and am just handling these errors in place (the CLI will/should catch-and-log any meaningful errors before they ever reach this point)

PR #28, opened 2020-11-23
REALMC-7315: Use go-cmp to handle test assertions

makesitgo:


internal/utils/test/test.go
```diff
@@ -1,64 +1,108 @@
...
 
-// MustMatch asserts the provided diff matches and fails the test if not
-func MustMatch(t *testing.T, o string) {
-	t.Helper()
-	if o != "" {
-		t.Error(o)
-	}
+// CloudAdminUsername returns the Cloud admin username to use for testing
+func CloudAdminUsername() string {
+	return os.Getenv("REALM_MONGODB_CLOUD_ADMIN_USERNAME")
```
tsedgwick, line 36: can we make this the same as baas? seems odd to have to define two different env vars for the same thing.

internal/utils/test/test.go
```diff
@@ -1,64 +1,108 @@
...
+}
+
+// CloudAdminAPIKey returns the Cloud admin api key to use for testing
+func CloudAdminAPIKey() string {
+	return os.Getenv("REALM_MONGODB_CLOUD_ADMIN_API_KEY")
```
tsedgwick, line 41: same here

internal/utils/test/test.go
```diff
@@ -1,64 +1,108 @@
...
-func NewTempDir(name string) (string, func(), error) {
-	dir, err := ioutil.TempDir("", name)
-	if err != nil {
-		return "", nil, err
+// AtlasServerURL returns the Atlas server url to use for testing
+func AtlasServerURL() string {
+	if !skipUnlessAtlasServerCalled {
+		panic("testutils.SkipUnlessAtlasServerRunning(t) must be called before testutils.AtlasServerURL()")
+	}
+	if uri := os.Getenv("REALM_MONGODB_CLOUD_API_BASE_URL"); uri != "" {
```
tsedgwick, line 49: same question

internal/cli/login.go
```diff
@@ -82,15 +84,39 @@ func (cmd *loginCommand) Handler(profile *Profile, ui terminal.UI, args []string
...
+
+	if _, err := sb.WriteString(loginErrorMsg); err != nil {
+		return loginErrorMsg
+	}
+
+	if err.reason != "" {
+		if _, err := sb.WriteString(fmt.Sprintf(": %s", err.reason)); err != nil {
+			return loginErrorMsg
+		}
+	}
```
tsedgwick, line 119: this seem overly complex - could this just become:

```
if err.reason != "" {
    return fmt.Sprintf("%s: %s", loginErrorMsg , err.reason)
}
return loginErrorMsg
```

internal/cli/home_test.go
```diff
@@ -15,7 +15,11 @@ func TestHomeDir(t *testing.T) {
...
 		home, err := homeDir()
-		u.MustMatch(t, cmp.Diff(nil, err))
-		u.MustMatch(t, cmp.Diff(fmt.Sprintf("./%s", profileDir), home))
+		assert.Nil(t, err)
+		assert.Match(t, fmt.Sprintf("./%s", profileDir), home)
 	})
 }
+
+func TestNPM(t *testing.T) {
+	assert.True(t, strings.Contains("a long string", "long"), "expected 'long' to be a substring of 'a long string'")
```
tsedgwick, line 24: what's this testing?

cmd/root.go
```diff
@@ -47,5 +48,12 @@ func init() {
...
-	cobra.OnInitialize(func() { cli.HandleErr(profile.Load()) })
+	cobra.OnInitialize(func() { handleErr(profile.Load()) })
+}
+
+func handleErr(err error) {
+	if err == nil {
+		return
+	}
+	log.Fatal(err)
 }
```
tsedgwick, line 59: is this the final result of this function? if so - can we just handle these in line to the calling func?

.evg.yml
```diff
@@ -37,6 +35,131 @@ functions:
     - command: git.get_project
       params:
         directory: realm-cli
+  "setup_mongod":
```
tsedgwick, line 38: did anything change in here from before? I'll glance over it if not.

internal/utils/test/test.go
```diff
@@ -1,64 +1,108 @@
...
-	t.Helper()
-	if o == nil {
-		t.Error("expected value to not be <nil>, but it was")
-	}
+const (
+	defaultServerURL = "http://localhost:9090"
+)
+
+var realmServerRunning = false
+var realmServerNotRunning = false
```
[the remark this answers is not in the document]
makesitgo: the two `bools` is a pattern carried over from `baas`, and is seemingly a micro-optimization.  I think the alternative (to achieve the same) would be to use a pointer to a boolean, or some other "tri-state" value.  basically, while, both are false we don't know if we answered either question of "is the server running?" or "is the server NOT running?"

but by having `realmServerNotRunning` able to be set to true, we avoid the extra network call to re-answer that same question each time.  this was probably more critical when it came to checking if mongo instances were running, as that call likely took longer than a simple HTTP GET request

internal/utils/test/test.go
```diff
@@ -1,64 +1,108 @@
...
+var skipUnlessRealmServerCalled = false
+var atlasServerRunning = false
+var atlasServerNotRunning = false
```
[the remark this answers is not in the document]
makesitgo: (same answer 🙃 )

internal/utils/test/test.go
makesitgo: So I carried over these env var names from the old `realm-cli` repo, primarily in an effort to reduce the friction of introducing this new code to the "same" evergreen build/environment.  Actually, to this point I had thought they were intentionally different to keep them separate.  Maybe @UnicodeSnowman could chime in here from a historical perspective?  I'd be happy to align them with `baas` if it doesn't break anything

internal/cli/login.go
makesitgo: Ugh, this is my bad for creating such a mess of PRs here, but this code is/should be gone away.  instead I am wrapping any errors caught from the `Handler` interface in `cli/cli.go` like this:

```go
		RunE: func(c *cobra.Command, a []string) error {
			return cmdDef.Command.Handler(
			err := cmdDef.Command.Handler(
				factory.Profile,
				newCobraUI(factory.Config.UI, c),
				a,
			)

			if err != nil {
				return fmt.Errorf("%s failed: %w", display, err)
			}
			return nil
		},
```

internal/cli/home_test.go
makesitgo: testing the reviewer?  :bushes:  started with this when fooling around with `go-cmp` options and stuff, I'll remove this.  Thanks for catching!

cmd/root.go
makesitgo: I toyed with the idea of this function "doing more", like checking for types of errors and subsequently printing more information if need be, but the more i think about it the more I think that all can be provided simply through the error string.

just for clairty: the two pieces of info on my mind that will want to _also_ be printed are any potential docs links related to the error that occurred, and any potential follow-up commands to run instead of the just-failed one.

so yeah, long story long, I think this is the final result of this function (but no regis, that's not my final answer).  right now, it's just a nil-safe process-ender, privately scoped to this `cmd/root` file and used in the 3 error-prone spots of this code.  if you'd like, more than happy to delete this helper and do the error check/log.Fatal dance in each spot above

.evg.yml
makesitgo: nothing's changed, i copy/pasted all these pieces over from the old `.evg.yml`

internal/utils/test/test.go
makesitgo: fyi: filed https://jira.mongodb.org/browse/REALMC-7449 to track the potential env var renaming work since this PR had some old/unrelated commits to it.  Resolving this thread here, for now

cmd/root.go
makesitgo: going to close this thread in favor of the comment [here](https://github.com/10gen/realm-cli/pull/26/files#r529842636) so we can continue it if need be

tsedgwick (approved):
nice work

kpatel71716 (approved):
LGTM, awesome stuff!

PR #30, opened 2020-11-30
REALMC-7154: Implement machine-friendly command output

makesitgo:
**Sample output:**
(from non-existing config)
```cmd
$ go run ./main.go whoami --profile=~/.config/realm-cli/not-a-config
INFO  14:55:51.362: No user is currently logged in.
$ go run ./main.go whoami --profile=~/.config/realm-cli/not-a-config --output-format=json
{"level":"info","time":"2020-11-30T14:55:56.687959-05:00","message":"No user is currently logged in."}
```
(from existing config)
```cmd
$ go run ./main.go whoami
INFO  14:08:57.376: User Credentials
---
{
  "API Key": "some-key",
  "Private API Key": ""
}
INFO  14:08:57.376: Session Info
---
{
  "Access Token": "",
  "Refresh Token": ""
}
$ go run ./main.go whoami --output-format=json
{"level":"info","time":"2020-11-30T14:09:04.139695-05:00","title":"User Credentials","doc":{"API Key":"some-key","Private API Key":""}}
{"level":"info","time":"2020-11-30T14:09:04.139696-05:00","title":"Session Info","doc":{"Access Token":"","Refresh Token":""}}
```


internal/terminal/log.go
```diff
@@ -0,0 +1,117 @@
...
+type LogData interface {
+	Message() (string, error)
+	Payload() ([]string, map[string]interface{}, error)
+}
+
+// Log is a terminal log
+type Log struct {
+	Level LogLevel
+	Time  time.Time
+	TZ    *time.Location
```
makesitgo, line 47: ⚠️ Setting a time zone here _works_, but doesn't feel great

The motivation behind it was to support displaying a `time.Local` formatted timestamp to the user when running `realm-cli` (hence the defaults in all `NewXXXLog` constructors), but allow for it to be overridden (specifically in tests, so that expected timestamps can be written in UTC)

Open to ideas here!

UnicodeSnowman: hmm do we _need_ local timestamps/is this a requirement? can we just use UTC everywhere and make everyone's lives easier?

internal/terminal/log_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+func TestLogConstructor(t *testing.T) {
+	assert.RegisterOpts(reflect.TypeOf(jsonDocument{}), cmp.AllowUnexported(jsonDocument{}))
+	assert.RegisterOpts(reflect.TypeOf(titledJSONDocument{}), cmp.AllowUnexported(titledJSONDocument{}), cmp.AllowUnexported(jsonDocument{}))
+
+	for _, tc := range []struct {
+		ctor          string
+		log           Log
+		expectedLevel LogLevel
+		exepctedData  LogData
```
UnicodeSnowman, line 22: [typo] `expectedData`

internal/terminal/log.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+var (
+	allLogLevels = []LogLevel{LogLevelInfo, LogLevelError}
+
+	longestLogLevel = func() LogLevel {
+		var longest LogLevel
+		var longestLength int
+		for _, level := range allLogLevels {
+			if l := len(level); l > longestLength {
```
UnicodeSnowman, line 28: [nit] could get rid of `longestLength`/avoid having to keep track of two vars here by just checking `l > len(longest)`. 🤷 there are only two log levels currently and this will only run once

UnicodeSnowman (approved):

internal/terminal/log.go
makesitgo: good enough for me (at best, there may be an added `DEBUG` but I can't imagine any more)

internal/terminal/log.go
makesitgo: no hard requirement here, I've just been trying to keep output as simple as possible, and this was a way to shave 5 characters of output off.

if no better way here for localizing a log, I'll revert back to UTC for everything (and/or remove the timestamp all together... I'll ask product/design about that one tomorrow at the checkin)

internal/terminal/log.go
```diff
@@ -0,0 +1,115 @@
...
+			}
+		}
+		return longest
+	}()
+)
+
+// LogData produces the log data
+type LogData interface {
+	Message() (string, error)
+	Payload() ([]string, map[string]interface{}, error)
```
tsedgwick, line 38: couldn't this just be the struct and the error? looks like we are just calling `Marshal` on the object at the end of the day. We could simplify this to return an `interface{}, error` and call json.Marshal directly on that returned `interface{}`

internal/terminal/log.go
tsedgwick: I agree with Chris here if there's no hard requirement. Better to make it simpler for us and not have to deal with timezones

internal/terminal/log.go
tsedgwick: Also - do we even need to return an error? Looks like we can just return an `interface{}`

tsedgwick (commented):
overall looks good, one question from me

internal/terminal/log.go
makesitgo: there are two things this approach solves:

1. allow for each `LogData` type to specify its own fields to be marshaled (e.g. `TextMessage` produces a `message` field whereas `ErrorMessage` produces an `err` field
1. ensure a certain ordering of the fields in the JSON log so that its easy/convenient to the human eye

to account for the first goal, I'm marshaling a `map[string]interface{}` so that we can support a flat yet dynamic JSON field structure (no need for a nested `data` field would would _then_ have `message` or `err` present).  since maps aren't deterministic in order, I'm using the returned `[]string` to control that

I've worked around with this a lot and this seemed to be my only path to achieve the above, but let me know if you think I'm missing an alternative approach here

---

with regards to returning the error, so far only `nil` is returned so it could be removed, but the error in the function returns was an intentional include here so that it mirrored the function type of `Message()`. these both are "computed values", and there's not really anything special about `Payload` compared to `Message` that says its should be free-from-errors

if you feel strongly about it, i'll go ahead and remove.  but in the context of how these functions are consumed, it doesn't save any downstream complexity.

internal/terminal/log.go
makesitgo: after checking with product and design in the checkin, timezones are now no more.  this has been removed

PR #31, opened 2020-12-02
REALMC-7155: Support piping command output to file

makesitgo:
Sample output:
```cmd
$ go run ./main.go whoami
22:35:29 UTC INFO  No user is currently logged in.

$ go run ./main.go login --api-key KCIMPNHJ --private-api-key ********-****-****-***-f09903cdc2a6
22:36:26 UTC ERROR login failed: failed to authenticate with MongoDB Cloud API: You are not authorized for this resource.
exit status 1

$ go run ./main.go login --api-key KCIMPNHJ --private-api-key ********-****-****-***-f09903cdc2a6 --base-url https://realm-dev.mongodb.com
22:36:39 UTC INFO  Successfully logged in.

$ go run ./main.go whoami -t json
{"time":"2020-12-01T17:37:08.901591-05:00","level":"info","title":"User Credentials","doc":{"API Key":"KCIMPNHJ","Private API Key":"********-****-****-****-f09903cdc2a6"}}

$ go run ./main.go logout -t json
{"time":"2020-12-01T17:37:35.826895-05:00","level":"info","message":"Successfully logged out."}
```


.evg.yml
```diff
@@ -261,7 +261,7 @@ buildvariants:
   expansions:
     atlas_server_url: "https://cloud-dev.mongodb.com"
     baas_supportlib_url: "https://s3.amazonaws.com/stitch-artifacts/stitch-support/stitch-support-rhel-70-4.3.2-721-ge791a2e-patch-5e2a6ad2a4cf473ae2e67b09.tgz"
-    go_url: "https://dl.google.com/go/go1.13.6.linux-amd64.tar.gz"
+    go_url: "https://dl.google.com/go/go1.14.12.linux-amd64.tar.gz"
```
makesitgo, line 264: ⚠️ this was prompted by the use of the `log.Lmsgprefix` flag when creating a `logger` in `cli/cli.go` (which allows the timestamp to be printed before log level)

internal/cli/cli.go
```diff
@@ -38,83 +42,180 @@ type CommandDefinition struct {
...
+	}
+
+	return &commandFactory{
+		config:  config,
+		profile: profile,
+		logger:  logger,
+	}
+}
+
+func (factory *commandFactory) Init() {
```
makesitgo, line 92: this function will run after flags have been parsed, but before the command's `Execute`

.evg.yml
tsedgwick: baas is on 14.10, we can easily align there and upgrade to 14.12

go.sum
```diff
@@ -464,6 +464,7 @@ github.com/ssgreg/nlreturn/v2 v2.1.0/go.mod h1:E/iiPB78hV7Szg2YfRgyIrk1AD6JVMTRk
 github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
 github.com/stretchr/objx v0.1.1 h1:2vfRuCMp5sSVIDSqO8oNnWJq7mPa6KVP3iPIwFBuy8A=
 github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/objx v0.2.0 h1:Hbg2NidpLE8veEBkEZTL3CvlkUIVzuU9jDplZO54c48=
 github.com/stretchr/objx v0.2.0/go.mod h1:qt09Ya8vawLte6SNmTgCsAVtYtaKzEcn8ATUoHMkEqE=
 github.com/stretchr/testify v1.2.1/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
 github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
```
tsedgwick, line 470: why do we have so many different versions here?

internal/cli/cli.go
```diff
@@ -38,83 +42,180 @@ type CommandDefinition struct {
...
+	Close()
+	Init()
+	Run(cmd *cobra.Command)
+	SetGlobalFlags(fs *flag.FlagSet)
+}
+
+type commandFactory struct {
+	in      *os.File
+	out     *os.File
+	err     *os.File
```
tsedgwick, line 56: this is odd to me to see `err` as an `*os.File` - can we change this field name at the very least?

internal/cli/cli.go
```diff
@@ -38,83 +42,180 @@ type CommandDefinition struct {
...
 }
 
 // CommandConfig holds the global config for a CLI command
 type CommandConfig struct {
 	RealmBaseURL string
 }
 
+// NewCommandFactory creates a new command factory
+func NewCommandFactory() CommandFactory {
+	logger := log.New(os.Stderr, "UTC ERROR ", log.Ltime|log.Lmsgprefix)
```
tsedgwick, line 76: logger is always Stderr?

internal/cli/cli.go
tsedgwick: this probably warrants a comment.

Also, can we call this Setup? I'm just biased against the word init.

tsedgwick, line 92: could we return an error here instead of being a void func?

internal/cli/cli.go
```diff
@@ -38,83 +42,180 @@ type CommandDefinition struct {
...
+		logger:  logger,
+	}
+}
+
+func (factory *commandFactory) Init() {
+	if err := factory.profile.Load(); err != nil {
+		factory.logger.Fatal(err)
+	}
+
+	if filepath := factory.config.OutputTarget.Path(); filepath != "" {
```
tsedgwick, line 97: [super nit] I always try to un-nest (de-nest?) as much as possible. Thoughts on the inverse here?
```
filepath := factory.config.OutputTarget.Path()
if filepath == "" {
    return
}
.
.
```

internal/cli/cli.go
```diff
@@ -38,83 +42,180 @@ type CommandDefinition struct {
...
+func (factory *commandFactory) Run(cmd *cobra.Command) {
+	if err := cmd.Execute(); err != nil {
+		if _, ok := err.(suppressUsageError); !ok {
+			fmt.Println(cmd.UsageString())
+		}
+
+		if factory.ui != nil {
+			err = factory.ui.Print(terminal.NewErrorLog(err))
+		}
+
```
tsedgwick, line 125: [nit] omit this line

internal/cli/cli.go
```diff
@@ -38,83 +42,180 @@ type CommandDefinition struct {
...
+		if err != nil {
+			factory.logger.Fatal(err)
+		}
+		os.Exit(1)
```
tsedgwick, line 129:
```suggestion
		if factory.ui != nil {
			err = factory.ui.Print(terminal.NewErrorLog(err))
			if err != nil {
			    factory.logger.Fatal(err)
			}
			os.Exit(1)
		}



	factory.logger.Fatal(err)
```

tsedgwick: thoughts on this? logger.Fatal seems like it will exit so refactored to os.Exit(1) only when ui isn't available

internal/terminal/flags.go
```diff
@@ -50,3 +51,27 @@ func isValidOutputFormat(outputFormat OutputFormat) bool {
...
+type OutputTarget string
+
+// Path returns the output target filepath
+func (ot OutputTarget) Path() string { return string(ot) }
+
+// String returns the output target display
+func (ot OutputTarget) String() string {
+	val := ot.Path()
+	if val == "" {
+		return "<none>"
```
tsedgwick, line 65: should "<none>" be a const?

internal/terminal/ui.go
```diff
@@ -77,34 +67,40 @@ func (ui *ui) Print(logs ...Log) error {
...
+	DisableColors bool
+	OutputFormat  OutputFormat
+	OutputTarget  OutputTarget
 }
 
-type noopFdReader struct {
+// FileDescriptor is a file descriptor
+type FileDescriptor interface {
+	Fd() uintptr
+}
```
tsedgwick, line 80: 🥂

tsedgwick (commented):
overall looks good! just some clarifying questions on my end

go.sum
makesitgo: couldn't tell you why, but I removed the `// indirect` dependencies from `go.mod` and ran `go mod tidy` which removed a little from `go.sum` (it's still big, but this part's gone)

internal/cli/cli.go
makesitgo: for sure, I'll add `Reader`/`Writer` as a suffix, since the change in types also stripped that bit of information too

internal/cli/cli.go
makesitgo: in theory it could, but it would still need to be wrapped by a void func to be passed into `cobra.OnInitialize`, a hook used to run stuff before any of the `*RunE` command phases (but ran after flags have been parsed).

internal/cli/cli.go
makesitgo: totally fine to switch to it if you feel strongly about that, but I don't love `return` here since it makes the ordering of these `Setup` steps important (all new `Setup` would need to go be between this block and `profile.Load()` above)

internal/cli/cli.go
makesitgo: you're right in that `logger.Fatal` exits, and so that's why when the `ui` is available, we'll need to `os.Exit(1)` as well.

full disclosure, this section is linter-driver.  the original version of the code:

```go
if factory.ui == nil {
	factory.logger.Fatal(err)
}
factory.ui.Print(terminal.NewErrorLog(err))
os.Exit(1)
```

the issue was an unchecked error on `factory.ui.Print`

I had no luck playing around with the `.errcheck` file we have (which has successfully ignored linter complaints for unchecked `res.Body.Close()` and `file.Close` errors).  if you've got any tips for this, I'm all ears b/c I think that above code is its simplest form

internal/terminal/flags.go
makesitgo: you raise a good point, in that this type isn't needed anymore.  going to remove it in favor of just a string flag

internal/cli/cli.go
makesitgo: yeah, `logger` probably isn't the right word for this, but this logger is created/kept for the sole purposes of calling `logger.Fatal` when a `ui` is not yet available.

`errLogger` felt verbose at the time, but I'll admit it pairs nicely with `errWriter` now. I'll update the name here

internal/cli/cli.go
makesitgo: I'm also realizing I could have also solved with this a simple ` // nolint: errcheck` inline with the `Print` call.  If you've no objections there, I'll take that route

tsedgwick (approved):
LGTM

internal/cli/cli.go
makesitgo: final resolution, just actually checking the error now (which adds another place for `factory.errLogger.Fatal` to be invoked, but doesn't involve any linting-overrides)

PR #32, opened 2020-12-03
REALMC-7497: Refactor package structure to support localized commands

makesitgo:
~Opening as draft PR and as reference for @kpatel71716 ... will not be merging this until after `app list` goes in to `next`~

We are good to go here.


internal/commands/app/command.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"github.com/10gen/realm-cli/internal/cli"
+)
+
+// Command defines the `app` command
+var Command = cli.CommandDefinition{
+	Use:         "app",
+	Aliases:     []string{"apps"},
+	Description: "Manage the apps associated with the current user",
+	Help:        "app help", // TODO(REALMC-7429): add help text description
```
kpatel71716, line 12: The list subcommand can be added here: `SubCommands: []cli.CommandDefinition{list.Command},`

internal/commands/login/inputs.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/AlecAivazis/survey/v2"
+)
+
+const (
+	inputFieldPublicAPIKey  = "publicAPIKey"
+	inputFieldPrivateAPIKey = "privateAPIKey"
+)
+
+type inputs struct {
```
kpatel71716, line 14: I spent some time thinking about this inputs struct and how it would grow/if we wanted to nest the API keys here. But this approach makes sense to me. I don't imagine this growing too massive so I think the resolve method, where we go one by one into the input fields and add a question, is a fine solution.

kpatel71716 (approved):
LGTM - this is awesome! Mostly follows what we discussed offline last week and the structure looks good to me.

Looks like the linter caught some minor things, but overall this looks great.

internal/commands/login/inputs.go
makesitgo: yeah... I'm toying around with a few ideas in the `user create` ticket I have in progress right now about this.  let's revisit this stuff on that PR

PR #33, opened 2020-12-04
REALMC-7510: Introduce event tracking to CLI

ermay12:


go.sum
```diff
@@ -187,6 +187,7 @@ github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/
 github.com/google/go-cmp v0.5.1/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.2 h1:X2ev0eStA3AbceY54o37/0PQ/UWqKEiiO2dKL5OPaFM=
 github.com/google/go-cmp v0.5.2/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
+github.com/google/go-cmp v0.5.4 h1:L8R9j+yAqZuZjsqh/z+F1NCffTKKLShY6zXTItVIZ8M=
```
ermay12, line 190: should I undo this?

internal/cli/cli.go
```diff
@@ -145,10 +162,14 @@ func (factory *commandFactory) Build(provider func() CommandDefinition) *cobra.C
 		Short: command.Description,
 		Long:  command.Help,
 		RunE: func(c *cobra.Command, a []string) error {
+			publicAPIKey := factory.profile.GetString(keyPublicAPIKey)
+			factory.tracker.Track(telemetry.NewCommandStartEvent(publicAPIKey, "idk how to get the cmd string"))
```
ermay12, line 166: how should I get the raw command string?

ermay12:
@makesitgo RFAL

also I don't think telemetry should be a global flag.   The usage as a global flag seems weird.  How often will somebody want to disable telemetry for just one command?  More likely, if someone wants to only turn off telemetry for all commands, they probably would expect there to be some config command.  Like `realm-cli config --telemetry off`. Without a config command, they might just try `realm-cli --telemetry off` which won't do what they want.

internal/cli/cli.go
```diff
@@ -145,10 +162,14 @@ func (factory *commandFactory) Build(provider func() CommandDefinition) *cobra.C
 		Short: command.Description,
 		Long:  command.Help,
 		RunE: func(c *cobra.Command, a []string) error {
+			publicAPIKey := factory.profile.GetString(keyPublicAPIKey)
```
makesitgo, line 165: I'm now thinking it might make sense to pass some extra details into the tracker constructor (after the `mode` arg, a `TrackerOptions` struct could work nicely here).  I think two right off the bat that would make sense (since they are constant throughout a command execution) would be this `PublicAPIKey` as well as a `primitive.NewObjectID().Hex()` (as the `executionID`).

that way, calls to `Track` can start to worry less about some of these other common details.  this alternate approach likely also forces you to tweak the `Track()` interface method a little so that information injected to the `tracker` (via its constructor) can be merged easily with the data being passed in at the time `Track` is called.  you definitely have a couple of way to approach this, feel free to take a stab or we can talk offline about some of them

internal/telemetry/Event.go
```diff
@@ -0,0 +1,73 @@
+package telemetry
+
+import (
+	"time"
+
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+// Event is a cli event
```
makesitgo, line 9: [nit] "is a *telemetry* event"

internal/telemetry/Event.go
```diff
@@ -0,0 +1,73 @@
...
+type Event struct {
+	ID     string
+	Type   EventType
+	UserID string    //public api key
```
makesitgo, line 13: these field comments can go away

internal/telemetry/Tracker.go
```diff
@@ -0,0 +1,39 @@
...
+
+// Tracker logs events
+type Tracker interface {
+	Track(event *Event)
+}
+
+// NewTracker constructs a new tracking service
+func NewTracker(mode Mode) Tracker {
+	switch mode {
+	case OnSelected:
```
makesitgo, line 15: go let's you combine cases, so this could be `case OnSelected, OnDefault:`

internal/telemetry/flags.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+	"strings"
+)
+
+// Mode is the Telemetry Mode
+type Mode string
+
+// String returns the string representation
+func (m Mode) String() string {
+	val := string(m)
```
makesitgo, line 13: i left you a bad pattern to follow here, but for this we can just `return string(m)` for now (and also make this a one-liner like `Type()` is below)

internal/telemetry/flags.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Type returns the Mode type
+func (m Mode) Type() string { return "string" }
+
+// Set validates and sets the output format value
+func (m *Mode) Set(val string) error {
+	mode := Mode(val)
+
+	if !isValidMode(mode) {
+		allModes := []string{OnSelected.String(),
```
makesitgo, line 25: if youre going for the multi-line declaration here, then put each item on its own line (which means the closing `}` here would also be on its own line

internal/telemetry/flags.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// set of supported telemetry modes
+const (
+	//User does not select an option
+	OnDefault Mode = "" // zero-valued to be flag's default
+	//User deliberately selects this option
+	OnSelected Mode = "on"
+	//User selects stdout
+	STDOut Mode = "stdout"
```
makesitgo, line 42: rename this to `ModeStdout`

makesitgo (approved):
i didn't go through the test files yet, but wanted to do a quick review of the code files and hopefully address the concern you raised (re: the telemetry global flag)

let me know what you think!

ermay12:
@makesitgo RFAL

internal/cli/profile.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/10gen/realm-cli/internal/telemetry"
```
makesitgo, line 7: add a newline after here to break it apart from the third part library imports

internal/cli/profile.go
```diff
@@ -182,3 +184,13 @@ func (p Profile) SetSession(accessToken, refreshToken string) {
 	p.SetString(keyAccessToken, accessToken)
 	p.SetString(keyRefreshToken, refreshToken)
 }
+
+// GetTelemetryMode gets the Telemetry Mode
```
makesitgo, line 188: [nit] lowercase "telemetry mode" here (and for SetTelemetryMode below)

internal/telemetry/Event.go
```diff
@@ -0,0 +1,80 @@
...
+// Event is a telemetry event
+type Event struct {
+	ID     string
+	Type   EventType
+	UserID string
+	Time   time.Time
+	Data   map[DataKey]interface{}
+}
+
+var (
```
makesitgo, line 18: friendly reminder to use `lower_case_names.go` for file names here

---

I see what youre accomplishing here, but i'd like to avoid this style of private configuration fields.  I think doing so ultimately it pushes us in a direction where we may want to tweak the `Track` interface signature, or introduce some sort of `EventSource` that also sits on the `commandFactory`.  we can chat offline about this one, id like to hear your thoughts and/or any other suggestions you might have

internal/telemetry/Tracker.go
```diff
@@ -0,0 +1,37 @@
+package telemetry
+
+import (
+	"fmt"
+)
+
+// Tracker logs events
+type Tracker interface {
+	Track(event *Event)
```
makesitgo, line 9: curious as to why this is a pointer to Event here?

internal/telemetry/event_test.go
```diff
@@ -0,0 +1,74 @@
+package telemetry
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/google/go-cmp/cmp"
```
makesitgo, line 9: add a newline above this import to separate out project imports and 3rd party library imports

in general, we have an import pattern across our realm repos:

```go
import (
  // go standard library imports

  // project imports

  // third party library imports
)
```

internal/cli/profile.go
```diff
@@ -182,3 +184,13 @@ func (p Profile) SetSession(accessToken, refreshToken string) {
...
+func (p Profile) GetTelemetryMode() telemetry.Mode {
+	return telemetry.NewMode(p.GetString(keyTelemetryMode))
```
makesitgo, line 190: we won't need the `NewMode` (and subsequent validation logic) here, just simply wrap `p.GetString(keyTelemetryMode)` with `telemetry.Mode( ... )`

We won't ever save an invalid telemetry mode b/c the flag that gets parsed uses `Set` on `telemetry.Mode`

internal/telemetry/flags_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	t.Run("Should have the correct type representation", func(t *testing.T) {
+		assert.Equal(t, "string", ModeOn.Type())
+	})
+
+	t.Run("Should set its value correctly with a valid output format", func(t *testing.T) {
+		tc := newMode()
+
+		assert.Nil(t, tc.m.Set("on"))
+		assert.Equal(t, "on", tc.m.String())
```
makesitgo, line 32: [nit] could use `Equal` on `ModeOn` here, and `ModeNil` below

makesitgo (commented):
one last remaining piece about how we want to build `Event`s (and/or the just the `data` associated with them)

internal/telemetry/Tracker.go
ermay12: I wasn't sure if we considered an event to be critically large enough to pass by reference instead of value.  Should I not make it a pointer?

internal/telemetry/Event.go
ermay12: I'm not sure why the file names haven't changed.  They are changed on my computer, but not in the git diff so it should have been pushed as lowercase by now.  IDK what is going on with git. maybe it's case insensitive???

internal/cli/profile.go
ermay12: why wouldn't we want to strengthen our guarantees here?  `telemetry.Mode( ... )` needlessly reduces orthogonality between profile and telemetry

internal/telemetry/Tracker.go
makesitgo: yeah, no pointer needed here

internal/cli/profile.go
makesitgo: mostly because its simpler.  it's not a situation that could or should arise based on how the whole system is designed.  another reason is for future-proof, there's no reason today that `Mode` should ever be "constructed" in code (it's passed in through flags), so we shouldn't export anything that allows developers to do so down the line

ermay12: It's literally being constructed in code right here.  However instead of using a public interface with telemetry, you want us to depend on the internal representation of Mode as a string.  Also it's possible for a user to edit the config file, so it's not a super strong guarantee that it will always be a valid Mode string.  If you still want me to do a cast I will, I just want to give one last argument for why I disagree with this style of code.

ermay12:
@makesitgo RFAL.  LMK if you don't like the tracker wrapping.  I'm happy to redo it

internal/flags/flags.go
```diff
@@ -39,4 +41,5 @@ const (
 	DisableColorsUsage = "disable output styling"
 	OutputFormatUsage  = `set the output format, available options: [json]`
 	OutputTargetUsage  = "write output to the specified filepath"
+	TelemetryModeUsage = `set the telemetry settings. This setting is sticky. Available options: ["on", "off", "stdout"]`
```
makesitgo, line 44: [nit] `enable or disable telemetry (this setting is remembered), available options: ["off", "on"]`

internal/telemetry/event.go
```diff
@@ -0,0 +1,66 @@
+package telemetry
+
+import (
+	"time"
+
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+// Event is a telemetry event
+type Event struct {
```
makesitgo, line 10: does `Event` need to be exported at this point?

internal/telemetry/event.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// NewCommandCompleteEvent creates a new event for a command's start
+func NewCommandCompleteEvent(command string) Event {
+	data := make(map[DataKey]interface{})
+	return NewEvent(EventTypeCommandStart, command, data)
+}
+
+// NewCommandErrorEvent creates a new event for a command that has errored
+func NewCommandErrorEvent(command string, err error) Event {
```
makesitgo, line 33: I think we can get rid of these three constructor functions, and simply just use `NewEvent` here.

For the `data` arg, we could leverage a variadic type so that the `ErrorEvent` can still pass in its error for now

internal/telemetry/event.go
```diff
@@ -0,0 +1,66 @@
...
+	data := make(map[DataKey]interface{})
+	data[DataKeyErr] = err
+	return NewEvent(EventTypeCommandStart, command, data)
+}
+
+// NewEvent creates a new event
+func NewEvent(eventType EventType, command string, data map[DataKey]interface{}) Event {
```
makesitgo, line 40: [nit/question] let's move `command string` ahead of `eventType EventType` in the arg positioning.  although i thought `command` was something we'd be able to construct our tracker with, along with `userID`?

with my previous comment, looking for something like this:

```go
func NewEvent(command string, eventType EventType, data ...EventData) Event {
}

type EventData struct {
  Key EventDataKey // renamed for consistency with the rest
  Value interface{}
}
```

internal/telemetry/event_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestEventConstructor(t *testing.T) {
+	assert.RegisterOpts(
```
makesitgo, line 14: can we omit this and run asserts on the individual fields here?

internal/telemetry/tracker.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Tracker logs events
+type Tracker interface {
+	Track(event Event)
+}
+
+type tracker struct {
+	userID          string
+	executionID     string
+	telemetryClient telemetryClient
```
makesitgo, line 15: this all looks fine, implementation wise.  I'd just like to reorganize and rename here, and I think we can combine the logic from `NewEvent` into what's going on here (and completely omit any event constructors being exported from `telemetry`:

```go
type Service interface {
  TrackEvent(eventType EventType, data ...EventData)
}

type service struct {
  command string
  executionID string
  userID string
  tracker Tracker
}
```

internal/telemetry/event.go
```diff
@@ -0,0 +1,66 @@
...
+
+// Event is a telemetry event
+type Event struct {
+	id          string
+	eventType   EventType
+	userID      string
+	time        time.Time
+	executionID string
+	command     string
+	data        map[DataKey]interface{}
```
makesitgo, line 17: for ease, this could probably just become a `[]EventData` until we figure out how we want to transform that data into a Segment construct

internal/cli/cli.go
```diff
@@ -89,11 +93,28 @@ func NewCommandFactory() CommandFactory {
...
+func (factory *commandFactory) configureTelemetry() error {
+	telemetryMode := factory.config.TelemetryMode
+	if telemetryMode == telemetry.ModeNil {
+		telemetryMode = factory.profile.GetTelemetryMode()
+	}
+	factory.tracker = telemetry.NewTracker(
+		telemetryMode,
+		factory.profile.GetUser().PublicAPIKey,
+		primitive.NewObjectID().Hex())
+	factory.profile.SetTelemetryMode(telemetryMode)
```
makesitgo, line 105: i'd like to only call set/save here on the profile if it differs from the flagged value.  can you check/do that before we create the `factory.tracker`?

makesitgo (commented):
next batch of comments here, just a few small questions/changes, but this looks great.  nice work!

internal/telemetry/event.go
ermay12: IDK if it needs to be, but it is the return value of an exported function so I think generally it is a good practice to export it.  Do you want me to un-export it?

makesitgo: if `NewEvent` goes away with the refactor to `telemetry.Tracker` below, is it still the return value of an exported function?

internal/telemetry/event.go
ermay12: do you want me to move configureTelemetry out of the setup function and into the build function then so we can do that?

makesitgo: yeah, that sounds fine by me, lets put it right b/w `display` initialization and the `cmd` declaration

internal/telemetry/event.go
ermay12: oh gotcha, I missed that at first

internal/telemetry/tracker.go
ermay12: Ok, but doing that kind of made the event_test.go useless, so I removed that file.  Should I try and create a tracker_test.go instead now?  Should I add a testTracker struct to help with testing purposes?

ermay12:
@makesitgo rfal

internal/telemetry/tracker.go
makesitgo: yeah that all sounds perfect... i'll take another look once you get that new test up

ermay12: @makesitgo ok just added the tests

internal/cli/cli.go
```diff
@@ -51,13 +53,14 @@ type CommandFactory interface {
...
-	errWriter *os.File
-	errLogger *log.Logger
+	config         *Config
+	profile        *Profile
+	ui             terminal.UI
+	inReader       *os.File
+	outWriter      *os.File
+	errWriter      *os.File
+	errLogger      *log.Logger
+	trackerService telemetry.Service
```
makesitgo, line 63: [nit] can we name this `telemetryService`?

internal/cli/cli.go
```diff
@@ -193,6 +206,30 @@ func (factory *commandFactory) SetGlobalFlags(fs *flag.FlagSet) {
 	fs.VarP(&factory.config.OutputFormat, flags.OutputFormat, flags.OutputFormatShort, flags.OutputFormatUsage)
 	fs.StringVarP(&factory.config.OutputTarget, flags.OutputTarget, flags.OutputTargetShort, "", flags.OutputTargetUsage)
 	fs.StringVar(&factory.config.RealmBaseURL, flags.RealmBaseURL, realm.DefaultBaseURL, flags.RealmBaseURLUsage)
+	fs.VarP(&factory.config.TelemetryMode, flags.TelemetryMode, flags.TelemetryModeShort, flags.TelemetryModeUsage)
+}
+
+func (factory *commandFactory) configureTelemetry(command string) error {
+	var telemetryMode telemetry.Mode
```
makesitgo, line 213: this is functionally fine, but we can consolidate some of the noise here with:

```go
telemetryMode := factory.config.TelemetryMode
existingTelemetryMode := factory.profile.GetTelemetryMode()

if telemetryMode == telemetry.ModeNil {
  telemetryMode = existingTelemetryMode
}

if telemetryMode != existingTelemetryMode {
  // set and save profile
}

// create tracker service
return nil
```

internal/telemetry/event.go
```diff
@@ -0,0 +1,40 @@
+package telemetry
+
+import (
+	"time"
+)
+
+// Event is a telemetry event
```
makesitgo, line 7: can remove the docstring here now

internal/telemetry/event.go
```diff
@@ -0,0 +1,40 @@
...
+const (
+	EventTypeCommandStart    EventType = "COMMAND_START"
+	EventTypeCommandComplete EventType = "COMMAND_COMPLETE"
+	EventTypeCommandError    EventType = "COMMAND_ERROR"
+)
+
+// EventDataKey used to pass data into the Event.data map
+type EventDataKey string
+
+// set of Data Keys
```
makesitgo, line 37: `// set of event data keys`

internal/telemetry/event.go
```diff
@@ -0,0 +1,40 @@
...
+	id          string
+	eventType   EventType
+	userID      string
+	time        time.Time
+	executionID string
+	command     string
+	data        []EventData
+}
+
+// EventData holds additional Event information
```
makesitgo, line 18: [nit] lowercae `event` here

internal/telemetry/flags.go
```diff
@@ -0,0 +1,49 @@
...
+// Mode is the Telemetry Mode
+type Mode string
+
+// String returns the string representation
+func (m Mode) String() string { return string(m) }
+
+// Type returns the Mode type
+func (m Mode) Type() string { return "string" }
+
+// Set validates and sets the output format value
```
makesitgo, line 17: `s/output format/mode`

internal/telemetry/flags.go
```diff
@@ -0,0 +1,49 @@
...
+func (m *Mode) Set(val string) error {
+	mode := Mode(val)
+
+	if !isValidMode(mode) {
+		allModes := []string{string(ModeOn), string(ModeStdout), string(ModeOff)}
```
makesitgo, line 22: to be consistent with our flag usage, can we omit `ModeStdout` here?

internal/telemetry/flags.go
```diff
@@ -0,0 +1,49 @@
...
+	return nil
+}
+
+// set of supported telemetry modes
+const (
+	ModeNil    Mode = "" // zero-valued to be flag's default
+	ModeOn     Mode = "on"
+	ModeStdout Mode = "stdout"
+	ModeOff    Mode = "off"
+	modeTest   Mode = "test" // note: not valid as a command line flag
```
makesitgo, line 36: we should be able to avoid this here

internal/telemetry/tracker.go
```diff
@@ -0,0 +1,78 @@
...
+// NewService creates a new service of type mode with userID and
+// executionID appended to each event it tracks
+func NewService(mode Mode, userID string, executionID string, command string) Service {
+	service := service{userID: userID, executionID: executionID, command: command}
+	switch mode {
+	case ModeOn, ModeNil:
+		service.tracker = &segmentTracker{}
+	case ModeStdout:
+		service.tracker = &stdoutTracker{}
+	case modeTest:
```
makesitgo, line 31: likewise avoiding this `case` here, in the test we should be able to create a `&service` directly.  and the `testTracker` struct definition can live in the test file

internal/telemetry/tracker_test.go
```diff
@@ -0,0 +1,49 @@
...
+			executionID: "executionID",
+			command:     "command",
+			tracker:     &segmentTracker{},
+		}
+		assert.Equal(t, expectedService, constructedService)
+	})
+}
+
+func TestEventTracking(t *testing.T) {
+
```
makesitgo, line 28: omit this newline

internal/telemetry/tracker_test.go
```diff
@@ -0,0 +1,49 @@
...
+	t.Run(fmt.Sprintf("Should track the expected event"), func(t *testing.T) {
+		trackerService := NewService(modeTest, "userID", "executionID", "command")
+		trackerService.TrackEvent(EventTypeCommandError, EventData{Key: EventDataKeyErr, Value: fmt.Errorf("error")})
+		trackerService.TrackEvent(EventTypeCommandError, EventData{Key: EventDataKeyErr, Value: fmt.Errorf("error")})
```
makesitgo, line 32: are the two `TrackEvent` calls here both required?

--

the linter's going to complain about `fmt.Errorf` here, switch it to `errors.New` instead (for L#39 below too)

internal/telemetry/tracker_test.go
```diff
@@ -0,0 +1,49 @@
+package telemetry
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/google/go-cmp/cmp"
```
makesitgo, line 9: newline between these two imports

makesitgo (approved):
lookin good, just a few more changes and once evergreen is ✅  , this is good to go

🆒  beans

PR #34, opened 2020-12-04
REALMC-7339 Add table printing to terminal

jo-tse-mongo:


internal/terminal/table.go
```diff
@@ -0,0 +1,151 @@
...
+			var val string
+			switch v := v1.(type) {
+			case string:
+				spaceRegex := regexp.MustCompile(`\s+`)
+				val = spaceRegex.ReplaceAllString(v, " ")
+			case float32, float64:
+				val = fmt.Sprintf("%g", v)
+			case fmt.Stringer:
+				val = v.String()
+			case nil:
```
makesitgo, line 54: move this to the top of the cases, remove the case body (remove `val = ""` below), and add an inline comment here:

```go
case nil: // leave zero-value
```

internal/terminal/table.go
```diff
@@ -0,0 +1,151 @@
...
+	for _, v := range values {
+		tempVal := make(map[string]string)
+		for k1, v1 := range v {
+			// convert the interface values into strings for printing
+			var val string
+			switch v := v1.(type) {
+			case string:
+				spaceRegex := regexp.MustCompile(`\s+`)
+				val = spaceRegex.ReplaceAllString(v, " ")
+			case float32, float64:
```
makesitgo, line 50: lets remove this `case` until we need it (feel free to let me know I'm missing something and we need it now)

internal/terminal/table.go
```diff
@@ -0,0 +1,151 @@
...
+	var stringValues []map[string]string
+	for _, v := range values {
+		tempVal := make(map[string]string)
+		for k1, v1 := range v {
+			// convert the interface values into strings for printing
+			var val string
+			switch v := v1.(type) {
+			case string:
+				spaceRegex := regexp.MustCompile(`\s+`)
+				val = spaceRegex.ReplaceAllString(v, " ")
```
makesitgo, line 49: lets avoid any transformation logic for now and just set the raw `string` value (feel free to let me know I'm missing something and we need this now)

internal/terminal/table.go
```diff
@@ -0,0 +1,151 @@
...
+				val = fmt.Sprintf("%g", v)
+			case fmt.Stringer:
+				val = v.String()
+			case nil:
+				val = ""
+			default:
+				val = fmt.Sprintf("%v", v)
+			}
+			valLen := len(val)
+			if valLen > maxHeaderWidth[k1] {
```
makesitgo, line 60: [nit]

```go
if v := len(normalizedValue); v > t.columnWidths[key] {
  t.columnWidths[key] = v
}
```

internal/terminal/table.go
```diff
@@ -0,0 +1,151 @@
...
+	resultTable := &table{
+		headers,
+		stringValues,
+		maxHeaderWidth,
+	}
+	return resultTable
+}
+
+func (t table) writeHeader() string {
+	var tempBuilder strings.Builder
```
makesitgo, line 77: [nit] `tempBuilder` => `builder`

also, do you think we could accomplish this with just one string Builder?  it seems like we ultimately want:

```go
h := color.New(color.Bold).SprintFunc()(header)
builder.WriteString(fmt.Sprintf("%s"+strings.Repeat(" ", spaceCount), h))
```

internal/terminal/table.go
```diff
@@ -0,0 +1,151 @@
...
+	var result strings.Builder
+	for i, h := range t.headers {
```
makesitgo, line 79: [nit] `for _, header := range t.headers`

internal/terminal/table.go
```diff
@@ -0,0 +1,151 @@
...
+		spaceCount := t.maxHeaderWidth[h] - len(h) + columnBreakWidth
+		// If it's not the last column
```
makesitgo, line 81: can you run me through this logic real quick?  having trouble figuring out why we're subtracting the gutter from the header width

internal/terminal/table.go
```diff
@@ -0,0 +1,151 @@
...
+// maxHeaderWidth is to determine the column width based on the max length of the values and header
+type table struct {
+	headers        []string
+	data           []map[string]string
+	maxHeaderWidth map[string]int
+}
+
+// columnBreakWidth is to set a standard for the break between columns for the table
+const (
+	 columnBreakWidth = 2
```
makesitgo, line 23: [nit] `columnBreakWidth` => `gutterWidth`

internal/terminal/table.go
```diff
@@ -0,0 +1,151 @@
...
+
+		result.WriteString(fmt.Sprintf(tempBuilder.String(), color.New(color.Bold).SprintFunc()(h)))
+		tempBuilder.Reset()
+	}
+	return result.String()
+}
+
+func (t table) writeValues() string {
+	var result strings.Builder
+	var tempBuilder strings.Builder
```
makesitgo, line 97: same thing here, i'll let you try to work through this implementation where we're only needing a single `var builder strings.Builder`

internal/terminal/table.go
```diff
@@ -0,0 +1,151 @@
...
+
+	for _, valueMap := range t.data {
```
makesitgo, line 99: same naming scheme here: `data` has many `row`s, `row` is build of `key` and `value`s, etc

internal/terminal/table.go
```diff
@@ -0,0 +1,151 @@
...
+			}
+			result.WriteString("\n")
+		}
+	}
+	return result.String()
+}
+
+func (t table) writeDividers() string {
+	var divider strings.Builder
+	var tempBuilder strings.Builder
```
makesitgo, line 122: same thing here, i'll let you try to work through this implementation where we're only needing a single `var builder strings.Builder`

internal/terminal/table.go
```diff
@@ -0,0 +1,151 @@
...
+			tempBuilder.WriteString("  ")
+		}
+		divider.WriteString(tempBuilder.String())
+		tempBuilder.Reset()
+	}
+	return divider.String()
+}
+
+const (
+	headers = "headers"
```
makesitgo, line 135: prefix this with `logField` (and for `data` below)

internal/terminal/table.go
```diff
@@ -0,0 +1,151 @@
...
+	data    = "data"
+)
+
+func (t table) Message() string {
```
makesitgo, line 139: to satisfy this interface, the return signature here will want to be `(string, error)`
also, at the start of this and `Payload()` below, we can add a `if err := t.validate(); err != nil { return "", err }` check.  inside that validate check is where you can move the logic you had for empty/nil headers

internal/terminal/table.go
```diff
@@ -0,0 +1,151 @@
...
+	var tempBuilder strings.Builder
+	tempBuilder.WriteString("\n")
+	tempBuilder.WriteString(strings.Join([]string{t.writeHeader(), t.writeDividers(), t.writeValues()}, "\n"))
+	return tempBuilder.String()
+}
+
+func (t table) Payload() ([]string, map[string]interface{}, error) {
+	return []string{data, headers}, map[string]interface{}{
+		"headers": t.headers,
```
makesitgo, line 148: just a reminder to use the `const`s you defined for these values here (`logFieldHeaders` and `logFieldData`)

makesitgo (commented):
did a first pass through the main `table.go` file.  really no major comments from me though, a lot of them are just some standard go practices (naming, organization and such) and other various nits of mine.  this is lookin really good so far!

internal/terminal/table.go
jo-tse-mongo: the float gets cut off at a certain number, so i make it formatted by %g (which handles long floats and defaults to %f if it's short)

makesitgo: neat!  i definitely did not know about the `%g` formatter... i guess my question here is: what `float`s are already trying to print here?  if this was just for future-proofing, then I'll still suggest we remove this `case` until the need/requirement for it shows up

tsedgwick:
before merging, please remember to update the PR title in the form of `REALMC-xxxx: imperative sentence`

internal/terminal/table.go
```diff
@@ -0,0 +1,149 @@
...
+	data           []map[string]string
+	maxHeaderWidth map[string]int
+}
+
+// columnBreakWidth is to set a standard for the break between columns for the table
+const (
+	 columnBreakWidth = 2
+)
+
+func newTable(headers []string, values []map[string]interface{}) *table {
```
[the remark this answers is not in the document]
jo-tse-mongo: I return a pointer so that I could return a nil instead of a table if the values are empty (header, data) 
I also have a question in regards to the other comment that involves implementing a validate function, since if I return a nil here, there should be no real way for them to create a table without going through the initial validation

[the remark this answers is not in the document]
makesitgo: yeah let's avoid using a pointer just to return `nil` here, the zero-value of the struct will be just fine

internal/terminal/table.go
jo-tse-mongo: If we don't replace any possible newlines, it breaks up the table formatting (\t, \n, etc) so I try to replace them here

makesitgo: i see, that makes a lot of sense.  the one thing is I'm not sure we have/expect any table content to contain any newlines at this point.  similar to the `float`s, I'd rather remove this logic now and only add it if/when we need it (i.e. once the requirement for such logic is more defined/concrete)

internal/terminal/table.go
```diff
@@ -0,0 +1,131 @@
...
+				spaces := strings.Repeat(" ", t.columnWidths[header]-len(dataMap[header]))
+				rowResult = append(rowResult, fmt.Sprintf("%s"+spaces, dataMap[header]))
+			}
+			result = append(result, strings.Join(rowResult, strings.Repeat(" ", gutterLength)))
+		}
+	}
+	return strings.Join(result, "\n")
+}
+
+func (t table) writeDivider() string {
```
makesitgo, line 92: can we change the names of the functions `write*` to `*String`?  the  verb `write` makes me think we're actually working with a `Writer` here, as opposed to just string building

makesitgo (commented):
ok, this should be the last batch of big comments for the `table.go` file.  wanted to get these in and i'll move on to the test file, but feel free to start addressing these

internal/terminal/table.go
```diff
@@ -0,0 +1,131 @@
...
+	data         []map[string]string
+	columnWidths map[string]int
+}
+
+const (
+	// gutterLength is to set a standard for the break between columns for the table
+	gutterLength = 2
+)
+
+func parseInterface(data interface{}) string {
```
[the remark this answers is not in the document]
jo-tse-mongo: Should I replace **all** references to data into value??

internal/terminal/table.go
```diff
@@ -0,0 +1,131 @@
...
+
+	for _, header := range headers {
+		if headerLen := len(header); headerLen > resultTable.columnWidths[header] {
+			resultTable.columnWidths[header] = headerLen
+		}
+	}
+
+	for _, row := range data {
+		normalizedRow := make(map[string]string)
+		for key, rowData := range row {
```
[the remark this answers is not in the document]
jo-tse-mongo: Since we need to go through all the values for each header, does it matter if we go through them in the header order? Is there a performance difference with looping through a map vs an array?

internal/terminal/table.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (t table) Message() (string, error) {
+	if err := t.validate(); err != nil {
+		return "", err
+	}
+	return "\n" + strings.Join([]string{
+		t.writeHeader(),
+		t.writeDivider(),
+		t.writeRows(),
```
mburdette, line 119: [naming-nit] I think nick has kind of already covered this, but naming these all `writeX` seems a bit odd since they aren't writing anything, they are just building strings. Maybe `printX` would be better for each since it is printing X and goes well with naming this function `Print`?  `Print` to me makes the most sense since go itself uses that naming convention when printing strings.

internal/terminal/table.go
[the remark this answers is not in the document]
makesitgo: no, just within this one function here

internal/terminal/table_test.go
```diff
@@ -0,0 +1,272 @@
...
+				"float":  "1",
+				"float2": "1.234567890123",
+				"array":  "[1 2 3]",
+				"nil":    "",
+			},
+		})
+	})
+}
+
+func TestTable_Message(t *testing.T) {
```
mburdette, line 82: [nit] I think we typically avoid `_` in Test names

internal/terminal/table.go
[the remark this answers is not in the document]
makesitgo: what this ensures is our `normalizedData` will be stripped of any row-cell values that are not specified by any of the headers.

i'm thinking of a situation like:

```go
table := newTable(
  []string{"a", "b", "c"},
  []map[string]interface{}{
    "a": "first",
    "b": "second",
    "c": "third",
    "d": "will not show up in Payload() now",
  },
)
```

internal/terminal/table_test.go
```diff
@@ -0,0 +1,272 @@
...
+	t.Run("Should print only a header for a table with no data", func(t *testing.T) {
+		table := newTable([]string{"header", "only", "no", "data"}, nil)
+		message, error := table.Message()
+		assert.Nil(t, error)
+		assert.Equal(t, message, "\n"+fmt.Sprintf("%s  %s  %s  %s",
+			[]interface{}{
+				color.New(color.Bold).SprintFunc()("header"),
+				color.New(color.Bold).SprintFunc()("only"),
+				color.New(color.Bold).SprintFunc()("no"),
+				color.New(color.Bold).SprintFunc()("data"),
```
mburdette, line 102: For all the occurrences of `"data"` and `"header"` wouldn't it be better to use the `const` values in `table.go` in the tests?

internal/terminal/table.go
```diff
@@ -0,0 +1,131 @@
...
+	}, "\n") + "\n", nil
+}
+
+func (t table) Payload() ([]string, map[string]interface{}, error) {
```
mburdette, line 123: Why is this called `Payload` ?  I feel like the term `Payload` is usually associate with api requests. Also can we get comments on exported methods similar to what we do for the server code? 

Also the return type of this seems odd, it just always returns the 2 keys for the data? This doesn't really seem necessary and it seems like we could just have this return `map[string]interface{}, error` Or better yet, wouldn't it make more sense to just have 2 separate functions `Headers()` and `Data()` that basically act as getters?

Maybe I'm missing something here though...

internal/terminal/table.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+func (t table) validate() error {
+	if len(t.headers) == 0 {
+		return errors.New("cannot create a table without headers")
+	}
+	return nil
+}
+
+func (t table) Message() (string, error) {
```
mburdette, line 112: Should this method be called `Print` since it is printing the table? Also can we get some comments on exported methods similar to what we do in the server code (Usually the linter picks this up there)?

internal/terminal/table.go
```diff
@@ -0,0 +1,131 @@
...
+func (t table) writeHeader() string {
+	var result []string
+	for _, header := range t.headers {
+		spaces := strings.Repeat(" ", t.columnWidths[header]-len(header))
+		result = append(result, fmt.Sprintf("%s"+spaces, color.New(color.Bold).SprintFunc()(header)))
+	}
+	return strings.Join(result, strings.Repeat(" ", gutterLength))
+}
+
+func (t table) writeRows() string {
```
mburdette, line 76: [nit] rename to `printRows`

internal/terminal/table.go
```diff
@@ -0,0 +1,131 @@
...
+				resultTable.columnWidths[key] = rowDataLen
+			}
+			normalizedRow[key] = normalizedRowData
+		}
+		resultTable.data = append(resultTable.data, normalizedRow)
+	}
+	return resultTable
+}
+
+func (t table) writeHeader() string {
```
mburdette, line 67: [nit] rename to `printHeader`

internal/terminal/table.go
mburdette, line 92: [nit] rename to `printDivider`

mburdette (changes requested):
Looking pretty good, mostly naming changes from me

internal/terminal/table.go
makesitgo: `Message` and `Payload` are interface methods of `LogData`, which supports representing arbitrary data in either string form (`Message`) or JSON (`Payload`).  the return type of `Payload` is the way it is so we can ensure a flat `Log` JSON payload with deterministic key ordering

internal/terminal/table.go
```diff
@@ -0,0 +1,131 @@
...
+	columnWidths map[string]int
+}
+
+const (
+	// gutterLength is to set a standard for the break between columns for the table
+	gutterLength = 2
+)
+
+func parseInterface(data interface{}) string {
+	var normalizedData string
```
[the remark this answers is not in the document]
makesitgo: yeah, same thing. a habit i picked up from eric, from I tend to prefer the `var` zero-value declaration if appropriate (I recently learned, however, Tim does not 🤔 )

internal/terminal/table.go
makesitgo: linter's probably not complaining b/c its an unexported type.  i think no comments is fine here, or otherwise a comments overhaul is in order b/c all the other types should be doing it too

I'd prefer we just document `LogData` interface really well, instead of docstrings here

internal/terminal/table.go
jo-tse-mongo: @mburdette  should it be headerString or printHeader ?

internal/terminal/table.go
```diff
@@ -0,0 +1,131 @@
...
+	var result []string
+	for _, header := range t.headers {
+		spaces := strings.Repeat(" ", t.columnWidths[header]-len(header))
+		result = append(result, fmt.Sprintf("%s"+spaces, color.New(color.Bold).SprintFunc()(header)))
+	}
+	return strings.Join(result, strings.Repeat(" ", gutterLength))
+}
+
+func (t table) writeRows() string {
+	var result []string
```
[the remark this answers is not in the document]
jo-tse-mongo: If I make the changes to use `make`, what happens is that rows that are empty are defaulted to `""`, so in the end, it adds a newline for each empty row. 
I could filter it out at the end, but it'd be another iteration as opposed to appending it using the `var` method

internal/terminal/table_test.go
jo-tse-mongo: Most of these occurrences of "data" and "header" are really values within the actual header that just so happened to be named "header" and "data"

I'll change it for the `Payload` test though

internal/terminal/table.go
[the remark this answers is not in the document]
makesitgo: i think you have it right, empty rows should be omitted.  does this same behavior occur when using `make` with the `0` valued-length? for example: `cells := make([]string, 0, len(t.data))` (as opposed to just `cells := make([]string, len(t.data))`)?

internal/terminal/table.go
```diff
@@ -0,0 +1,139 @@
...
+	for i, header := range t.headers {
+		cells[i] = fmt.Sprintf("%s%s",
+			color.New(color.Bold).SprintFunc()(header),
+			strings.Repeat(" ", t.columnWidths[header] - len(header)),
+		)
+	}
+	return strings.Join(cells, gutter)
+}
+
+func (t table) rowString() string {
```
makesitgo, line 93: sorry about this, i was being dumb earlier!  I didn't connect the dots this was stringifying all the rows, the name should probably either `rowsString` or `dataString` (I think I'm leaning towards the latter, but your call there).

and then `cells` should probably turn into `rows` instead.... sorry about that :bushes:

internal/terminal/table.go
```diff
@@ -0,0 +1,139 @@
...
+	cells := make([]string, len(t.data))
+	for i, dataMap := range t.data {
+		// if there's no data for that row, then ignore
+		if len(dataMap) > 0 {
+			row := make([]string, len(t.headers))
```
makesitgo, line 98: then `row` here could be `cells`

internal/terminal/table.go
```diff
@@ -0,0 +1,139 @@
...
+				)
+			}
+			cells[i] = strings.Join(row, gutter)
+		}
+	}
+	return strings.Join(cells, "\n")
+}
+
+func (t table) dividerString() string {
+	divider := make([]string, len(t.headers))
```
makesitgo, line 113: can this name be something pluralized, since its an array?  even `parts` would be fine, or `cells` to remain consistent with the other `*String` methods

internal/terminal/table.go
```diff
@@ -0,0 +1,139 @@
...
+	if len(t.headers) == 0 {
+		return errors.New("cannot create a table without headers")
+	}
+	return nil
+}
+
+func parseValue(value interface{}) string {
+	parsed := ""
+	switch v := value.(type) {
+	case nil: // leave zero data
```
makesitgo, line 130: [nit] `// leave zero-value` or `// do nothing`

internal/terminal/table_test.go
```diff
@@ -0,0 +1,263 @@
+package terminal
+
+import (
+	"fmt"
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/fatih/color"
+	"testing"
+)
+
+func TestEmptyTableConstructor(t *testing.T) {
```
makesitgo, line 10: I think we can combine the `TestEmptyTableConstructor` and `TestTableConstructor` into a single `TestNewTable` test function.

the test table would then look like:

```go
for _, tc := range []struct{
  description string
  header []string
  data []map[string]interface{}
  expectedData []map[string]string
  expectedWidths []map[string]int
}{
  // test cases
} {
  // inside the test here, you could then just assert.Equal(t, tc.expectedData, t.data) and the like
}
```

what this will allow is:
1. the two empty constructor tests you have here are effectively equal, then they could be defined with the above test case struct as simply:

```go
{
  description: "Should return an empty table if no headers are provided",
}
```

(no declarations for `tc.header` nor `tc.data` are needed here, and the equality checks with `expectedData` and `expectedWidths` also defined as `nil` should pass just fine)

the other two test cases you have for `TestTableConstructor` should fit in nicely here as well

internal/terminal/table_test.go
```diff
@@ -0,0 +1,263 @@
...
+		assert.Equal(t, table.data, []map[string]string{
+			{
+				"test":   "should show up",
+			},
+		})
+	})
+}
+
+func TestTableMessage(t *testing.T) {
+	t.Run("Should return an empty string and an error if the table has no headers", func(t *testing.T) {
```
makesitgo, line 78: with enough test cases written for `TestNewTable`, then the test for `TestTableMessage` and `TestTablePayload` should end up being very simple, we'll likely only want to confirm:
1. with normal data, normal output is produced (edge case-y tests should likely all live in `TestNewTable`)
1. with `nil` headers, we return an error

I see that's basically what you already have with `TestTablePayload`, so lets just port the edge cases from `TestTableMessage` here to `TestNewTable`

makesitgo (commented):
I think its nearly there! just this last bit, and I'll wait to see how the test file shakes out before finally approving here.

again, really nice work here.  i'm loving how this is all coming together!

internal/terminal/table_test.go
```diff
@@ -0,0 +1,264 @@
...
+		header         []string
+		data           []map[string]interface{}
+		expectedHeader []string
+		expectedData   []map[string]string
+		expectedWidths map[string]int
+	}{
+		{
+			description: "Should return empty table if there is nil for the header",
+			header:      nil,
+			data: []map[string]interface{}{
```
makesitgo, line 22: can remove `data` here, it's irrelevant to the test path

internal/terminal/table_test.go
```diff
@@ -0,0 +1,264 @@
...
+				{"test": "should not show up"},
+			},
+		},
+		{
+			description: "Should return empty table if there is no header information",
+			header:      []string{},
+			data: []map[string]interface{}{
```
makesitgo, line 29: can remove `data` here, it's irrelevant to the test path

internal/terminal/table_test.go
```diff
@@ -0,0 +1,264 @@
...
+					color.New(color.Bold).SprintFunc()("two")}...,
+			) + `
+------  ---  -----
+-3.123  1    2    
+        1    2    
+        1         
+             [2 3]`,
+		},
+		{
+			description: "Should return a table where the longest length is determined by the longest value in the column as a whole",
```
makesitgo, line 154: this test case could be accomplished with an assert on `expectedColumnWidths` in `TestNewTable` above (and will prove to be a little easier to read without all the string formatting required in the message assertion here)

internal/terminal/table_test.go
```diff
@@ -0,0 +1,264 @@
...
+					color.New(color.Bold).SprintFunc()("two"),
+					color.New(color.Bold).SprintFunc()("three")}...,
+			) + `
+---  -----  -----
+1    2.222  three
+1    2      3    
+1.0  2      3    `,
+		},
+		{
+			description: "Should return rows with values specified in the header, not insert new columns, and not print empty rows",
```
makesitgo, line 128: this test case could be accomplished with an assert on `expectedData` in `TestNewTable` above (and will prove to be a little easier to read without all the string formatting required in the message assertion here)

internal/terminal/table_test.go
```diff
@@ -0,0 +1,264 @@
...
+----------------------  ------  ----  -------------------------  -------------
+[1 test this]           12.34   1     {1234 12.345}              tester string
+[1 2 3 4]               12.34   1     map[test:1 this:2]         test         
+[1 2.3 4.5555555555 6]  123.34  -2    map[what happens:[1 2 3]]  hello        `,
+		},
+	} {
+		t.Run(tc.description, func(*testing.T) {
+			table := newTable(tc.header, tc.data)
+			message, err := table.Message()
+			if err != nil {
```
makesitgo, line 222: the error and success cases should be separate `t.Run`s, instead of using conditional asserts here

internal/terminal/table_test.go
```diff
@@ -0,0 +1,264 @@
...
+					color.New(color.Bold).SprintFunc()("short"),
+					color.New(color.Bold).SprintFunc()("longer header that will determine length"),
+					color.New(color.Bold).SprintFunc()("short again")}...,
+			) + `
+----------------------------------------------------  ----------------------------------------  -----------
+1                                                     tiny                                      123        
+the word short is actually longer than the word long  tiny again                                90         `,
+		},
+		{
+			description: "Should return correctly formatted values in the table",
```
makesitgo, line 171: this test case should be duplicated somehow, but with `color.DisableColor = true` set at the start of one of the tests. then we can assert the string is printed without bold formatting as well

internal/terminal/table_test.go
```diff
@@ -0,0 +1,264 @@
...
+	}{
+		{
+			description: "Should return an empty string and an error if the table has no headers",
+			data: []map[string]interface{}{
+				{"test": 123, "this": "456", "data": []string{"7", "8 9", "10!"}},
+			},
+			expectedError: "cannot create a table without headers",
+		},
+		{
+			description: "Should print only a header for a table with no data",
```
makesitgo, line 94: this test case could be accomplished with an assert on `expectedData` in `TestNewTable` above (and will prove to be a little easier to read without all the string formatting required in the message assertion here)

internal/terminal/table_test.go
```diff
@@ -0,0 +1,264 @@
...
+					color.New(color.Bold).SprintFunc()("only"),
+					color.New(color.Bold).SprintFunc()("no"),
+					color.New(color.Bold).SprintFunc()("data"),
+				}...,
+			) + `
+------  ----  --  ----
+`,
+		},
+		{
+			description: "Should return a string representing the table",
```
makesitgo, line 108: this test case could be accomplished with an assert on `expectedData` in `TestNewTable` above (and will prove to be a little easier to read without all the string formatting required in the message assertion here)

internal/terminal/table_test.go
```diff
@@ -0,0 +1,264 @@
...
+				assert.Equal(t, "", tc.expectedError)
+			}
+			assert.Equal(t, message, tc.expectedMessage)
+		})
+	}
+}
+
+func TestTablePayload(t *testing.T) {
+	t.Run("Payload should return an error with a table without a header", func(t *testing.T) {
+		data := []map[string]interface{}{
```
makesitgo, line 234: you can omit `data` here, its not part of the test path

internal/terminal/table_test.go
```diff
@@ -0,0 +1,264 @@
...
+			{"test": 123, "this": "456", "data": []string{"7", "8 9", "10!"}},
+		}
+		table := newTable(nil, data)
+		payloadKeys, payloadData, err := table.Payload()
```
makesitgo, line 238: this is fine as is, just commenting here to say for error tests, it's totally ok to not run assertions on the returned values other than `err`

doing `_, _, err := table.Payload()` would be acceptable here as well

internal/terminal/table_test.go
```diff
@@ -0,0 +1,264 @@
...
+		data := []map[string]interface{}{
+			{"test": 123, "this": "456", "data": []string{"7", "8 9", "10!"}},
+		}
+		table := newTable(header, data)
+		payloadKeys, payloadData, err := table.Payload()
+		assert.Nil(t, err)
+		assert.Equal(t, payloadKeys, []string{logFieldData, logFieldHeaders})
+		assert.Equal(t, payloadData[logFieldHeaders], header)
+
+		dataStr := []map[string]string{
```
makesitgo, line 255: does this still want to be named `dataStr`?  I think you could go `expectedData` here, or just `data` (which would necessitate you to inline `header` and `data` above into the `newTable` constructor)

makesitgo (approved):
just a few last cleanup items/questions, but it will be good to go after that

let's just sync up prior to merging here so we can quickly touch base on the rebase/squash/commit message protocol here

nice work!

🆒  beans

internal/terminal/table_test.go
jo-tse-mongo: For these tests, I left it mostly to check for the whitespace after the values as well as to see what the result of arrays, maps, etc. look like in this table printing format

makesitgo (approved):
🆒  beans

mburdette:
@jo-tse-mongo the PR title and main commit don't need `realm-cli` in them, since we know repo we are in 😅

internal/terminal/table.go
mburdette: the string print methods in golang use `sprint` so maybe that makes sense ?
https://golang.org/pkg/fmt/#Sprint

mburdette (approved):
LGTM

PR #35, opened 2020-12-04
REALMC-7156 add app list command + tests

kpatel71716:


internal/cli/app/list_command.go
```diff
@@ -0,0 +1,67 @@
+package app
```
kpatel71716, line 1: I put this file flat with the app command.go file temporarily. I felt it was easier to do it this way and won't matter in the long run once this is refactored with the subcommand pattern discussed earlier

internal/cli/profile.go
```diff
@@ -157,21 +158,15 @@ func (p Profile) SetUser(publicAPIKey, privateAPIKey string) {
 	p.SetString(keyPrivateAPIKey, privateAPIKey)
 }
 
-// Session is the CLI profile session
-type Session struct {
```
kpatel71716, line 161: This may be a bit controversial, but I removed the session from profile and put it in client.go. The intent was to allow a user to create an Auth client that had session information so I moved session to the realm client since we would have a circular import if we imported profile.Session in the realm client. Don't mind a different approach at all here

internal/cloud/realm/app.go
```diff
@@ -0,0 +1,137 @@
...
+	return profile, nil
+}
+
+func (c *client) FindProjectAppByClientAppID(groupIDs []string, appFlag string) ([]App, error) {
+	var appArr []App
+	appArr = make([]App, 0)
+	for _, groupID := range groupIDs {
+		apps, err := c.FetchAppsByGroupID(groupID)
+		if err != nil && err != errGroupNotFound {
+			//TODO REALMC-7156: should we fail silently or hard here
```
kpatel71716, line 88: I had a silent failure here in the scenario where we try to get apps from one of the project IDs tied to the user and one of the calls fails. Here we just keep going as if nothing went wrong as long as the error is not a GroupNotFound error. I am fine letting all errors go through too. I copied this logic from the Realm CLI but in our use cases, I don't think it matters why we failed to fetch the apps for a group ID

internal/cloud/realm/app.go
```diff
@@ -0,0 +1,137 @@
...
+			return nil, err
+		}
+		appSet := make(map[string]bool)
+		for _, appElem := range apps {
+			if len(appFlag) > 0 {
+				if appElem.Name == appFlag {
+					appArr = append(appArr, *appElem)
+					// Greedily return the first app that matches if --app flag provided
+					// Note: User could have multiple projects with an app of the same name]
+					return appArr, nil
```
kpatel71716, line 98: We can remove this return here to account for scenarios where a user has an app that shares the name of an app from another project

internal/cloud/realm/client.go
```diff
@@ -17,16 +18,31 @@ const (
...
 
 // NewClient creates a new Realm client
 func NewClient(baseURL string) Client {
-	return &client{baseURL}
+	return &client{baseURL: baseURL}
+}
+
+// NewAuthClient creates a new Realm client with a session used for Authorization
+func NewAuthClient(baseURL string, session *Session) Client {
+	// TODO: REALMC-7156 should we return an error here?
```
kpatel71716, line 33: Not sure if we want some kind of validation on a session here, or if we want to use the refresh token when creating the client to get a fresh token

internal/cloud/realm/client.go
```diff
@@ -17,16 +18,31 @@ const (
...
+	return &client{baseURL, session}
 }
 
 type client struct {
 	baseURL string
+	session *Session
+}
+
+// Session is the CLI profile session TODO REALMC-7156 figure out this approach i.e. moving Session to here
```
kpatel71716, line 42: See comment in profile.go

internal/cloud/realm/client.go
```diff
@@ -41,6 +57,14 @@ func (c *client) do(method, path string, options api.RequestOptions) (*http.Resp
...
 	req.Header.Set(requestOriginHeader, cliHeaderValue)
 
+	// TODO REALMC-7156 How do we want to handle refresh tokens here
+	if c.session != nil {
+		if len(c.session.AccessToken) == 0 {
+			return nil, errors.New("the current session is invalid. This can happen if you have not yet logged in or if your refresh token has expired")
+		}
+		req.Header.Add("Authorization", "Bearer "+c.session.AccessToken)
+	}
+
```
kpatel71716, line 67: This logic is pretty 'dumb'. We have no idea if the session here is valid or not and don't really check anything here. This means that even though we could know that a future request will fail, we do nothing and let that happen. Open to some ideas/discussions here.

internal/cloud/realm/client.go
```diff
@@ -17,16 +18,31 @@ const (
...
+func NewAuthClient(baseURL string, session *Session) Client {
+	// TODO: REALMC-7156 should we return an error here?
+	return &client{baseURL, session}
 }
 
 type client struct {
 	baseURL string
+	session *Session
+}
+
```
kpatel71716, line 41: I added a session to the client object here to allow us to create an 'AuthClient', which can implicitly handle authorization for requests that use an Auth client. I created a separate NewAuthClient function to create a client with a session, and the rest of the methods are shared. 'client.do' adds an Authorization header to a request if the session in the client instance is not nil, which I felt was fine

internal/cli/profile.go
makesitgo: The package cycle callout makes sense, I think this is a fine approach for now (we'll be doing something similar with `telemetry` and the profile as well).

My only suggestions around this would be:
1. Move this to `realm/auth.go` and replace `AuthResponse` with this struct (keeping its name `Session`)
1. Make sure to add the `json` tags there for unmarshaling the server response, and update some variable names in the login code below
1. Keep `session` as a field on the `client`, but just as a value (don't use a pointer)

I'll touch on the validation part of all this (and how we should handle auth/refresh) in a later comment

internal/cli/app/list_command.go
makesitgo: this is fine here for now (though, the `list_command` and its corresponding test file should have names that match each other).

ultimately this file would live in a directory at `internal/cli/app/list/command.go`, but again, this is totally fine as is for now.  i'll be handling the major refactor stuff in a follow-up PR to this one

internal/cloud/realm/client.go
makesitgo: you can keep this as is (minus the pointer to `Session`).  error/validation logic will be put at the point of request execution time: if auth (or refresh) is required for an api call and we don't see any value for the particular access/refresh token we need, that's when we can return a "must authenticate" error.

internal/cloud/realm/client.go
makesitgo: OK, so I like where this is headed, but I think we could make a few tweaks here to set us up nicely in the future.

first change will be to the `api.RequestOptions` struct that this `do` function takes in as an arg:
1. Remove the `Header` field
1. Add `ContentType string`, `UseAuth bool`, and `RefreshAuth bool` fields
1. We can also remove the `JSONRequestOptions` func, we'll handle that elsewhere now

now in `client.go`, we're going to have two private methods responsible for executing requests:
`func (c *client) do(method, path string, options api.RequestOptions) (*http.Response, error)`
`func (c *client) doJSON(method, path string, payload interface{}, options api.AuthOptions) (*http.Response, error)`

I'll leave the implementations to you, but it stands to reason `doJSON` would be doing the stuff you saw in `JSONRequestOptions` and then just returning `c.do(...)`

So you'll have to tweak some of the implementation in the `do` function to accommodate the new options (with content type, make sure you only `Set` it on the header if its not blank).  For authorization, I think it'd make sense to have the following helper function:
`func (c *client) getAuth(options api.RequestOptions) (string, error)`
which was responsible for returning the appropriate session token based on the `bool`s found in `RequetOptions`.  It is here also that we'd return an error if say `UseAuth` was `true` but then `Session.AccessToken` was blank

thoughts?

internal/cloud/realm/app.go
```diff
@@ -0,0 +1,137 @@
...
+
+		if role.GroupID == projectFlag {
+			return []string{role.GroupID}
+		}
+	}
+
+	return groupIDs
+}
+
+func (c *client) GetUserProfile() (UserProfile, error) {
```
makesitgo, line 59: let's move `GetUserProfile` to the `auth.go` file of this package (along with its `struct`s, it's probably worth including the `AllGroupIDs() []string` function on the `UserProfile` type that the old `realm-cli` had)

internal/cloud/realm/app.go
```diff
@@ -0,0 +1,137 @@
...
+
+	var profile UserProfile
+	if err := dec.Decode(&profile); err != nil {
+		return UserProfile{}, err
+	}
+
+	return profile, nil
+}
+
+func (c *client) FindProjectAppByClientAppID(groupIDs []string, appFlag string) ([]App, error) {
```
makesitgo, line 82: for the client, let's define/implement the following two methods for your purposes here:

```go
func (c *client) GetAppsByUser(groupIDs []string) ([]App, error) { /* iterate over groupIDs and collect FindAppsByGroupID */ }
func (c *client) GetAppsByGroupID(groupID string) ([]App, error) { /* lets fail early for now if anything goes wrong here */ }
// TODO: just for demo purposes, you won't need this one yet, no need to implement
func (c *client) GetApp(groupID, clientAppID string) (App, error) { /* ... */ }
```

this should be all you need on the `realm/client` side to get `app list` running

kpatel71716: Done! Commented out the tests that previously checked app id (though I can just remove if preferred). I have an if statement in the app list handler that calls GetApps if the project flag is provided. Let me know if that was not the intended solution here but I imagine that the same logic would be applied with the app flag as well

internal/cloud/realm/app.go
kpatel71716: Done!

internal/cloud/realm/client.go
kpatel71716: Discussed offline, done!

internal/cloud/realm/app.go
kpatel71716: Changed to fail hard, as per offline discussion + other comment

internal/cloud/realm/app.go
```diff
@@ -0,0 +1,77 @@
...
+				appSet[appElem.Name] = true
+			}
+		}
+	}
+	return appArr, nil
+}
+
+// GetApps fetches all Realm Apps associated with the given groupID
+func (c *client) GetApps(groupID string) ([]App, error) {
+	res, err := c.do(http.MethodGet, fmt.Sprintf(appsByGroupIDEndpoint, groupID), api.RequestOptions{UseAuth: true})
```
makesitgo, line 56: can we rename this `err` to something more unique as it sits in the entire function scope here?

internal/cloud/realm/app.go
```diff
@@ -0,0 +1,77 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		if res.StatusCode == http.StatusNotFound {
```
makesitgo, line 65: you can move this `if res.StatusCode == http.StatusNotFound` before the "not OK" check to de-nest things a bit

internal/cloud/realm/auth.go
```diff
@@ -12,35 +12,80 @@ type loginPayload struct {
 	PrivateAPIKey string `json:"apiKey"`
 }
 
-// AuthResponse is the Realm login response
-type AuthResponse struct {
+// Session is the Realm login response
```
makesitgo, line 15: [nit] `// Session is the Realm session`

internal/cloud/realm/auth.go
```diff
@@ -12,35 +12,80 @@ type loginPayload struct {
...
+
+// Role contains a GrouID field that maps to a project ID
+type Role struct {
+	GroupID string `json:"group_id"`
+}
+
+// AllGroupIDs returns all available group ids for a given user
+func (pd UserProfile) AllGroupIDs() []string {
+	groupIDs := []string{}
+	set := make(map[string]bool)
```
makesitgo, line 34: I prefer using `struct{}` instead of `bool` here as a value for a set map

also, lets just keep the initializations to 

```go
var arr []string
set := map[string]struct{}{}
```

internal/cloud/realm/auth.go
```diff
@@ -12,35 +12,80 @@ type loginPayload struct {
...
+	for _, role := range pd.Roles {
+		if !set[role.GroupID] {
```
makesitgo, line 36: [nit] lets rearrange this loop here:

```go
for _, role := range pd.Roles {
    if role.GroupID == "" {
        continue
    }
    if _, ok := set[role.GroupID]; ok {
        continue
    }
    arr = append(arr, role.GroupID)
    set[role.GroupID] = struct{}{}
}
```

internal/cloud/realm/app.go
```diff
@@ -0,0 +1,77 @@
...
+	ClientAppID string `json:"client_app_id"`
+	Name        string `json:"name"`
+}
+
+var (
+	errGroupNotFound = errors.New("group could not be found")
+)
+
+var (
+	appsByGroupIDEndpoint = adminAPI + "/groups/%s/apps"
```
makesitgo, line 25: [nit] lets rename this to `appsPathPattern`

internal/cloud/realm/app.go
```diff
@@ -0,0 +1,77 @@
...
+
+// GetApps fetches all Realm Apps associated with the given groupID
+func (c *client) GetApps(groupID string) ([]App, error) {
+	res, err := c.do(http.MethodGet, fmt.Sprintf(appsByGroupIDEndpoint, groupID), api.RequestOptions{UseAuth: true})
+
+	if err != nil {
+		return nil, err
+	}
+
+	defer res.Body.Close()
```
makesitgo, line 62: move this `defer` to the line right after initializing the json decoder

internal/cloud/realm/auth.go
```diff
@@ -12,35 +12,80 @@ type loginPayload struct {
...
+
+	var session Session
+	if err := dec.Decode(&session); err != nil {
+		return Session{}, err
 	}
+	return session, nil
+}
 
-	res, resErr := c.do(http.MethodPost, path, opts)
+func (c *client) GetUserProfile() (UserProfile, error) {
```
makesitgo, line 72: i'd like to consolidate the naming b/w `UserProfile` and `AuthProfile` (between path variables and response structures, it'd be nice to have a unified language here)

internal/cloud/realm/client.go
```diff
@@ -35,12 +58,36 @@ func (c *client) do(method, path string, options api.RequestOptions) (*http.Resp
...
 	req.Header.Set(requestOriginHeader, cliHeaderValue)
 
+	if options.ContentType != "" {
+		req.Header.Set(api.HeaderContentType, options.ContentType)
+	}
+
+	if auth, err := c.getAuth(options); err != nil {
+		return nil, err
+	} else if auth != "" {
+		req.Header.Set(api.HeaderAuthorization, "Bearer "+c.session.AccessToken)
```
makesitgo, line 71: you need to use `auth` instead of `c.session.AccessToken` here

makesitgo (commented):
did a full review here.... generally looks good to me, but i'll approve once we get the PR (its still a draft) and commits cleaned up, and evg passing

internal/cloud/realm/auth.go
kpatel71716: Done, change to AuthProfile since this logic lives in Auth

internal/cloud/realm/client.go
kpatel71716: 😬 Thanks, great catch

internal/cloud/realm/app.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+func (c *client) GetAppsForUser() ([]App, error) {
+	profile, err := c.GetAuthProfile()
+	if err != nil {
+		return nil, err
+	}
+	groupIDs := profile.AllGroupIDs()
+	var appArr []App
+	appArr = make([]App, 0)
```
makesitgo, line 35: this line isn't needed, you can immediately start appending to `var appArr []App` as is

also a nit, let's just call this `arr`, the type and context of this function makes `app` a bit redundant

internal/cloud/realm/app.go
```diff
@@ -0,0 +1,74 @@
...
+	for _, groupID := range groupIDs {
+		apps, err := c.GetApps(groupID)
+		if err != nil {
+			// Request will fail if any GetApps call fails
+			return nil, err
+		}
+		appSet := make(map[string]bool)
```
makesitgo, line 42: question: why are we making a set of apps based on name here?  Since we're just getting all apps per groupID now, we should probably just be returning all of those apps, regardless if they share a name.  i might be missing something here?

if we can get rid of this second for loop here, then the rest just becomes

```go
appArr = append(appArr, apps...)
```

internal/cloud/realm/app.go
```diff
@@ -0,0 +1,74 @@
...
+// App represents basic Realm App data
+type App struct {
+	ID          string `json:"_id"`
+	GroupID     string `json:"group_id"`
+	ClientAppID string `json:"client_app_id"`
+	Name        string `json:"name"`
+}
+
+var (
+	errGroupNotFound = errors.New("group could not be found")
```
makesitgo, line 21: for now let's just inline this error, I'm not sure where else we'd be using it yet

internal/cloud/realm/app.go
```diff
@@ -0,0 +1,74 @@
...
+func (c *client) GetApps(groupID string) ([]App, error) {
+	res, fetchAppsErr := c.do(http.MethodGet, fmt.Sprintf(appsPathPattern, groupID), api.RequestOptions{UseAuth: true})
+	if fetchAppsErr != nil {
+		return nil, fetchAppsErr
+	}
+	if res.StatusCode == http.StatusNotFound {
+		return nil, errGroupNotFound
+	}
+	if res.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("HTTP status error: %v", res.StatusCode)
```
makesitgo, line 63: this error return should just be: `return nil, UnmarshalServerError(res))

internal/cloud/realm/auth.go
```diff
@@ -12,35 +12,82 @@ type loginPayload struct {
...
+// Role contains a GrouID field that maps to a project ID
+type Role struct {
+	GroupID string `json:"group_id"`
+}
+
+// AllGroupIDs returns all available group ids for a given user
+func (pd AuthProfile) AllGroupIDs() []string {
+	var arr []string
+	set := map[string]struct{}{}
+	for _, role := range pd.Roles {
```
makesitgo, line 35: [nit] add a newline here above the `for` (or remove the newline after the `for` loop)

internal/cli/app/app_list_test.go
```diff
@@ -0,0 +1,170 @@
...
+		assert.Nil(t, err)
+		assert.NotNil(t, &cmd.realmClient)
+	})
+}
+
+func TestAppListHandler(t *testing.T) {
+	project1 := "project1"
+	project2 := "project2"
+
+	appName1 := "name1"
```
makesitgo, line 36: to help test code readability, can we try to omit the variables that do not need to be declared (in that they aren't used anywhere else downstream).  at a glance here, it looks like `appNameX` and all the `role` stuff can be inlined where its used

in general, I'd like to try to keep test code as direct and quick to read as possible.  every line counts :)

internal/cli/app/app_list_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/10gen/realm-cli/internal/utils/test/mock"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+func TestAppListSetup(t *testing.T) {
+	t.Run("Setup creates a realm client with a session", func(t *testing.T) {
+		config := cli.CommandConfig{RealmBaseURL: "http://localhost:8080"}
+		cmd := &appListCommand{}
+
+		authToken := "authToken"
```
makesitgo, line 19: likewise here, these strings can just be used inline

makesitgo (approved):
looks good to merge, there'll be some cleanup and tests (for the api methods) I add in a followup for the final refactor.  nice work here!

🆒  beans

internal/cloud/realm/app.go
kpatel71716: Good point, this set was redundant and the code looks a lot nicer now

internal/cloud/realm/client.go
```diff
@@ -35,12 +58,36 @@ func (c *client) do(method, path string, options api.RequestOptions) (*http.Resp
 		return nil, err
 	}
 
-	req.Header = options.Header
-	if req.Header == nil {
-		req.Header = http.Header{}
-	}
+	req.Header = http.Header{}
```
tsedgwick, line 61: I don't follow what this is doing, req.Header should already be created. Are we recreating it for a specific purpose?

internal/cloud/realm/client.go
```diff
@@ -16,17 +19,37 @@ const (
...
+	if err != nil {
+		return nil, err
+	}
+	options.Body = bytes.NewReader(body)
+	options.ContentType = api.MediaTypeApplicationJSON
+
+	return c.do(method, path, options)
 }
 
 func (c *client) do(method, path string, options api.RequestOptions) (*http.Response, error) {
```
tsedgwick, line 55: do we even need this func? seems like a little bit too much abstraction for this small codebase. I understand we do something similar in baas, but thoughts on removing this piece here? The most important piece in this func to me is the auth, and that's already abstracted into its own func.

tsedgwick (approved):
lgtm outside of very minor nits

internal/cloud/realm/client.go
makesitgo: I put this here, carryover from the previous codebase too.  my understanding was that since `http.Header` was a type alias of `map`, we needed to worry about `nil` maps before calling `Set`.  a few quick tests seems to confirm that is not the case...

internal/cloud/realm/client.go
makesitgo: I imagine we'd be using `c.do` quite a bit for any `GET`/`DELETE` requests, and it handles both setting auth and adding the `X-BAAS-Request-Origin` header.

to manage even just that for every API call we're making seems like a useful abstraction to me.  for what its worth, I see this function eventually expanding to handle some of the response handling as well (the first thing that comes to mind is when an `HttpUnauthorized` comes back, to either automatically attempt a refresh or clear the `profile`'s access token value (to indicate the session is no longer active)

internal/cloud/realm/client.go
kpatel71716: I just tested this out too, I will remove this

internal/cloud/realm/client.go
kpatel71716: I am fine with the current abstraction, but it may be worth revisiting further down the line as we expand the code as Nick mentioned

PR #36, opened 2020-12-10
REALMC-7157: Add user create command

makesitgo:
TODO:
- [x] write tests for `internal/cli/context.go`
- [x] write Setup tests for `internal/commands/app/list/command.go`
- [x] write tests for `internal/input/project_app.go`


internal/cli/context.go
```diff
@@ -0,0 +1,98 @@
...
+const (
+	appConfigJSON = "config.json"
+)
+
+var (
+	maxDirectoryContainSearchDepth = 8
+)
+
+// Context is the CLI context
+type Context struct {
```
tsedgwick, line 22: [q] I know we use Context in baas, but this gets confusing between the std lib context and our own implementation of context. is there a better name we can use? potentially Settings or even just adding this to the config directly?

internal/cli/context.go
```diff
@@ -0,0 +1,98 @@
...
+		return AppData{}, err
+	}
+	return appData, nil
+}
+
+// AppData is MongoDB Realm application data
+type AppData struct {
+	ID   string `json:"client_app_id"`
+	Name string `json:"name"`
+}
```
tsedgwick, line 98: [nit] can we move this struct to the top?

tsedgwick (commented):
mostly looks good, just one major question on my end

internal/cli/context.go
makesitgo: running out of good names here, I admit.  as it stands, the go std lib Context is not used in this repo right now, so there's no cross-contamination of names there.

this was initially `CommandConfig`, but "config"/"settings" doesn't feel right to me for this data. in my opinion, this data is more contextual than configurable

tsedgwick: since context is so embedded in the std lib, I would really not like to muddy the waters here. could we just remove this struct altogether and use these fields in line to where we use Context now?

UnicodeSnowman (approved):
👍

tsedgwick (approved):
lgtm!

PR #39, opened 2020-12-16
REALMC-7574: Create List feedback component

ermay12:


ermay12:
@makesitgo RFAL

internal/terminal/list_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"testing"
+
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestList(t *testing.T) {
+	assert.RegisterOpts(reflect.TypeOf(list{}), cmp.AllowUnexported(list{}))
+	message := "a list message"
```
makesitgo, line 14: `message` is shadowed here later on in the `list.Message()` test.  in light of that, lets shift these tests around a bit:

```go
func TestList(t *testing.T) {
  assert.RegisterOpts...
  t.Run("Should create a new list with its data parsed", func(t *testing.T) {
    list := newList(
      "a list message",
      []interface{}{ ... },
    )

    expectedList := list{ ... }
    assert.Equal(t, expectedList, list)

    t.Run("And Message should ...", func(t *testing.T){ /* can use the list var directly here */ })

    t.Run("And Payload should ...", func(t *testing.T){ /* can use the list var directly here */ })
  })
}
```

of course, you cold simply just rename this var to `listMessage` or something unique too, but it doesn't seem we use/reference this much in the first place, just mostly for creating the initial test list.  let me know what you think!

internal/terminal/table.go
```diff
@@ -110,7 +110,7 @@ func (t table) dataString() string {
 				strings.Repeat(" ", t.columnWidths[header]-len(row[header])),
 			)
 		}
-		rows[i] = strings.Join(cells, gutter)
+		rows[i] = "  " + strings.Join(cells, gutter)
```
makesitgo, line 113: this "two space" thing is used a lot of places (check out `gutter` in this file).

I think it'd be a good idea at this point to make `const gutter = "  " // gutter or spacer as a name?` and then use that in each place where appropriate.  (I also think we can get rid of that `strings.Repeat(" ", 2)` and just type out the string with two spaces)

internal/terminal/table.go
```diff
@@ -65,8 +65,8 @@ func (t table) Message() (string, error) {
 		return "", err
 	}
 	return fmt.Sprintf(`%s
-%s
-%s
+  %s
```
makesitgo, line 68: seeing as though we inject the spacer row-by-row for the `dataString`, let's do it the same way for `headerString` and `dividerString`

makesitgo (commented):
the code is there, this all looks good.  I'd like to add a few more tests though:

- [ ] `log_test` has test variants for each of our types so far, so let's include some for `List` now as well
- [ ]  this might be worth pulling `parseValue` into its own file, but lets start a unit test for that function so we can really hammer away at this conversion logic (rather than rely on writing complex tests in `table_test` or `list_test` to verify the same)

ermay12:
@makesitgo RFAL

internal/terminal/table.go
```diff
@@ -17,7 +17,7 @@ var (
 	tableFields = []string{logFieldMessage, logFieldData, logFieldHeaders}
 
 	// gutter is the gap between table columns
-	gutter = strings.Repeat(" ", 2)
+	gutter = "  "
```
makesitgo, line 20: [meganit] can move this now to a

```go
const (
  gutter = "  "
)
```
block

internal/terminal/table_test.go
```diff
@@ -256,3 +260,29 @@ func TestTablePayload(t *testing.T) {
 		assert.Equal(t, data, payloadData[logFieldData])
 	})
 }
+
+func TestParseValue(t *testing.T) {
+	t.Run("parseValue should create reasonable string representations", func(t *testing.T) {
```
makesitgo, line 265: can we convert this to a test table with descriptions for each?  something like:

```go
for _, tc := range []struct{

}{
  {"Should parse a nil value as blank string", nil, ""},
  // ...
}{
  t.Run(tc.description, func(t *testing.T){ ... })
}
```

makesitgo (commented):
just a few comments here, but I forgot to mention to do a project search for your ticket number here, there are a couple of `TODO`s related to this change as well

ermay12:
@makesitgo RFAL.  Also, I didn't see any mention of "7574" in the project.  Is there something I'm missing?

ermay12:
@makesitgo RFAL for real this time

internal/commands/app/list/command.go
```diff
@@ -45,7 +45,13 @@ func (cmd *command) Handler(profile *cli.Profile, ui terminal.UI) error {
...
 func (cmd *command) Feedback(profile *cli.Profile, ui terminal.UI) error {
-	return ui.Print(terminal.NewTextLog(fmt.Sprintf("results are: %v", cmd.apps)))
+	if len(cmd.apps) == 0 {
+		return ui.Print(terminal.NewTextLog("no available apps to show"))
+	}
+	apps := make([]interface{}, 0, len(cmd.apps))
+	for _, app := range cmd.apps {
+		apps = append(apps, app)
+	}
+	return ui.Print(terminal.NewListLog("available apps", apps...))
```
makesitgo, line 56: let's make this message `NewListLog(fmt.Sprintf("Found %d apps", len(apps)"), apps...)`

internal/terminal/table.go
```diff
@@ -11,13 +11,14 @@ import (
 const (
 	logFieldHeaders = "headers"
 	logFieldData    = "data"
+	// gutter is the gap between table columns
```
makesitgo, line 14: [meganit] add a newline between `logFieldData` and this comment here to break up the const block a little

internal/terminal/table_test.go
```diff
@@ -256,3 +260,90 @@ func TestTablePayload(t *testing.T) {
 		assert.Equal(t, data, payloadData[logFieldData])
 	})
 }
+
+func TestParseValue(t *testing.T) {
+
```
makesitgo, line 265: [nit] omit this newline

internal/terminal/table_test.go
```diff
@@ -256,3 +260,90 @@ func TestTablePayload(t *testing.T) {
...
+		{
+			description: "structs",
+			value: struct {
+				foo int
+				bar string
+			}{foo: 42, bar: "bar"},
+			expectedString: "{42 bar}",
+		},
+	} {
+		t.Run("parseValue should correctly parse "+tc.description, func(t *testing.T) {
```
makesitgo, line 340: [nit] instead of `"...should correctly parse..."` let's have these test descriptions say exactly what the assert is:

`"parseValue should parse 'a nil value as an empty string'"`
`"parseValue should parse 'an empty string as an empty string'"`

I know it seems like overkill, but really helps the reading test output if/when any error occurs.  the same comment applies for the "pointers" test below

makesitgo (approved):
this is ready to go pending my last few comments and evergreen.  really nice turnaround on this!

🆒  beans

PR #40, opened 2020-12-16
REALMC-7632: Add import and export methods to Realm client

makesitgo:
The purpose of these import/export implementations was primarily to support testing the other Realm client actions (instead of implementing a "CreateAuthProvider" function, which the CLI's client doesn't actually need, tests should leverage importing an app with the necessary auth providers included).

Some functionality for import/export is intentionally left out here, and should/will be filled in respectively once the push and pull command tickets are picked up.

This work does however enable to use fully test the Realm client `User` methods.

**Open to ideas on naming/organization here:** Does the `realm` package make sense for all the import/export related stuff: filepath-ing, config versions, payload structs that don't necessarily line up with the API response structure, etc?

### DIFF
--- a/internal/cloud/realm/export.go
+++ b/internal/cloud/realm/export.go
@@ -1,0 +1,80 @@
+package realm
+
+import (
+	"archive/zip"
+	"bytes"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"mime"
+	"net/http"
+
+	"github.com/10gen/realm-cli/internal/utils/api"
+)
+
+const (
+	exportPathPattern = appPathPattern + "/export"
+
+	exportQueryForSourceControl = "source_control"
+	exportQueryIsTemplated      = "template"
+	exportQueryVersion          = "version"
+
+	mediaParamFilename = "filename"
+
+	trueVal = "true"
+)
+
+// ExportRequest is a Realm application export request
+type ExportRequest struct {
+	ConfigVersion AppConfigVersion
+	IsTemplated   bool
+}
+
+func (c *client) Export(groupID, appID string, req ExportRequest) (string, *zip.Reader, error) {
+	options := api.RequestOptions{
+		UseAuth: true,
+		Query: map[string]string{
+			exportQueryVersion:          DefaultAppConfigVersion.String(),
+			exportQueryForSourceControl: trueVal,
+		},
+	}
+
+	if req.ConfigVersion != AppConfigVersionNil {
+		options.Query[exportQueryVersion] = req.ConfigVersion.String()
+	}
+	if req.IsTemplated {
+		options.Query[exportQueryIsTemplated] = trueVal
+	}
+
+	res, resErr := c.do(http.MethodGet, fmt.Sprintf(exportPathPattern, groupID, appID), options)
+	if resErr != nil {
+		return "", nil, resErr
+	}
+	if res.StatusCode != http.StatusOK {
+		defer res.Body.Close()
+		return "", nil, unmarshalServerError(res)
+	}
+
+	_, mediaParams, mediaErr := mime.ParseMediaType(res.Header.Get(api.HeaderContentDisposition))
+	if mediaErr != nil {
+		return "", nil, mediaErr
+	}
+
+	filename := mediaParams[mediaParamFilename]
+	if filename == "" {
+		return "", nil, errors.New("export response is missing filename")
+	}
+
+	defer res.Body.Close()
+	body, bodyErr := ioutil.ReadAll(res.Body)
+	if bodyErr != nil {
+		return "", nil, bodyErr
+	}
+
+	zipPkg, zipErr := zip.NewReader(bytes.NewReader(body), int64(len(body)))
+	if zipErr != nil {
+		return "", nil, zipErr
+	}
+
+	return filename, zipPkg, nil
+}
--- a/internal/cloud/realm/import.go
+++ b/internal/cloud/realm/import.go
@@ -1,0 +1,48 @@
+package realm
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/10gen/realm-cli/internal/utils/api"
+)
+
+const (
+	importPathPattern = appPathPattern + "/import"
+
+	importQueryStrategy = "strategy"
+
+	importStrategyReplaceByName = "replace-by-name"
+)
+
+// ImportRequest is a Realm application import request
+type ImportRequest struct {
+	ConfigVersion AppConfigVersion `json:"config_version"`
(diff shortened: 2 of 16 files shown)


internal/cloud/realm/export.go
```diff
@@ -0,0 +1,72 @@
...
+	if req.IsTemplated {
+		options.Query[exportQueryIsTemplated] = trueVal
+	}
+
+	res, resErr := c.do(http.MethodGet, fmt.Sprintf(exportPathPattern, groupID, appID), options)
+	if resErr != nil {
+		return "", nil, resErr
+	}
+	if res.StatusCode != http.StatusOK {
+		// res.Body.Close()
```
makesitgo, line 55: so this was carried over from the previous cli, but oddly this was causing an (extra? unrelated?) error: `http: read on closed response body` (I coerced the non-StatusOK response by sending an unsupported config version).

Which makes me want to ask the crowd: when should it be appropriate to call `res.Body.Close()`.  I think ideally its just handled with a single `defer res.Body.Close()` inside the closure of this function, but than forces the client here to perform extra processing on the zipped response.  I don't think the client should also be responsible for writing the zip to disk, but I can't think of other any alternatives to allow us to defer the `res.Body.Close()`

anyone have thoughts here?

UnicodeSnowman: hmm odd, that likely means something was still trying to read the body after it had been closed (the `UnmarshalServerError` constructor, perhaps?). In this case, that's my guess, and we might just have to close the body _after_ pulling the error out of the response. otherwise, any time you're returning a response body/`ReadCloser`, you're relying on the caller to manage closing out that resource when it's done with it (which seems fine to me). otherwise, a `defer res.Body.Close()` should do the trick so that everything gets cleaned up prior to return.

internal/cloud/realm/export.go
```diff
@@ -0,0 +1,72 @@
...
+		UseAuth: true,
+		Query: map[string]string{
+			exportQueryVersion: DefaultAppConfigVersion.String(),
+		},
+	}
+
+	if req.ConfigVersion != AppConfigVersionNil {
+		options.Query[exportQueryVersion] = req.ConfigVersion.String()
+	}
+	if req.ForSourceControl {
```
UnicodeSnowman, line 43: do we actually need/want this option? considering that we shouldn't be exporting resource `ObjectId`s anymore, I would assume we would just always want to use this option?

internal/cloud/realm/import.go
```diff
@@ -0,0 +1,47 @@
...
+			Query: map[string]string{
+				importQueryStrategy: importStrategyReplaceByName,
+			},
+		},
+	)
+	if resErr != nil {
+		return resErr
+	}
+	if res.StatusCode != http.StatusNoContent {
+		return UnmarshalServerError(res)
```
UnicodeSnowman, line 44: related, do we need a `defer res.Body.Close()` here?

internal/cloud/realm/import_export_test.go
```diff
@@ -0,0 +1,165 @@
...
+			})
+
+			t.Run("And the auth provider contents should be as expected", func(t *testing.T) {
+				apiKeyConfigFilepath := realm.FileAuthProvider("api-key")
+				apiKeyConfigPayload, apiKeyConfigOK := exported[apiKeyConfigFilepath]
+				assert.True(t, apiKeyConfigOK, "expected exported app to have file: %s", apiKeyConfigFilepath)
+
+				var apiKeyConfig map[string]interface{}
+				assert.Nil(t, json.Unmarshal([]byte(apiKeyConfigPayload), &apiKeyConfig))
+				assert.NotEqualf(t, "", apiKeyConfig["id"], "expected api-key.json to have non-blank 'id' field")
```
UnicodeSnowman, line 92: does this mean we're including `id`s?

internal/cloud/realm/realm.go
```diff
@@ -0,0 +1,41 @@
...
+
+// AppConfigVersion is the Realm application config version for import/export
+type AppConfigVersion int
+
+func (v AppConfigVersion) String() string { return strconv.Itoa(int(v)) }
+
+// set of supported app config versions
+const (
+	AppConfigVersionNil      AppConfigVersion = 0
+	AppConfigVersion20210101 AppConfigVersion = 20210101
```
UnicodeSnowman, line 22: 🎉  🎩  🍾

UnicodeSnowman (approved):
LGTM overall, just a few questions

internal/cloud/realm/export.go
makesitgo: You know I just silently removed the `defer ` when copying this over?  That's totally it, thank you for this.  I'll add the defer back and we should be good to go.  The caller will _still_ be responsible for closing the resource if/when a successful response is returned

internal/cloud/realm/export.go
makesitgo: yeah, that's a good point.  I'll remove it for now and hard-code us to use this option until we need it to be otherwise

internal/cloud/realm/import.go
makesitgo: yeah, so actually I put a `defer res.Body.Close()` at the start of `UnmarshalServerError`, b/c I'm not closing the response body anywhere I call this yet

internal/cloud/realm/import_export_test.go
makesitgo: Your suggestion around hard-coding "for source control" eliminates this now

internal/cloud/realm/import.go
UnicodeSnowman: that's _probably_ okay in this case, but I'm a bit hesitant to pass off the responsibility of closing a `ReadCloser` to a helper like this. it feels like it should still be the responsibility of the "owner" of that `ReadCloser` to clean things up.

makesitgo: I don't disagree about the general responsibility of things, but it also means that virtually everywhere we call `return unmarshalServerError(res)` we have to place `defer res.Body.Close()` above it.  and the one thing that bugs me there is we still aren't closing that response body on successful `Export`.

happy to revert this and be explicit with `defer res.Body.Close()` in all the `unmarshalServerError` spots.  the other change i'm including is not exporting `UnmarshalServerError` from the `realm` package, not sure why I needed that before.

UnicodeSnowman (approved):

internal/cloud/realm/import.go
makesitgo: **UPDATE:** took `defer res.Body.Close()` out of the helper, and made sure each client api call was properly calling `Close()` whenever body reads were attempted.  should be all set here

PR #41, opened 2020-12-17
REALMC-7167: Add app init command

makesitgo:


internal/cli/command_factory.go
```diff
@@ -153,7 +142,8 @@ func (factory *CommandFactory) Close() {
 // Run executes the command
 func (factory *CommandFactory) Run(cmd *cobra.Command) {
 	if err := cmd.Execute(); err != nil {
-		if _, ok := err.(errSupressUsage); !ok {
+		cause, isUsagePrinter := errors.Unwrap(err).(UsagePrinter)
```
tsedgwick, line 145: [nit] it's more idiomatic to use `ok` here instead of the `isUsagePrinter`

internal/cli/fs.go
```diff
@@ -0,0 +1,56 @@
...
+	if _, err := io.Copy(f, r); err != nil {
+		return fmt.Errorf("failed to write file at %s: %w", path, err)
+	}
+	return nil
+}
+
+// WriteZip writes the zip contents to the specified filepath
+func WriteZip(basePath string, zipPkg *zip.Reader) error {
+	for _, zipFile := range zipPkg.File {
+		fmt.Println("opening zip file:", basePath, zipFile.Name)
```
tsedgwick, line 36: 👋🏻

internal/cli/fs.go
```diff
@@ -0,0 +1,56 @@
...
+		path := filepath.Join(basePath, zipFile.Name)
+
+		if zipFile.FileInfo().IsDir() {
+			if err := Mkdir(path); err != nil {
+				return err
+			}
+		} else {
```
tsedgwick, line 43: [nit] could just `return Mkdir(path)` here or at the very least return early here and drop the else

internal/commands/app/initialize/command.go
```diff
@@ -0,0 +1,97 @@
...
+func (cmd *command) initializeFromApp(wd, groupID, appID string) error {
+	_, zipPkg, exportErr := cmd.realmClient.Export(groupID, appID, realm.ExportRequest{IsTemplated: true})
+	if exportErr != nil {
+		return exportErr
+	}
+	return cli.WriteZip(wd, zipPkg)
+}
+
+func (cmd *command) initializeFromTemplate(wd, templateID string) error {
+	// TODO(REALMC-XXXX): implement Realm app templates
```
tsedgwick, line 95: can we get the real ticket here?

internal/commands/app/initialize/command.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	return cmd.initialize(profile.WorkingDirectory)
+}
+
+func (cmd *command) Feedback(profile *cli.Profile, ui terminal.UI) error {
+	return ui.Print(terminal.NewTextLog("Successfully initialized app"))
+}
+
+func (cmd *command) initialize(wd string) error {
+	data := []byte(fmt.Sprintf(`{
```
tsedgwick, line 68: what was the reasoning here between using a struct / json marshalling vs doing this? I feel like a struct is less likely for us to make a mistake typing this out.

internal/commands/app/initialize/command_test.go
```diff
@@ -0,0 +1,153 @@
...
+			},
+			realmClient: client,
+		}
+
+		assert.Nil(t, cmd.Handler(profile, nil))
+
+		t.Run("Should have the expected contents in the app config file", func(t *testing.T) {
+			config, err := ioutil.ReadFile(filepath.Join(profile.WorkingDirectory, realm.FileAppConfig))
+			assert.Nil(t, err)
+			assert.Equal(t, `{
```
tsedgwick, line 100: if we go with a struct above, we can test that here as well

internal/commands/app/initialize/errors.go
```diff
@@ -0,0 +1,7 @@
+package initialize
+
+type errProjectExists struct{}
+
+func (err errProjectExists) Error() string { return "a project already exists" }
```
tsedgwick, line 5: will we need any additional contextual information for this error?

internal/commands/app/initialize/inputs.go
```diff
@@ -0,0 +1,100 @@
...
+		return appDataErr
+	}
+	if appData.Name != "" {
+		return errProjectExists{}
+	}
+
+	if i.From != "" {
+		i.FromType = determineFromType(i.From)
+
+		// TODO(REALMC-XXXX): implement Realm app templates
```
tsedgwick, line 45: can we get the realm ticket for this? 

also - is the commented out code below for that ticket implementation?

internal/commands/app/initialize/inputs.go
```diff
@@ -0,0 +1,100 @@
...
+		// if err := ui.AskOne(&i.FromType, &survey.Select{
+		// 	Message: "Initialize app from what?",
+		// 	Options: []string{fromApp, fromTemplate},
+		// }); err != nil {
+		// 	return err
+		// }
+	} else {
```
tsedgwick, line 52: [nit] once we implement the log above, we can return early in that if block and drop this else

internal/commands/app/initialize/inputs.go
```diff
@@ -0,0 +1,100 @@
...
+	switch i.FromType {
+	case fromApp:
+		app, err := cli.ResolveApp(ui, client, realm.AppFilter{GroupID: i.Project, App: i.From})
+		if err != nil {
+			return from{}, err
+		}
+		f.GroupID = app.GroupID
+		f.AppID = app.ID
+	case fromTemplate:
+		// do nothing
```
tsedgwick, line 80: 👀

tsedgwick (approved):
mostly nits from me, otherwise lgtm!

internal/cli/command_factory.go
makesitgo: 👍 ... I generally go with these names when the `var` isn't confined to a very tight closure, so to avoid any potential
 (now or future) downstream shadowing of `ok`.  but i'll change it here (there's no shadowing, little chance this function/closure grows much more)

internal/cli/fs.go
makesitgo: not sure I follow, this is done inside a for loop of all the zip's files.  I _think_ we still want to be iterating through the rest of the package, right?

internal/commands/app/initialize/command.go
makesitgo: I'll follow up with drew, but I don't even think this work is planned yet (if there's even an epic for this yet).  If it makes more sense, I'll rip all this commented code out, but I wanted to at least prove to myself this structure would support this pattern.

theres a good conversation around this stuff on the spec here: https://docs.google.com/document/d/15Y0_jCxU2OOVx5sspQZjdewaUbFtkvWsxZUDrnkk0vc/edit?disco=AAAAKWWDkRU

internal/commands/app/initialize/command.go
makesitgo: simplicity? no real reasoning...we don't rely on knowing the structure anywhere else so I just went for the string payload here.  In general, we won't care much about a lot of the various file contents/structures, though I can see the argument being made fairly quickly that we do care at least for this app/root `config.json` while we attempt to pull app information out of it (id, name) and/or merge data into it (when creating from template or existing app).

i'll add it as a struct with json tags and work with it from that perspective

internal/commands/app/initialize/errors.go
makesitgo: i hope not? though that's likely answered in practice.  it _should_ be showing up as a fatal error that terminates command execution and says "ERROR app init failed: a project already exists"

we have some work currently scheduled that allow for errors to associate followup commands and/or documentation links that this would probably benefit from ("Try running the same command inside an empty directory" or something along those lines)

internal/commands/app/initialize/inputs.go
makesitgo: commented out code is for the implementation, of which I'm not sure when it will actually be planned (I think "app templates" will be its own project/epic).

I mentioned this above, but I can rip all the commented code out if you feel strongly about that.  This was primarily to help prove to myself that this command structure _could_ support some flavor of a app/templates flag (see the spec conversation I linked above for more details there)

internal/commands/app/initialize/inputs.go
makesitgo: i prefer these types of comments just to remind future devs the empty case statement is in fact not an error

internal/cli/command_factory.go
```diff
@@ -153,7 +142,8 @@ func (factory *CommandFactory) Close() {
 // Run executes the command
 func (factory *CommandFactory) Run(cmd *cobra.Command) {
 	if err := cmd.Execute(); err != nil {
-		if _, ok := err.(errSupressUsage); !ok {
+		cause, ok := errors.Unwrap(err).(UsagePrinter)
+		if !ok || cause.PrintUsage() {
```
tsedgwick, line 146: the way we handle PrintUsage is a little confusing as returning false seems like an antipattern. Since false is the zero value of a bool, would it make more sense to flip this to `DisablePrintUsage`.

internal/cli/fs.go
tsedgwick: yes you're right. I meant to do a continue here instead of a return.

```if zipFile.FileInfo().IsDir() {
			if err := Mkdir(path); err != nil {
				return err
			}
		}
continue
```
that way we can drop the else and the nested block

internal/commands/app/initialize/command.go
tsedgwick: to stick to our policy of no TODOs without valid jira tickets, let's remove this if we don't have this planned.

internal/commands/app/initialize/errors.go
tsedgwick: depending on how we refactor PrintUsage, this might not be needed as its own error struct.

internal/commands/app/initialize/inputs.go
tsedgwick: same as above

tsedgwick (commented):
overall looks good. I'd really like to address that PrintUsage though before merging this.

internal/cli/fs.go
makesitgo: roger that

internal/commands/app/initialize/command.go
makesitgo: added the branch [REALMC-4543_TemplateStarterApplications]https://github.com/makesitgo/realm-cli/tree/REALMC-4543_TemplateStarterApplications) to my fork with the commented code in tact

removed/simplified that code [here](https://github.com/10gen/realm-cli/pull/41/commits/39ef8093387b0fc18cd156bd525f71cfc730a348)

internal/commands/app/initialize/inputs.go
makesitgo: this has gone away

internal/commands/app/initialize/errors.go
makesitgo: Moved `PrintUsage` to `DisableUsage` in the refactored approach

PR #43, opened 2020-12-23
REALMC-7573 attempt auth refresh when an HttpUnauthorized is returned

kpatel71716:


internal/auth/session.go
```diff
@@ -0,0 +1,7 @@
+package auth
+
+// Session is the Realm session
+type Session struct {
```
kpatel71716, line 4: I pulled Session into the auth package and as discussed earlier, if we want to stay with this approach, there are a couple of other items I can pull out as well. We could add just one auth.go file and move structs, vars here. Or we could just pull the whole internal/cloud/auth.go file here

internal/cloud/realm/client.go
```diff
@@ -91,5 +93,52 @@ func (c *client) do(method, path string, options api.RequestOptions) (*http.Resp
...
+	}
+	req.Header.Set(api.HeaderAuthorization, "Bearer "+authToken)
+	// TODO REALMC-7580 change these lines to just return client.Do(req) as session will be cleared downstream
+	res, err = client.Do(req)
+	if res.StatusCode == http.StatusUnauthorized {
+		c.profile.ClearSession()
+		return nil, err
+	}
+	return res, err
+}
```
kpatel71716, line 114: I had the code this way, instead of another method on client because it felt like the simplest approach for avoiding an infinite loop of retries without having to pass in some information to know to stop retrying if the retry call keeps giving 401s

internal/profile/profile.go
```diff
@@ -1,10 +1,10 @@
-package cli
+package profile
```
kpatel71716, line 1: This felt a little dirty but I had to move profile out of cli due to dependency cycles. It is a bit weird because of how much profile is a part of the components in the cli package but I couldn't find a better way for the client to have save the new auth token to the profile

internal/cloud/realm/client.go
```diff
@@ -45,13 +47,13 @@ func NewClient(baseURL string) Client {
...
-func NewAuthClient(baseURL string, session Session) Client {
-	return &client{baseURL, session}
+func NewAuthClient(baseURL string, profile *profile.Profile) Client {
+	return &client{baseURL, profile}
 }
 
 type client struct {
 	baseURL string
-	session Session
+	profile *profile.Profile
```
kpatel71716, line 56: since we get the baseURL from the client, we could just have profile in the client struct. It does raise the question of how necessary it is to have a client that wraps around a single value. But I do like the notion of having the doJson, do, and other methods on client rather than Profile

PR #45, opened 2021-01-05
REALMC-7573 add client refresh logic

kpatel71716:


internal/cli/command_factory.go
```diff
@@ -106,6 +107,13 @@ func (factory *CommandFactory) Build(command CommandDefinition) *cobra.Command {
 			factory.telemetryService.TrackEvent(telemetry.EventTypeCommandStart)
 
 			err := command.Command.Handler(factory.profile, factory.ui)
+			if err == realm.ErrInvalidSession {
+				factory.profile.ClearSession()
+				profileErr := factory.profile.Save()
+				if profileErr != nil {
+					factory.errLogger.Fatal(fmt.Errorf("failed to clear session: %s", profileErr.Error()))
```
kpatel71716, line 114: I wasn't sure of the _best_ spot to save the profile, but I placed it above the error check and added a log in case the profile does file to save for whatever reason. I imagine like we can reduce the log level too, considering that this isn't a big issue. Maybe an info?

internal/cloud/realm/client.go
```diff
@@ -91,5 +91,56 @@ func (c *client) do(method, path string, options api.RequestOptions) (*http.Resp
...
+		return nil, refreshErr
+	}
+	// TODO: REALMC-7719 save the new access token to prevent unnecessary retries
+	c.session.AccessToken = authToken
+	options.PreventRefresh = true
+
+	return c.do(method, path, options)
+}
+
+func (c *client) refreshAuth(httpClient *http.Client) (string, error) {
```
kpatel71716, line 118: open to changing the signature or moving this function into the auth package. I felt this function was for internal client behavior so I left it here and re-used the httpClient

internal/cloud/realm/client_test.go
```diff
@@ -0,0 +1,39 @@
+package realm_test
```
kpatel71716, line 1: open to adding whatever tests we need here

internal/cloud/realm/client.go
```diff
@@ -91,5 +91,56 @@ func (c *client) do(method, path string, options api.RequestOptions) (*http.Resp
 		req.Header.Set(api.HeaderAuthorization, "Bearer "+auth)
 	}
 	client := &http.Client{}
-	return client.Do(req)
+	res, err := client.Do(req)
+	if err != nil {
+		return res, err
```
makesitgo, line 96: [meganit] could just return `nil, err` here

internal/cloud/realm/client.go
```diff
@@ -91,5 +91,56 @@ func (c *client) do(method, path string, options api.RequestOptions) (*http.Resp
...
+	}
+	if res.StatusCode != http.StatusUnauthorized || !options.UseAuth || options.PreventRefresh {
```
makesitgo, line 98: let's change this check to just see if `res.StatusCode` is in between 200 and 299, and if it is, then we can just return `res, nil` here

internal/cloud/realm/client.go
```diff
@@ -91,5 +91,56 @@ func (c *client) do(method, path string, options api.RequestOptions) (*http.Resp
...
+		return res, err
+	}
+
+	serverError := unmarshalServerError(res).(ServerError)
```
makesitgo, line 102: if we make it to this point, we know the `err` from `client.Do` is nil but our response status code was not successful, at this point we should _try_ to unmarshal the server error

i think we'll need to update the `unmarshServerError` to more gracefully handle any type of response.  you should check how we actually send this error from our server, but I imagine we have some `content-type` header present to indicate a JSON body.  if we see that, unmarshal the error, check for `InvalidSessionCode` and either refresh auth and retry the request or simply return the server error

if we don't see the `content-type` as JSON, then we should construct a generic error (`errors.New(res.Status)` or something like that) and return that instead.  this is a similar flow to how we do it in [baas-admin-sdk](https://github.com/10gen/baas-admin-sdk/blob/master/src/Client.ts#L199-L238)

internal/cloud/realm/client.go
makesitgo: could this refreshAuth/refreshSession method implementation leverage `func (c *client) do` and take care of a lot of this code by just specifying the proper `RequestOptions`?

internal/cloud/realm/error.go
```diff
@@ -6,6 +6,11 @@ import (
 	"net/http"
 )
 
+const (
+	// InvalidSessionCode is the error code returned when the user's sesison is invalid
+	InvalidSessionCode = "InvalidSession"
```
makesitgo, line 11: does this need to be exported?

internal/cloud/realm/client_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	u "github.com/10gen/realm-cli/internal/utils/test"
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+)
+
+func TestRealmClient(t *testing.T) {
+	u.SkipUnlessRealmServerRunning(t)
+
+	t.Run("Should work correctly execute request with valid credentials", func(t *testing.T) {
```
makesitgo, line 14: this test feels similar to the one defined in `TestRealmAuthProfile`.  in addition to potentially changing the implementation of `c.refreshAuth` to leverage `c.do`,  I think we should probably move that function defintion to the `auth.go` file in this package.  thoughts?

internal/cloud/realm/client_test.go
```diff
@@ -0,0 +1,39 @@
...
+		client := newAuthClient(t)
+		authProfile, err := client.AuthProfile()
+		assert.Equal(t, nil, err)
+		assert.NotEqualf(t, realm.AuthProfile{}, authProfile, "expected auth profile to not be empty")
+
+	})
+
+	t.Run("Should return the correct error when credentials are invalid", func(t *testing.T) {
```
makesitgo, line 22: can we add a test that asserts we see the `ErrInvalidSession` is properly returned somewhere?

internal/cloud/realm/client_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+		client = realm.NewAuthClient(u.RealmServerURL(), session)
+		_, err = client.AuthProfile()
+
+		serverError := err.(realm.ServerError)
+		assert.Equal(t, "401 Unauthorized", serverError.Message)
+
+	})
+
+	// TODO: REALMC-7719 add test for expired credentials
```
makesitgo, line 38: [nit] use the TODO format of `// TODO(REALMC-7719): add test for expired credentials`

internal/cli/command_factory.go
makesitgo: I think this is fine here, we shouldn't be making any api calls from the `Setup` or `Feedback` stages

and yeah, we don't want to use `errLogger.Fatal` for a failure to save the profile after attempting to clear the session.  the error that still shows to the user is "invalid session" and they'll still be prompted to re-login (effectively clearing their session at that point).

in the interest of not having unhandled errors, and considering we're in the `RunE` phase which means we know we've ensured our `ui` exists, we should just use that to print a text message with a WARN severity (you'll have to add that severity)

makesitgo (commented):
first batch of comments here, this is lookin good so far.  nice work!

internal/cli/command_factory.go
```diff
@@ -111,7 +114,7 @@ func (factory *CommandFactory) Build(command CommandDefinition) *cobra.Command {
 				factory.profile.ClearSession()
 				profileErr := factory.profile.Save()
 				if profileErr != nil {
-					factory.errLogger.Fatal(fmt.Errorf("failed to clear session: %s", profileErr.Error()))
+					factory.warnLogger.Printf("failed to clear session: %s", profileErr.Error())
```
makesitgo, line 117: we should use `factory.ui.Print(terminal.NewTextLog("..."))` here.  the one catch is we still don't have a WARN level severity, which you could add to the `terminal` package and maybe create a `terminal.NewWarning` function (which does the same thing as NewTextLog but uses a different LogLevel

internal/cloud/realm/auth.go
```diff
@@ -97,6 +97,25 @@ func (c *client) getAuth(options api.RequestOptions) (string, error) {
 	return "", nil
 }
 
+//Refresh Auth uses the user's refresh token to attempt to create a new access token
+func (c *client) refreshAuth() (string, error) {
+	res, resErr := c.do(http.MethodPost, refreshAuthPath, api.RequestOptions{RefreshAuth: true})
+
```
makesitgo, line 103: omit this newline

internal/cloud/realm/auth.go
```diff
@@ -97,6 +97,25 @@ func (c *client) getAuth(options api.RequestOptions) (string, error) {
...
+	if res.StatusCode != http.StatusCreated {
+		return "", ErrInvalidSession
+	}
+	defer res.Body.Close()
+
+	var session Session
+	if err := json.NewDecoder(res.Body).Decode(&session); err != nil {
+		return "", err
+	}
+	return session.AccessToken, nil
```
makesitgo, line 116: I think we want this to return the session, ultimately once we're able to save the new details on the profile we'll probably want to save both access and refresh tokens

internal/cloud/realm/client.go
```diff
@@ -93,54 +93,27 @@ func (c *client) do(method, path string, options api.RequestOptions) (*http.Resp
...
-	if res.StatusCode != http.StatusUnauthorized || !options.UseAuth || options.PreventRefresh {
-		return res, err
+	if res.StatusCode >= 200 && res.StatusCode <= 299 {
+		return res, nil
 	}
 
-	serverError := unmarshalServerError(res).(ServerError)
-	if serverError.Code != InvalidSessionCode {
-		return res, err
+	serverError, ok := unmarshalServerError(res).(ServerError)
```
makesitgo, line 102: we shouldn't be returning `res, nil` down below at all.  once we're at this point in the code, we know we don't have a successful response so _some_ error needs to be returned.  a couple of changes here:

- it probably makes sense to change the existing name: `unmarshalServerError` to something like `parseResponseError`
- **IF** the error is not of type `ServerError` **OR** its error code is not an `invalidSessionCode`, then simply return `nil, parsedErr`
- **ELSE** do the retry logic, and if an error comes from there we should return an invalid session error

something like:

```go
parsedErr := parseResponseError(res)
if err, ok := parsedErr.(ServerError); !ok {
  return nil, parsedErr
} else if err.Code != invalidSessionCode {
  return nil, e
}
session, sessionErr := c.refreshAuth()
if sessionErr != nil {
  return nil, ErrInvalidSession
}
c.session = session
options.PreventRefresh = true
return c.do(method, path, options)
```

makesitgo (commented):
couple more changes here

internal/cloud/realm/auth.go
kpatel71716: The session endpoint appears to only return the access token (unless I completely whiffed on something). Although I can see the confusion since we reuse the session object here. If there is a way to get the refresh token, then I can make that fix

internal/cloud/realm/client.go
kpatel71716: Ah yeah this is a really good point. Will make that change

internal/cloud/realm/auth.go
makesitgo: ah gotcha, yeah that makes sense.  feel free to leave as is!

internal/cli/command_factory.go
kpatel71716: Done! I added a nolint line over the factory print to ignore the error. Wasn't sure of a cleaner alternative since we want to prioritize the caught upstream error in this scenario

internal/cli/command_factory.go
```diff
@@ -106,6 +107,14 @@ func (factory *CommandFactory) Build(command CommandDefinition) *cobra.Command {
 			factory.telemetryService.TrackEvent(telemetry.EventTypeCommandStart)
 
 			err := command.Command.Handler(factory.profile, factory.ui)
+			if err == realm.ErrInvalidSession {
+				factory.profile.ClearSession()
+				profileErr := factory.profile.Save()
+				if profileErr != nil {
+					//nolint:errcheck
```
makesitgo, line 114: instead of the `nolint` here, you can print the original `profileErr` with `factory.errLogger.Printf` and then proceed along

internal/cloud/realm/auth.go
```diff
@@ -11,6 +11,7 @@ import (
 const (
 	authenticatePath = adminAPI + "/auth/providers/mongodb-cloud/login"
 	authProfilePath  = adminAPI + "/auth/profile"
+	refreshAuthPath  = adminAPI + "/auth/session"
```
makesitgo, line 14: [nit] call this `authSessionPath`

internal/cloud/realm/auth.go
```diff
@@ -96,6 +97,24 @@ func (c *client) getAuth(options api.RequestOptions) (string, error) {
 	return "", nil
 }
 
+//Refresh Auth uses the user's refresh token to attempt to create a new access token
```
makesitgo, line 100: docstring can be omitted, or otherwise it should be cleaned up (i'd just remove it)

```go
// refreshAuth uses the refresh token to create a new access token
```

internal/cloud/realm/auth.go
```diff
@@ -96,6 +97,24 @@ func (c *client) getAuth(options api.RequestOptions) (string, error) {
...
+func (c *client) refreshAuth() (string, error) {
+	res, resErr := c.do(http.MethodPost, refreshAuthPath, api.RequestOptions{RefreshAuth: true})
```
makesitgo, line 102: does this also want `PreventRefresh: true`?

internal/cloud/realm/client.go
```diff
@@ -91,5 +91,32 @@ func (c *client) do(method, path string, options api.RequestOptions) (*http.Resp
 		req.Header.Set(api.HeaderAuthorization, "Bearer "+auth)
 	}
 	client := &http.Client{}
-	return client.Do(req)
+	res, err := client.Do(req)
```
makesitgo, line 94: [nit] change `err` to `resErr` as it sits in the larger function scope

internal/cloud/realm/client.go
```diff
@@ -91,5 +91,32 @@ func (c *client) do(method, path string, options api.RequestOptions) (*http.Resp
...
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode >= 200 && res.StatusCode <= 299 {
+		return res, nil
+	}
+
+	parsedErr := parseResponseError(res)
+	if err, ok := parsedErr.(ServerError); !ok {
```
makesitgo, line 103: this is still missing the mark a bit, we need this:

```go
parsedErr := parseResponseError(res)
if err, ok := parsedErr.(ServerError); !ok {
  return nil, parsedErr
}
if options.PreventRefresh || err.Code != invalidSessionCode {
  return nil, err
}
```

internal/cloud/realm/client.go
```diff
@@ -91,5 +91,32 @@ func (c *client) do(method, path string, options api.RequestOptions) (*http.Resp
...
+		return nil, parsedErr
+	}
+
+	if options.PreventRefresh {
+		return nil, parsedErr
+	}
+
+	authToken, refreshErr := c.refreshAuth()
+	if refreshErr != nil {
+		return nil, refreshErr
```
makesitgo, line 115: `return nil, ErrInvalidSession` here.  if _anything_ fails in the `refreshAuth` method, we'll chalk it up to "unable to refresh your access token" and clear out your session

internal/cloud/realm/error.go
```diff
@@ -3,7 +3,15 @@ package realm
...
 	"bytes"
 	"encoding/json"
+	"errors"
 	"net/http"
+
+	"github.com/10gen/realm-cli/internal/utils/api"
+)
+
+const (
+	// InvalidSessionCode is the error code returned when the user's sesison is invalid
```
makesitgo, line 13: docstring can be removed for an unexported const (i'd remove, but if you want to keep it then do some cleanup)

```go
// invalidSessionCode is the server error code for an invalid session
```

internal/cloud/realm/error.go
```diff
@@ -16,9 +24,12 @@ func (se ServerError) Error() string {
 	return se.Message
 }
 
-// unmarshalServerError attempts to read and unmarshal a server error
+// parseResponseError attempts to read and unmarshal a server error
 // from the provided *http.Response
```
makesitgo, line 28: can you add on to this docstring?  something like:

```
// from the provided *http.Response.  a generic error will be returned if unable to parse the response body.
```

internal/cloud/realm/client.go
```diff
@@ -91,5 +91,32 @@ func (c *client) do(method, path string, options api.RequestOptions) (*http.Resp
...
 	client := &http.Client{}
-	return client.Do(req)
+	res, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode >= 200 && res.StatusCode <= 299 {
+		return res, nil
+	}
+
```
makesitgo, line 101: need a `defer res.Body.Close()` here

internal/terminal/log.go
```diff
@@ -75,6 +76,16 @@ func NewErrorLog(err error) Log {
 	return newLog(LogLevelError, errorMessage{err})
 }
 
+// NewWarningLog creates a new warn log
```
makesitgo, line 79: [nit] "...creates a new *warning* log"

internal/cloud/realm/auth.go
```diff
@@ -34,14 +35,14 @@ func (c *client) Authenticate(publicAPIKey, privateAPIKey string) (Session, erro
 		http.MethodPost,
 		authenticatePath,
 		authenticateRequest{publicAPIKey, privateAPIKey},
-		api.RequestOptions{},
+		api.RequestOptions{PreventRefresh: true},
```
kpatel71716, line 38: without this param, there are times we hide a more descriptive invalid session message. I felt that this is the one call where it would be fine to not retry. And that way we can return "failed to authenticate with MongoDB Cloud API: You are not authorized for this resource." instead of a generic "invalid session" message if the login credentials are invalid

internal/terminal/log.go
```diff
@@ -75,6 +76,16 @@ func NewErrorLog(err error) Log {
...
+func NewWarningLog(format string, args ...interface{}) Log {
+	message := format
```
makesitgo, line 81: might be worth extracting this to an unexported helper func within this file (share it b/w textlog and warninglog)

makesitgo (commented):
just a last bit of cleanup, and then i'll look over the tests and all.  nearly there!

internal/cloud/realm/auth.go
kpatel71716: That is a great call out. I believe we don't get an InvalidSession code from this call but I am lot more comfortable adding it

internal/cloud/realm/client.go
kpatel71716: Yep, that was my bad from moving stuff around. Should fix the test failures too

makesitgo (commented):
talked offline, but posting here for reference

internal/cloud/realm/error.go
```diff
@@ -16,9 +20,9 @@ func (se ServerError) Error() string {
 	return se.Message
 }
 
-// unmarshalServerError attempts to read and unmarshal a server error
-// from the provided *http.Response
-func unmarshalServerError(res *http.Response) error {
+// parseResponseError attempts to read and unmarshal a server error
+// from the provided *http.Response.
```
makesitgo, line 24: [meganit] remove the period

internal/cloud/realm/auth_test.go
```diff
@@ -49,6 +49,41 @@ func TestRealmAuthProfile(t *testing.T) {
...
+		session, err := client.Authenticate(u.CloudUsername(), u.CloudAPIKey())
+		assert.Equal(t, nil, err)
+
+		// invalidate the access token
+		session.AccessToken = session.RefreshToken
+
+		client = realm.NewAuthClient(u.RealmServerURL(), session)
+		_, err = client.AuthProfile()
+		serverError := err.(realm.ServerError)
+		assert.Equal(t, "invalid session: valid Issuer required", serverError.Message)
```
makesitgo, line 67: [meganit] any chance we could get this assert to check Equal on the `ServerError` struct itself here?

makesitgo (approved):
nice stuff, just a last few things and make sure evergreen and the commit history are all clean

🆒 beans

PR #46, opened 2021-01-06
REALMC 7158: Buildout command user list

ermay12:


ermay12:
@makesitgo a little confused about some of the details like the userID multiselect.  Does that mean it should fetch every filtered user, and then have the user select which users they want to list?

makesitgo:
great point, that's a copy/paste error on my end.  The `--user` flag here doesn't need a prompt.  For now, let's keep it as a string array flag, and we'll assume the values provided are valid user ids to include in the client filter

internal/commands/user/list/command.go
```diff
@@ -13,8 +20,106 @@ var Command = cli.CommandDefinition{
...
+	var logs []terminal.Log
+	for provider, users := range usersByProvider {
+		sort.Slice(
+			users,
+			func(i, j int) bool {
+				return users[i].LastAuthenticationDate < users[j].LastAuthenticationDate
+			},
+		)
+
+		userTable := make([]map[string]interface{}, 0, len(users))
```
makesitgo, line 99: talked offline on the approach here, should be straightforward enough to implement for both `api-key` and `userpass` (we might want to figure out a common place to define those values, maybe `cloud/realm/auth_provider.go`) as the same things (I think) are used in `user create` flags

makesitgo (commented):
lookin good so far!  couple of comments here and a few tests to write as well

internal/commands/user/list/command.go
ermay12: Sorry, I just saw this comment.  Can you clarify what you are saying here?  I think I might have misunderstood the offline conversation.

ermay12:
@makesitgo RFAL

ermay12:
@makesitgo RFAL

internal/commands/user/list/command.go
```diff
@@ -13,8 +20,129 @@ var Command = cli.CommandDefinition{
...
+		for _, identity := range user.Identities {
+			usersByProvider[identity.ProviderType] = append(usersByProvider[identity.ProviderType], user)
+		}
+	}
+
+	var logs []terminal.Log
+	for provider, users := range usersByProvider {
+		sort.Slice(
+			users,
+			func(i, j int) bool {
```
makesitgo, line 95: let's extract this `func` out to an unexported one in this file, something like `sortUsersByLastAuthentication` or something

internal/commands/user/list/command.go
```diff
@@ -13,8 +20,129 @@ var Command = cli.CommandDefinition{
...
+		headerID,
+		headerEnabled,
+		headerType,
+		headerLastAuthenticationDate,
+	)
+	return headers
+}
+
+func userTableRow(providerType string, user realm.User) map[string]interface{} {
+	timeString := ""
```
makesitgo, line 131: can we rename this to just `lastAuthenticationDate`, and maybe have it initialized to `"n/a"` or something similar (e.g. `"--"`, etc.)

internal/commands/user/list/command.go
```diff
@@ -13,8 +20,129 @@ var Command = cli.CommandDefinition{
...
+	}
+	row := map[string]interface{}{
+		headerID:                     user.ID,
+		headerEnabled:                !user.Disabled,
+		headerType:                   user.Type,
+		headerLastAuthenticationDate: timeString,
+	}
+	switch providerType {
+	case providerTypeAPIKey:
+		row[headerName] = user.Data["name"]
```
makesitgo, line 143: can we make `"name"` and `"email"` unexported consts here?  `userDataName` and `userDataEmail` respectively

internal/commands/user/list/command_test.go
```diff
@@ -0,0 +1,192 @@
+package list
+
+import (
+	//"errors"
+
+	//"strings"
```
makesitgo, line 6: can remove these commented imports

internal/commands/user/list/command_test.go
```diff
@@ -0,0 +1,192 @@
...
+			testApp.ID = filter.App
+			return []realm.App{testApp}, nil
+		}
+		return realmClient
+	}
+
+	t.Run("Should find app users", func(t *testing.T) {
+		testUsers := []realm.User{{ID: "user1"}, {ID: "user2"}}
+
+		realmClient := newMockClient()
```
makesitgo, line 49: it doesn't appear that `newMockClient` is used anywhere else.  if thats the case, i think we can just simply create the entire mock client directly, rather than leverage a helper function here.

internal/commands/user/list/command_test.go
```diff
@@ -0,0 +1,192 @@
...
+	testApp := realm.App{
+		ClientAppID: "eggcorn-abcde",
+		Name:        "eggcorn",
+	}
+
+	newMockClient := func() mock.RealmClient {
+		realmClient := mock.RealmClient{}
+		realmClient.FindAppsFn = func(filter realm.AppFilter) ([]realm.App, error) {
+			testApp.GroupID = filter.GroupID
+			testApp.ID = filter.App
```
makesitgo, line 40: [question] why are we overwriting the `testApp` details here? if for testing purposes, i'd rather us have a new variable(s) like `capturedAppFilter`/`actualAppFilter` or something similar

internal/commands/user/list/command_test.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+		return realmClient
+	}
+
+	t.Run("Should find app users", func(t *testing.T) {
+		testUsers := []realm.User{{ID: "user1"}, {ID: "user2"}}
+
+		realmClient := newMockClient()
+		realmClient.FindUsersFn = func(groupID, appID string, filter realm.UserFilter) ([]realm.User, error) {
+			assert.Equal(t, "projectID", groupID)
```
makesitgo, line 51: let's avoid `assert`s in the mocked functions themselves.  instead, use a `var` to capture details from here, and then run the asserts down below in the test file where appropriate

internal/commands/user/list/command_test.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func TestUserListFeedback(t *testing.T) {
+	for _, tc := range []struct {
+		description    string
+		users          []realm.User
+		expectedOutput string
+	}{
+		{
+			description: "",
```
makesitgo, line 123: I think you're missing a test case for the empty users case, as well as a proper description here

internal/commands/user/list/flags.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+	"strings"
+)
+
+const (
+	flagState      = "state"
+	flagStateShort = "s"
+	flagStateUsage = `select the state of users to list, available options: ["enabled", "disabled"]`
+
+	flagStatus      = "status-pending"
```
makesitgo, line 13: still need to rename this flag value to `pending`, and can we also update the `var` names to be `flagPending...`

internal/commands/user/list/flags.go
```diff
@@ -0,0 +1,65 @@
...
+	flagStatusShort = "p"
+	flagStatusUsage = `select the state of users to list, available options: ["enabled", "disabled"]`
+
+	flagProviderTypes      = "provider"
```
makesitgo, line 17: similarly here, can we rename the vars to match the flag value, so it's `flagProvider`, `flagProviderShort`, ...

internal/commands/user/list/flags.go
```diff
@@ -0,0 +1,65 @@
...
+	flagProviderTypesShort = "t"
+	flagProviderTypesUsage = `todo add description`
+
+	flagUsers      = "users"
```
makesitgo, line 21: [nit] make this flag singular ("user") and update the var names accordingly

internal/commands/user/list/flags.go
```diff
@@ -0,0 +1,65 @@
...
+	flagUsersShort = "u" //idk what you want this to be
```
makesitgo, line 22: i like `"u"` here like you have it!

internal/commands/user/list/flags.go
```diff
@@ -0,0 +1,65 @@
...
+	flagStateShort = "s"
+	flagStateUsage = `select the state of users to list, available options: ["enabled", "disabled"]`
+
+	flagStatus      = "status-pending"
+	flagStatusShort = "p"
+	flagStatusUsage = `select the state of users to list, available options: ["enabled", "disabled"]`
+
+	flagProviderTypes      = "provider"
+	flagProviderTypesShort = "t"
+	flagProviderTypesUsage = `todo add description`
```
makesitgo, line 19: set the provider types for which to filter the list of app users with

internal/commands/user/list/flags.go
```diff
@@ -0,0 +1,65 @@
...
+
+	flagUsers      = "users"
+	flagUsersShort = "u" //idk what you want this to be
+	flagUsersUsage = `todo add description`
```
makesitgo, line 23: set the user ids for which to filter the list of app users with

internal/commands/user/list/flags.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+const (
+	providerTypeLocalUserPass string = "local-userpass"
```
makesitgo, line 27: the `string` declarations here are unnecessary

internal/commands/user/list/flags.go
```diff
@@ -0,0 +1,65 @@
...
+	providerTypeAPIKey        string = "api-key"
+	providerTypeFacebook      string = "oauth2-facebook"
+	providerTypeGoogle        string = "oauth2-google"
+	providerTypeAnonymous     string = "anon-user"
+	providerTypeCustom        string = "custom-token"
+)
+
+var (
+	errInvalidProviderType = func() error {
+		allProviderTypes := []string{
```
makesitgo, line 37: if you leave these typed as strings, then you won't need to do the `func() error` approach here and can simply define the error inline:

```go
errInvalidProviderType = fmt.Errorf(
  "unsupported value, use one of [%s] instead",
  strings.Join(
    []string{providerTypeAPIKey, providerTypeFacebook, ...},
    ", ",
  ),
)
```

makesitgo (commented):
lookin good so far, mostly just minor nits and cleanup stuff from me, and a missing test.  should be good to go after all that

internal/commands/user/list/command_test.go
ermay12: I was doing that for testing purposes.  Not so much to capture the values there, but to return an app with the correct ID and GroupID.  I added capturing variables when I find the users with the app.  LMK if you don't want this

ermay12:
@makesitgo rfal.  Also, I talked to Nathan and we're both doing the same thing.  I'm gonna put a different idea in a new commit for you to consider though.

internal/commands/user/list/command.go
```diff
@@ -13,8 +20,142 @@ var Command = cli.CommandDefinition{
...
+			userTableRows(providerType, users)...,
+		))
+	}
+	return ui.Print(logs...)
+}
+
+func sortUsersByLastAuthentication(users []realm.User) {
+	sort.Slice(
+		users,
+		func(i, j int) bool {
```
makesitgo, line 111: sorry i wasn't very clear here, I meant pull out the `func(i, j int) bool` ... function into its own, unexported standalone function here.  Then in the `usersByProviderType` for loop you could just have 

```go
for providerType, users := range usersByProviderType {
  sort.Slice(users, sortUsersByLastAuthentication)
  // log appending here...
}

File: internal/commands/user/list/command.go [hunk 19]
```diff
@@ -13,8 +20,142 @@ var Command = cli.CommandDefinition{
...
+			return users[i].LastAuthenticationDate > users[j].LastAuthenticationDate
+		},
+	)
+}
+
+func userTableHeaders(providerType string) []string {
```
makesitgo, line 117: can we add a few quick unit tests for `userTableHeaders` and `userTableRow` (I'm thinking we're ok without unit tests for `userTableRows`, but your call as to how to slice it.)

internal/commands/user/list/command_test.go
makesitgo: we shouldn't need to be mutating a `testApp` struct's data to ensure a test passes (especially a `testApp` struct that's outside of this test closure and is used in downstream/subsequent tests)

in general, we want to use mock functions to either return static data/errors, or simply pass-through/capture the data it receives through args.  in this case, I think just adding `ID` and `GroupID` fields to the `testApp` initialization up front would be a fine alternative.

also, can we add a test (or build on to this one) to make sure we are passing in the appropriate `realm.UserFilter` based on the inputs provided?

internal/commands/user/list/command_test.go
```diff
@@ -0,0 +1,201 @@
...
+					"  myName  id4  true     type1  2005-03-18 01:58:31 +0000 UTC",
+					"",
+				},
+				"\n",
+			),
+		},
+		{
+			description: "Should indicate no users found when none are found",
+			users:       []realm.User{},
+			expectedOutput: strings.Join(
```
makesitgo, line 180: [nit] just keep this as a simple string: `"01:23:45 UTC INFO  No available users to show\n"` (and maybe move this test case to the top of the test table?)

internal/commands/user/list/flags.go
```diff
@@ -0,0 +1,70 @@
...
+					providerTypeAnonymous,
+					providerTypeCustom,
+				},
+				", ",
+			),
+		)
+	}()
+)
+
+func areValidProviderTypes(providers []string) bool {
```
makesitgo, line 54: 😄 
personally I still lean towards using "is" here, but who am I to stand in the way of good grammar...

internal/commands/user/list/inputs.go
```diff
@@ -0,0 +1,27 @@
...
+type inputs struct {
+	cli.ProjectAppInputs
+	UserState     realm.UserState
+	Pending       bool
+	ProviderTypes []string
+	Users         []string
+}
+
+func (i *inputs) Resolve(profile *cli.Profile, ui terminal.UI) error {
+	if !areValidProviderTypes(i.ProviderTypes) {
```
makesitgo, line 18: it's probably worth adding an `inputs_test.go` that makes sure provider types are resolved/validated properly.  that and/or just some unit tests in `flags_test.go` here could work

makesitgo (commented):
last few comments from me...great work here.  i'll hold off on the green light until we talk more about those provider types flags (you mentioned another approach to consider).

but otherwise this all looks good

internal/cloud/realm/user.go
```diff
@@ -26,13 +26,49 @@ const (
 // UserState is a Realm application user state
 type UserState string
 
+// String returns the user state string
+func (us UserState) String() string { return string(us) }
+
+// Type returns the user state type
+func (us UserState) Type() string { return "string" }
```
ermay12, line 33: this is failing the evergreen linter.  It wants me to make "string" a constant.  Should I do that or suppress the warning? Where would I put the constant, or how do i suppress the warning?

ermay12:
@makesitgo RFAL.  Yeah idk about another approach for the provider types flag.  I was thinking i could do something compositional but I forgot that there is no way to get the stringSliceValue.

ermay12:
> @makesitgo RFAL. Yeah idk about another approach for the provider types flag. I was thinking i could do something compositional but I forgot that there is no way to get the stringSliceValue.

wait I see how I can get the stringSliceValue.  Let me try that real quick

ermay12:
@makesitgo RFAL

internal/commands/user/create/flags.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/10gen/realm-cli/internal/utils/flags"
```
makesitgo, line 7: add a newline to separate these import groups

internal/commands/user/list/command_test.go
```diff
@@ -0,0 +1,277 @@
...
+		assert.Nil(t, cmd.realmClient)
+
+		assert.Nil(t, cmd.Setup(profile, nil))
+		assert.NotNil(t, cmd.realmClient)
+	})
+}
+
+func TestUserListHandler(t *testing.T) {
+
+	const projectID = "projectID"
```
makesitgo, line 29: [curious] as to why you're using `const` declarations here?  a simple `projectID, appID := "projectID", "appID"` should suffice here

internal/utils/flags/flags.go
```diff
@@ -0,0 +1,122 @@
...
+	//TypeString is the type of strings
+	TypeString = "string"
+)
+
+// enumSetValue is modified copy of stringSliceValue from
+// the cobra pflags package
+type enumSetValue struct {
+	value          *[]string
+	changed        bool
+	validValuesSet map[string]bool
```
makesitgo, line 23: [nit] for a set, use `struct{}` instead of `bool` for the value

internal/utils/flags/flags.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// NewEnumSetValue creates a stringSliceValue that validates inputs and prevents duplicates
+func NewEnumSetValue(p *[]string, val []string, validValues []string) pflag.Value {
+	esv := new(enumSetValue)
+	esv.value = p
+	*esv.value = val
+	esv.validValuesSet = make(map[string]bool)
```
makesitgo, line 31: can use `len(validValues)` to initialize the size of this map

internal/utils/flags/flags.go
```diff
@@ -0,0 +1,122 @@
...
+		*esv.value = v
+	} else {
+		*esv.value = append(*esv.value, v...)
+	}
+	esv.changed = true
+	return esv.validateAndRemoveDuplicates()
+}
+
+func (esv *enumSetValue) Type() string {
+	return "stringSlice"
```
makesitgo, line 53: let's change this up to be `"enumSet"`

internal/utils/flags/flags.go
```diff
@@ -0,0 +1,122 @@
...
+func (esv *enumSetValue) Replace(val []string) error {
+	*esv.value = val
+	return esv.validateAndRemoveDuplicates()
+}
+
+func (esv *enumSetValue) GetSlice() []string {
+	return *esv.value
+}
+
+func (esv *enumSetValue) validateAndRemoveDuplicates() error {
```
makesitgo, line 75: [question] is this necessary?  couldn't we also keep a `valueSet map[string]struct{}` field on `enumSetValue` populated as we call `Set`/`Append`/`Replace` so we can avoid building this set each and every time `validateAndRemoveDuplicates` is called?

internal/utils/flags/flags.go
```diff
@@ -0,0 +1,122 @@
...
+	return nil
+}
+
+func (esv *enumSetValue) errInvalidEnumValue() error {
+	validEnumValues := make([]string, 0, len(esv.validValuesSet))
+	for validEnumValue := range esv.validValuesSet {
+		validEnumValues = append(validEnumValues, validEnumValue)
+	}
+	// sort to ensure a consistent error message
+	sort.Strings(validEnumValues)
```
makesitgo, line 98: we can avoid this call if we just keep a hold to the `validValues []string` arg that's passed in to the constructor.  that way the caller has control over the preferred ordering, and enumSet needs no knowledge of how to do that

internal/utils/flags/flags_test.go
```diff
@@ -0,0 +1,177 @@
+package flags
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/spf13/pflag"
```
makesitgo, line 9: add a newline to separate these imports

internal/utils/flags/flags_test.go
```diff
@@ -0,0 +1,177 @@
...
+			enumSetValue := NewEnumSetValue(&values, tc.initialValues, validValues)
+			assert.Nil(t, enumSetValue.(pflag.SliceValue).Append(tc.newValue))
+			assert.Equal(t, tc.expectedValues, values)
+		})
+	}
+	t.Run("Invalid values should cause an error", func(t *testing.T) {
+		values := []string{}
+		enumSetValue := NewEnumSetValue(&values, []string{}, validValues)
+		expectedError := fmt.Errorf("unsupported value, use one of [%s] instead", strings.Join(validValues, ", "))
+		assert.Equal(t, expectedError, enumSetValue.(pflag.SliceValue).Append("eggcorn"))
```
makesitgo, line 132: these casts can go away in the `Replace`/`Append`/`GetSlice` tests if you change the return type of `NewEnumSet` to be `*EnumSetValue`

internal/utils/flags/flags.go
```diff
@@ -0,0 +1,122 @@
...
+// enumSetValue is modified copy of stringSliceValue from
+// the cobra pflags package
+type enumSetValue struct {
+	value          *[]string
+	changed        bool
+	validValuesSet map[string]bool
+}
+
+// NewEnumSetValue creates a stringSliceValue that validates inputs and prevents duplicates
+func NewEnumSetValue(p *[]string, val []string, validValues []string) pflag.Value {
```
makesitgo, line 27: can we just return an exported type of `*EnumSetValue` here?  I like the idea here, but with the fields already un-exported we aren't gaining much by returning the interface here (and only half of the implemented interfaces since we've omitted `SliceValue` here)

also, can we just call this `NewEnumSet`?  I think the `flags` package prefix gives us enough context into what we're creating here is actually a flag value

internal/utils/flags/flags_test.go
```diff
@@ -0,0 +1,177 @@
...
+			values := []string{}
+			enumSetValue := NewEnumSetValue(&values, tc.initialValues, validValues)
+			assert.Nil(t, enumSetValue.(pflag.SliceValue).Append(tc.newValue))
+			assert.Equal(t, tc.expectedValues, values)
+		})
+	}
+	t.Run("Invalid values should cause an error", func(t *testing.T) {
+		values := []string{}
+		enumSetValue := NewEnumSetValue(&values, []string{}, validValues)
+		expectedError := fmt.Errorf("unsupported value, use one of [%s] instead", strings.Join(validValues, ", "))
```
makesitgo, line 131: can we extract this `expectedError` to some `var` or other variable defined near the top of all these test scopes?  also, for simplicity's sake, just construct the string directly:

```go
expectedErr := errors.New(`unsupported value, use one of [1-value, 2-value, 3-value, with comma] instead`)
```

(just looking at that above now, maybe it makes sense to wrap each "valid value" inside quotes, like the `writeCSV` method would do for any value that had a comma in it.  what do you think?)

makesitgo (approved):
just a last bit of cleanup here, and then once the commits and evergreen are in line, this is good to go!  thanks again for all this good work!

🆒  beans

internal/utils/flags/flags.go
ermay12: Sure, I didn't think performance was at all a consideration for flag parsing, so I just did it this way cuz it reads simpler.  I'll add it to the struct then

internal/utils/flags/flags.go
```diff
@@ -6,97 +6,115 @@ import (
...
+	}
+	esv.valuesSet = make(map[string]struct{}, len(defaultValues))
+	for _, value := range defaultValues {
+		esv.valuesSet[value] = setMember
+	}
+	err := esv.validateAndRemoveDuplicates()
+	if err != nil {
+		// what should I do here?  I'm pretty sure this would be a great example of when
+		// to panic (coding error in a single threaded, non-critical application) but I know
+		// you don't like that so lmk what to do.
```
ermay12, line 46: this

ermay12, line 46: this

ermay12 (commented):
@makesitgo RFAL

internal/utils/flags/flags.go
makesitgo: for this, i would either remove `defaultValues` all together (we don't/won't use them for array flags in this cli) and its corresponding validation in this constructor, or simply only include the defaults that are valid values, and silently ignore invalid ones here

Looking at this more, I also think this `EnumSetValue` would benefit from having a `func (es *EnumSetValue) set(vals ...string) error` function defined on it, so we can consolidate this validation/append logic and use them where need be in `Set`/`Append`/`Replace`.  you could use it here to, but just `set` with each individual default value so you could `continue` on any error

ermay12:
@makesitgo RFAL

PR #47, opened 2021-01-07
REALMC-7580 Add follow up message functionality for errors 

jo-tse-mongo:
[WIP] PR until REALMC-7573 is merged finalized and merged in.


makesitgo:
@jo-tse-mongo ... you should be able to run a fetch/pull/rebase off of `next` branch now to pull in Kush' recent changes so you can work on the rest of this ticket now.  Not going to submit a formal review but at a quick glance things look to be in good shape!  (Check the linter errors here for a few quick cleanups..)

internal/cli/command_factory.go
```diff
@@ -160,7 +162,20 @@ func (factory *CommandFactory) Run(cmd *cobra.Command) {
 			factory.errLogger.Fatal(err)
 		}
 
-		if printErr := factory.ui.Print(terminal.NewErrorLog(err)); printErr != nil {
+		logs := []terminal.Log{
+			terminal.NewErrorLog(err),
```
makesitgo, line 166: [nit] can keep all of this declaration as a one-liner

internal/cli/command_factory.go
```diff
@@ -160,7 +162,20 @@ func (factory *CommandFactory) Run(cmd *cobra.Command) {
...
+		}
+
+		if e, ok := err.(realm.ErrInvalidSession); ok {
```
makesitgo, line 169: so we actually don't care at this point if it's an `ErrInvalidSession` or not, we'll allow any type of error here define suggested commands or reference links.  what you're looking for here is something along the lines of:

```go
if e, ok := err.(CommandSuggester); ok {
  logs = append(logs, terminal.NewSuggestedCommandsLog(e.SuggestedCommands()))
}
if e, ok := err.(LinkReferrer); ok {
  logs = append(logs, terminal.NewReferenceLinksLog(e. ReferenceLinks()))
}
```

internal/cli/error.go
```diff
@@ -0,0 +1,22 @@
...
+	DisableUsage() struct{}
+}
+
+type errDisableUsage struct {
+	error
+}
+
+func (err errDisableUsage) DisableUsage() struct{} { return struct{}{} }
+
+// CommandSuggester handles any suggestions to run if the current command isn't working
```
makesitgo: [nit] `// CommandSuggester provides a list of suggested commands that will display to the user when an error occurs`

internal/cli/error.go
```diff
@@ -0,0 +1,22 @@
...
+type CommandSuggester interface {
+	SuggestedCommands() []string
+}
+
+// LinkReferrer gives a list of links that relate to this command to give the user more context
```
makesitgo: [nit] `// LinkReferrer provides a list of reference links that will display to the user when an error occurs`

internal/cloud/realm/error.go
```diff
@@ -3,13 +3,31 @@ package realm
...
 	"net/http"
 )
 
 const (
 	invalidSessionCode = "InvalidSession"
 )
 
+// ErrInvalidSession is an invalid session error that has a follow up message
+type ErrInvalidSession struct {
+	error
```
makesitgo, line 16: we can get rid of this `error` embed, and the constructor for this type.  instead, declare/return it with just `realm.ErrInvalidSession{}`

what you can do here for now is add a field `profileName string` to this struct definition.  for now, don't worry about setting that field's value whenever you declare this error (that'll be for another ticket).  but we can use/reference that field in the `SuggestedCommands` implementation below

you will also need to implement `func (err ErrInvalidSession) Error() string { return "invalid session" }` here

internal/cloud/realm/error.go
```diff
@@ -3,13 +3,31 @@ package realm
...
+}
+
+func (err ErrInvalidSession) SuggestedCommands() []string {
+	return []string{}
+}
+
+func (err ErrInvalidSession) ReferenceLinks() []string {
```
makesitgo, line 25: you can remove this function declaration, `ErrInvalidSession` will not need to provide any ReferenceLinks at this time

internal/cloud/realm/error.go
```diff
@@ -3,13 +3,31 @@ package realm
...
 	invalidSessionCode = "InvalidSession"
 )
 
+// ErrInvalidSession is an invalid session error that has a follow up message
+type ErrInvalidSession struct {
+	error
+}
+
+func (err ErrInvalidSession) SuggestedCommands() []string {
+	return []string{}
```
makesitgo, line 21: here is where you'll now want to return:

```go
func (err ErrInvalidSession) SuggestedCommands() []string {
  suggestion := "realm-cli login"
  if err.profileName != "" {
    suggestion += " --profile " + err.profileName
  }
  return []string{suggestion}
}
```

internal/cloud/realm/error.go
```diff
@@ -3,13 +3,31 @@ package realm
...
+}
+
+func (err ErrInvalidSession) ReferenceLinks() []string {
+	return []string{}
+}
+
+func NewErrInvalidSession() ErrInvalidSession {
```
makesitgo, line 30: can remove this

internal/terminal/log.go
```diff
@@ -76,6 +76,16 @@ func NewWarningLog(format string, args ...interface{}) Log {
 	return newLog(LogLevelWarn, newTextMessage(format, args...))
 }
 
+// NewSuggestedCommandsLog creates a log with a follow up message about suggested commands for the error
+func NewSuggestedCommandsLog(commands []string) Log {
+	return newLog(LogLevelInfo, newFollowUpMessage(commandMessage, commands))
```
makesitgo, line 81: let's make this message: `"Try running instead"`

internal/terminal/log.go
```diff
@@ -76,6 +76,16 @@ func NewWarningLog(format string, args ...interface{}) Log {
...
+}
+
+// NewReferenceLinksLog creates a log with a follow up message about reference links for the error
+func NewReferenceLinksLog(links []string) Log {
+	return newLog(LogLevelInfo, newFollowUpMessage(linkMessage, links))
```
makesitgo, line 86: let's make this message "For more information"

makesitgo (commented):
ok, here's a first batch of comments for you, and i have one question:

in your opinion: is there much value added from the `followupMessage` struct, when its compared to how the `list` struct works?  I might have sent you down a bad path there with this ticket, but if you don't think there's much difference then I'd say we can remove the `followup` stuff and just use `newList` when creating the new log types.

what do you think?

internal/terminal/list.go
```diff
@@ -26,6 +26,14 @@ func newList(message string, data []interface{}) list {
 }
 
 func (l list) Message() (string, error) {
+	if len(l.data) == 0 {
+		return l.message, nil
+	}
+
+	if len(l.data) == 1 {
+		return fmt.Sprintf("%s: %s", l.message, l.data[0]), nil
```
makesitgo, line 34: let's add some logic within this case here to account for a potentially empty `l.message` (the other cases don't need that logic just due to the way its displayed)

internal/terminal/log.go
```diff
@@ -12,11 +12,16 @@ import (
 // LogLevel is the level of a terminal log
 type LogLevel string
 
-// set of supported log levels
```
makesitgo, line 15: undo this change (it's failing the linter), you can put the unexported `linkMessage`/`commandMessage` in a separate `const` block, and maybe place that block closer to the constructor functions where these are used

internal/terminal/log.go
```diff
@@ -76,6 +81,16 @@ func NewWarningLog(format string, args ...interface{}) Log {
 	return newLog(LogLevelWarn, newTextMessage(format, args...))
 }
 
+// NewSuggestedCommandsLog creates a log with a follow up message about suggested commands for the error
+func NewSuggestedCommandsLog(commands []interface{}) Log {
```
makesitgo, line 85: [nit] make this arg variadic `commands ...interface{}`

internal/terminal/log.go
```diff
@@ -76,6 +81,16 @@ func NewWarningLog(format string, args ...interface{}) Log {
...
+	return newLog(LogLevelDebug, newList(commandMessage, commands))
+}
+
+// NewReferenceLinksLog creates a log with a follow up message about reference links for the error
+func NewReferenceLinksLog(links []interface{}) Log {
```
makesitgo, line 90: [nit] make this arg variadic `links ...interface{}`

makesitgo (approved):
this all looks good to me with just a few more cleanup and getting the commits/evergreen in line

🆒  beans

makesitgo (approved):
looks right to me... just make sure to squash those commits and leave evergreen happy

mburdette (approved):

PR #48, opened 2021-01-12
REALMC-7159: Buildout command: user delete

nathantfrank:
Will be adding more tests. Wanted to get this PR going to continue on the conversation.


nathantfrank:
Additionally want to extend the output, to a simple table of failed and deleted users potentially.

internal/commands/user/delete/inputs.go
```diff
@@ -0,0 +1,193 @@
...
+				i.State = realm.UserState(selectedStates[0])
+			} else {
+				i.State = realm.UserStateNil
+			}
+		}
+	}
+	return nil
+}
+
+func (i *inputs) ResolveUsers(ui terminal.UI, client realm.Client, app realm.App) error {
```
kpatel71716, line 118: Since user revoke needs this as well, it might be a good idea to move this function into a common place and not have it be a receiver to inputs. I assume that the return could then be changed to ([]string, error) where we return a list of userID strings

internal/commands/user/delete/inputs.go
```diff
@@ -0,0 +1,193 @@
...
+			}
+			if len(selectedStates) == 1 {
+				i.State = realm.UserState(selectedStates[0])
+			} else {
+				i.State = realm.UserStateNil
+			}
+		}
+	}
+	return nil
+}
```
kpatel71716, line 116: This seems like a good idea to have for user revoke as well. If we move this into a common function, then I would be happy to reuse it 😆

internal/commands/user/delete/flags.go
```diff
@@ -0,0 +1,64 @@
...
+	"strings"
+)
+
+const (
+	flagUsers      = "user"
+	flagUsersShort = "u"
+	flagUsersUsage = "users to be deleted"
+
+	flagInteractive      = "interactive-filter"
+	flagInteractiveShort = "i"
```
kpatel71716, line 14: I ran into conflicts when testing out this branch with a profile since the profile flag shorthand is also 'I'. Unfortunately 'f' seems to be taken as well, but maybe there is another letter that makes sense if we need to rename this.

internal/commands/user/delete/command.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	"github.com/10gen/realm-cli/internal/commands/user/shared"
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/10gen/realm-cli/internal/utils/flags"
+	"github.com/spf13/pflag"
+)
+
+// Command is the `user delete` command
+var Command = cli.CommandDefinition{
```
kpatel71716, line 15: You may have already noticed this, but you will need to add the delete command to the SubCommands section of the user command definition

kpatel71716 (commented):
Nice stuff! Have some comments from testing this out/integrating with my user revoke code

internal/commands/user/shared/flags.go
```diff
@@ -53,3 +66,51 @@ func IsValidProviderType(pt string) bool {
...
+		return ErrInvalidStatusType
+	}
+
+	*st = newStatusType
+	return nil
+}
+
+// set of StatusType(s)
+const (
+	StatusTypeNil       StatusType = ""
```
tsedgwick, line 93: [nit] can we make this StatusTypeEmpty? Empty is more idiomatic go since nil denotes that it is truly a nil value, not an empty one. 

I know you moved this code but first time I'm seeing it. @makesitgo is this even necessary? I'm leaning towards a `status == ""` is a lot cleaner and more straight forward than explicitly having this empty status type.

internal/commands/user/shared/input_test.go
```diff
@@ -0,0 +1,79 @@
...
+	} {
+		t.Run(fmt.Sprintf("%s Setup should prompt for Users", tc.description), func(t *testing.T) {
+
+			// out, outErr := mock.FileWriter(t)
+			// assert.Nil(t, outErr)
+			// defer out.Close()
+
+			// c, err := expect.NewConsole(expect.WithStdout(out))
+			// assert.Nil(t, err)
+			// defer c.Close()
```
tsedgwick, line 48: is this still needed?

internal/commands/user/shared/flags.go
```diff
@@ -0,0 +1,116 @@
...
+// StatusType error msg
+var (
+	ErrInvalidStatusType = func() error {
+		allStatusTypes := []string{StatusTypeConfirmed.String(), StatusTypePending.String()}
+		return fmt.Errorf("unsupported value, use one of [%s] instead", strings.Join(allStatusTypes, ", "))
+	}()
+)
+
+// IsValidStatusType checks if a valid StatusType
+func IsValidStatusType(st StatusType) bool {
```
tsedgwick, line 107: if we just allow a string here instead of a typed string, we can easily check for ""


@makesitgo is the typed string really necessary here? I'm not seeing much benefit over just having two typed string constants. These seem like they can either just be string constants (preferred), put into a slice (less preferred), or put into a map (least preferred).

internal/commands/user/shared/flags.go
```diff
@@ -0,0 +1,116 @@
...
+const (
+	StatusTypeNil       StatusType = ""
+	StatusTypeConfirmed StatusType = "confirmed"
+	StatusTypePending   StatusType = "pending"
+)
+
+// StatusType error msg
+var (
+	ErrInvalidStatusType = func() error {
+		allStatusTypes := []string{StatusTypeConfirmed.String(), StatusTypePending.String()}
```
tsedgwick, line 101: we can create this as a global var instead of recreating each time this func is called.

internal/commands/user/shared/input_test.go
```diff
@@ -0,0 +1,79 @@
...
+				ID:      primitive.NewObjectID().Hex(),
+				GroupID: primitive.NewObjectID().Hex(),
+			}
+
+			_, err := tc.inputs.ResolveUsers(ui, realmClient, app)
+
+			console.Tty().Close() // flush the writers
+			<-doneCh              // wait for procedure to complete
+
+			// assert.Equal(t, tc.expectedUsers, users) // Still failing as of now
```
tsedgwick, line 75: is this still needed?

internal/commands/user/shared/inputs.go
```diff
@@ -0,0 +1,96 @@
...
+	Status            StatusType
+	InteractiveFilter bool
+}
+
+// ResolveUsers will use the provided Realm client to resolve the users specified by the realm.App through inputs
+func (i *UsersInputs) ResolveUsers(ui terminal.UI, client realm.Client, app realm.App) ([]string, error) {
+	var err error
+
+	userFilter := realm.UserFilter{}
+	if i.Status != StatusTypeNil {
```
tsedgwick, line 25: this check doesn't seem necessary since if it is true `userFilter.Pending = i.Status == StatusTypePending` will be false as well.

internal/commands/user/shared/inputs.go
```diff
@@ -0,0 +1,96 @@
...
+		case ProviderTypeGoogle:
+			selectableUserOptions[x] = fmt.Sprintf(userOptionPattern, "Google", user.ID)
+		case ProviderTypeFacebook:
+			selectableUserOptions[x] = fmt.Sprintf(userOptionPattern, "Facebook", user.ID)
+		case ProviderTypeCustom:
+			selectableUserOptions[x] = fmt.Sprintf(userOptionPattern, "Custom JWT", user.ID)
+		case ProviderTypeCustomFunction:
+			selectableUserOptions[x] = fmt.Sprintf(userOptionPattern, "Custom Function", user.ID)
+		}
+		x++
```
tsedgwick, line 77: [nit] let's drop this counter (should be named i fwiw), and use append to make this a little cleaner.

selectableUserOptions = append(selectableUserOptions, fmt.Sprintf(userOptionPattern, "Anonymous", user.ID))

internal/commands/user/shared/inputs.go
```diff
@@ -0,0 +1,96 @@
...
+			Options: selectableUserOptions,
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+	users := make([]string, len(selectedUsers))
+	for x, userPattern := range selectedUsers {
+		users[x] = strings.Split(userPattern, " - ")[1]
+	}
```
tsedgwick, line 93: [nit] same here as above with append, or change the var to `i`

internal/commands/user/shared/inputs.go
```diff
@@ -0,0 +1,96 @@
...
+			Message: "Which user(s) would you like to delete?",
+			Options: selectableUserOptions,
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+	users := make([]string, len(selectedUsers))
+	for x, userPattern := range selectedUsers {
+		users[x] = strings.Split(userPattern, " - ")[1]
```
tsedgwick, line 92: is there a better way to do this? seems a little odd to me we have control over this userPattern / string yet we have to do this string manipulation.

internal/commands/user/shared/inputs.go
```diff
@@ -0,0 +1,96 @@
...
+	if i.Status != StatusTypeNil {
+		userFilter.Pending = i.Status == StatusTypePending
+	}
+	if len(i.ProviderTypes) > 0 {
+		userFilter.Providers = i.ProviderTypes
+	}
+	if i.State != realm.UserStateNil {
+		userFilter.State = i.State
+	}
+	selectableUsersSet := make(map[string]realm.User)
```
tsedgwick, line 34: [mega-nit] we don't gain much w.r.t. the make here and maps. make is best used for slices, maps can just be created inline since we don't create with a len/capacity like we do slices `map[string]realm.User{}`

internal/commands/user/shared/inputs.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	if i.Status == StatusTypePending || i.Status == StatusTypeNil {
+		foundUsers, err := client.FindUsers(app.GroupID, app.ID, userFilter)
+		if err != nil {
+			return nil, err
+		}
+		for _, user := range foundUsers {
+			selectableUsersSet[user.ID] = user
+		}
+	}
```
tsedgwick, line 52: these blocks appear to be the same, is something missing from one of these?

internal/commands/user/delete/inputs.go
```diff
@@ -0,0 +1,110 @@
...
+			},
+		)
+		if err != nil {
+			return err
+		}
+		if len(selectedStatuses) == 1 {
+			i.Status = shared.StatusType(selectedStatuses[0])
+		} else {
+			i.Status = shared.StatusTypeNil
+		}
```
tsedgwick, line 60: if we don't type these strings, this code goes away (see comment below on statustypes)

internal/commands/user/delete/inputs.go
```diff
@@ -0,0 +1,110 @@
...
+					Default: i.ProviderTypes,
+				},
+			)
+			if err != nil {
+				return err
+			}
+		}
+
+		// Interactive set State
+		if i.InteractiveFilter {
```
tsedgwick, line 79: we're doing the same check twice here. Was this intentional? if not, can we just flip the first conditional, or add this conditional to the status check and return early?

```
if i.Status != shared.StatusTypeConfirmed && i.Status != shared.StatusTypeNil && i.InteractiveFilter {
	return nil
}


err := ui.AskOne(
. // rest of interactive code here
.
.
```

tsedgwick (changes requested):
pretty good, some minor callouts from my end. One major callout pre-existed this PR I believe

makesitgo:
https://github.com/10gen/realm-cli/pull/48#discussion_r556557972

@tsedgwick ... what the typed string allows us to do is bake validation into "allowed values" so that on `flag.Parse`, we're able to fail the command if/when an invalid value is present.  what's nice with that approach is we know that error will occur before any CLI code (input/setup) is executed, and when the error arises it will be printed alongside the CLI help/usage text.

and I'm still not 100% if this type will ever need to be prompted/interactive, but if/when that happens the typed string allows us to implement interfaces that allow it to be used within the `survey` library (again, with validation baked in)

makesitgo:
https://github.com/10gen/realm-cli/pull/48#discussion_r556558353

toyed with that idea, but it felt weird/unnecessary to create a global array of string displays for these types (they would just be used for the error message display).  i eschewed performance and had the string array recreated each time that function's called since it's only called at-most once in the flow of code here

nathantfrank:
@tsedgwick @makesitgo  @kpatel71716 RFAL

internal/commands/user/shared/flags.go
nathantfrank: Totally fine with using the empty terminology here. @makesitgo

internal/commands/user/shared/inputs.go
```diff
@@ -0,0 +1,102 @@
...
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+	users := []string{}
+	for _, userOption := range selectedUsers {
+		users = append(users, selectableUsers[userOption])
+	}
+	return users, nil
```
kpatel71716, line 101: [q] Should we return an error here if we found no users (or even handle that upstream in the command after this function returns)?

internal/commands/user/delete/command.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+	} else {
+		users = cmd.inputs.Users
+	}
+
+	for _, userID := range users {
+		err := cmd.realmClient.DeleteUser(app.GroupID, app.ID, userID)
+		if err != nil {
+			cmd.outputs.failed = append(cmd.outputs.failed, fmt.Errorf("failed to delete user (%s): %s", userID, err))
+		}
```
kpatel71716, line 78: Unless I am mistaken, if a random user id were to be passed in with the `--user` flag, we would show a misleading `unable to delete user (<username>)` message instead of an error indicating that no matching user was found. Maybe we could check the error message for this scenario. Thoughts @makesitgo ?

kpatel71716 (commented):
Looks good to me. Just had a couple of minor things

internal/commands/user/delete/command.go
nathantfrank: Yeah I was wondering whether we should do some kind of validation on the user provided ids. But I figured delete would return a failure either way. If we are to do this it might make more sense for ResolveUsers to return actual Users vs Ids. That way we can do more custom messaging. I'll work on an iteration where that is the case and see how we like it.

internal/commands/user/shared/inputs.go
nathantfrank: I personally think it should be handled slightly upstream. With potentially a feedback message saying "no users to delete".

nathantfrank:
@tsedgwick @makesitgo @kpatel71716 RFAL, included more detailed output, slight change to resolveUsers

internal/commands/user/delete/command.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (cmd *command) Handler(profile *cli.Profile, ui terminal.UI) error {
+	app, err := cli.ResolveApp(ui, cmd.realmClient, cmd.inputs.Filter())
+	if err != nil {
+		return err
+	}
+
+	var users []realm.User
+	if len(cmd.inputs.Users) < 1 {
```
makesitgo, line 67: [nit] let's have the check be `== 0`

[thought] could we push this logic into the `inputs` code, something like `users, usersErr := cmd.inputs.ResolveUsers(...)`?

internal/commands/user/delete/flags.go
```diff
@@ -0,0 +1,7 @@
+package delete
+
+const (
```
makesitgo, line 3: with the soon-to-be-rebase, i've been merging this file with any `inputs.go` file we have

internal/commands/user/delete/inputs.go
```diff
@@ -0,0 +1,79 @@
...
+	shared.UsersInputs
+	Users []string
+}
+
+func (i *inputs) Resolve(profile *cli.Profile, ui terminal.UI) error {
+	if err := i.ProjectAppInputs.Resolve(ui, profile.WorkingDirectory); err != nil {
+		return err
+	}
+	// Interactive set Status
+	if i.Status == shared.StatusTypeInteractive {
```
makesitgo, line 22: do you think its possible for us to ignore any "interactive mode" for now, and implement that in a separate ticket?  im ok with this command being temporarily verbose while we figure out how we want to go about allowing these interactive-by-default flags to work

internal/utils/test/mock/terminal_ui.go
```diff
@@ -38,81 +39,85 @@ var (
 	StaticTime = time.Date(1989, 6, 22, 1, 23, 45, 0, time.UTC)
 )
 
-type ui struct {
+// UI is a mocked terminal.UI
+type UI struct {
 	terminal.UI
+
+	AskOneFn func(answer interface{}, prompt survey.Prompt) error
```
makesitgo, line 46: I know we discussed this offline before, but remind me again where/why this was needed?

makesitgo (commented):
just a couple of early thoughts and questions, i like where this is going!  i figure i'll save the detailed review for after I merge #49 and you rebase

internal/commands/user/delete/command.go
```diff
@@ -0,0 +1,164 @@
...
+	fs.StringSliceVarP(&cmd.inputs.Users, flagUsers, flagUsersShort, []string{}, flagUsersUsage)
+	fs.VarP(
+		flags.NewEnumSet(&cmd.inputs.ProviderTypes, shared.ValidProviderTypes),
+		shared.FlagProvider,
+		shared.FlagProviderShort,
+		shared.FlagProviderUsage,
+	)
+	fs.Lookup(shared.FlagProvider).NoOptDefVal = shared.ProviderTypeInteractive
+	fs.VarP(&cmd.inputs.State, shared.FlagStateType, shared.FlagStateTypeShort, shared.FlagStateTypeUsage)
+	fs.Lookup(shared.FlagProvider).NoOptDefVal = shared.ProviderTypeInteractive
```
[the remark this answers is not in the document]
nathantfrank: Ahh good catch. Should be trying to look at the FlagStateType. The lookup is to have reference to the flag that was created so that we can set the default value for when a value is not provided along with the flag. i.e. `--provider`

We talked about testing out the interactive without having an explicit interactive flag

internal/utils/test/mock/terminal_ui.go
nathantfrank: This was so I could mock the AskOne() and have it spit out an error. I didn't see any examples in the code base so far.

internal/commands/user/delete/command.go
nathantfrank: I'll move it

internal/commands/user/delete/inputs.go
nathantfrank: Removed interactive flag modes in next commit

internal/commands/user/flags.go
```diff
@@ -35,5 +51,101 @@ var (
...
+func isValidUserStateType(us userStateType) bool {
+	switch us {
+	case
+		userStateTypeEmpty, // allow state to be optional
+		userStateTypeEnabled,
+		userStateTypeDisabled:
+		return true
+	}
+	return false
+}
```
nathantfrank, line 151: Initially separate from realm.UserState for interactivity default for flag, since not an option for api values. @makesitgo or anyone else say the word and I'll revert back to using realm.UserState

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,115 @@
...
+	userDataName  = "name"
+)
+
+// usersInputs are the filtering inputs for a user command
+type usersInputs struct {
+	State         userStateType
+	ProviderTypes []string
+	Status        statusType
+	Users         []string
+}
```
nathantfrank, line 24: Put this into its own input file since it can be reused by revoke.

internal/utils/test/mock/terminal_ui.go
nathantfrank: Example test case is on line 154 of inputs_test.go

nathantfrank:
@tsedgwick @makesitgo @kpatel71716 RFAL

internal/commands/user/delete.go
```diff
@@ -0,0 +1,169 @@
...
+	return nil
+}
+
+// Feedback is the command feedback
+func (cmd *CommandDelete) Feedback(profile *cli.Profile, ui terminal.UI) error {
+	if len(cmd.outputs) == 0 {
+		return ui.Print(terminal.NewTextLog("No users to delete"))
+	}
+
+	var outputByProviderType = make(map[string][]deleteOutput)
```
tsedgwick, line 99: [nit] make map can just be map[string][]deleteOutput{}

internal/commands/user/delete.go
```diff
@@ -0,0 +1,169 @@
...
+	for _, output := range cmd.outputs {
+		for _, identity := range output.user.Identities {
+			outputByProviderType[identity.ProviderType] = append(outputByProviderType[identity.ProviderType], output)
+		}
+	}
+	var logs []terminal.Log
```
tsedgwick, line 105: we can preallocate this slice since we know the length of it. 
`make([]terminal.Log, len(outputByProviderType))`

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,115 @@
...
+		// case providerTypeApple:
+		// 	opt = strings.Join([]string{"Apple", user.ID}, sep)
+		case providerTypeGoogle:
+			opt = strings.Join([]string{"Google", user.ID}, sep)
+		case providerTypeFacebook:
+			opt = strings.Join([]string{"Facebook", user.ID}, sep)
+		case providerTypeCustom:
+			opt = strings.Join([]string{"Custom JWT", user.ID}, sep)
+			// case providerTypeCustomFunction:
+			// 	opt = strings.Join([]string{"Custom Function", user.ID}, sep)
```
tsedgwick, line 93: should we have a TODO with a jira ticket attached?

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,115 @@
...
+		for _, user := range foundUsers {
+			selectableUsersSet[user.ID] = user
+		}
+	}
+
+	// Interactive User Selection
+	selectableUsers := map[string]realm.User{}
+	selectableUserOptions := make([]string, len(selectableUsersSet))
+	sep := " - "
+	userOptIndex := 0
```
tsedgwick, line 74: can we drop this declared index? we can either use the one in that's currently omitted in the for loop (`for _, user `) or use append on the slice.

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,115 @@
...
+	for _, user := range selectableUsersSet {
+		var opt string
+		switch user.Identities[0].ProviderType {
```
tsedgwick, line 77: is there any chance `len(Identities) == 0`? I would suggest a check here to be sure.

tsedgwick, line 77: is there any chance `len(Identities) == 0`? I would suggest a check here to be sure.

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,115 @@
...
+		case providerTypeAnonymous:
+			opt = strings.Join([]string{"Anonymous", user.ID}, sep)
```
tsedgwick, line 79: we can avoid a slice allocation here if we use `fmt.Sprintf` instead. This goes for this whole select statement

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,115 @@
...
+		case providerTypeLocalUserPass:
+			opt = strings.Join([]string{"User/Password", fmt.Sprint(user.Data[userDataEmail]), user.ID}, sep)
+		case providerTypeAPIKey:
+			opt = strings.Join([]string{"ApiKey", fmt.Sprint(user.Data[userDataName]), user.ID}, sep)
+		// case providerTypeApple:
```
tsedgwick, line 84: should we have a TODO with a jira ticket attached?

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,115 @@
...
+		case providerTypeGoogle:
+			opt = strings.Join([]string{"Google", user.ID}, sep)
+		case providerTypeFacebook:
+			opt = strings.Join([]string{"Facebook", user.ID}, sep)
+		case providerTypeCustom:
+			opt = strings.Join([]string{"Custom JWT", user.ID}, sep)
+			// case providerTypeCustomFunction:
+			// 	opt = strings.Join([]string{"Custom Function", user.ID}, sep)
+		}
+		selectableUserOptions[userOptIndex] = opt
```
tsedgwick, line 95: since there's no default cause, this will always happen. Should we check if `opt = ""` or have a default:
```
default:
continue
```

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,115 @@
...
+		&selectedUsers,
+		&survey.MultiSelect{
+			Message: "Which user(s) would you like to delete?",
+			Options: selectableUserOptions,
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+	users := []realm.User{}
```
tsedgwick, line 110: can we allocate this slice using make since we know the size?

internal/commands/user/delete.go
```diff
@@ -0,0 +1,169 @@
...
+		))
+	}
+	return ui.Print(logs...)
+}
+
+func getOutputComparerBySuccess(outputs []deleteOutput) func(i, j int) bool {
+	return func(i, j int) bool {
+		return !(outputs[i].err != nil && outputs[i].err == nil)
+	}
+}
```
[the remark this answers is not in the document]
nathantfrank: I'll be using it in another user command so I moved to out to be more shareable

internal/commands/user/delete.go
```diff
@@ -0,0 +1,169 @@
...
+
+func userDeleteTableHeaders(providerType string) []string {
+	var headers []string
+	switch providerType {
+	case providerTypeAPIKey:
+		headers = append(headers, headerName)
+	case providerTypeLocalUserPass:
+		headers = append(headers, headerEmail)
+	}
```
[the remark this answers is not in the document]
nathantfrank: If it's neither we are for now ignoring the column for the other types of Provider Types. Which shouldn't be disrupting. 

Is there a way to guarantee for the oauth provider types? jwt claims. might be able to provide a human readable identifier like the email on the jwt.

internal/commands/user/delete.go
```diff
@@ -0,0 +1,169 @@
...
+	msg := "n/a"
+	if output.err != nil {
+		msg = output.err.Error()
+	}
+	row := map[string]interface{}{
+		headerID:      output.user.ID,
+		headerType:    output.user.Type,
+		headerDeleted: output.err == nil,
+		headerDetails: msg,
+	}
```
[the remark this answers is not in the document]
nathantfrank: Yes, we aren't using the map for order

internal/commands/user/inputs.go
nathantfrank: I was wondering whether we should include them or not. I would assume so, but the user list command did not have these 2 provider types at time of the PR merge.

internal/commands/user/inputs.go
nathantfrank: Since we are iterating over key & val, we don't have an index there. And since we know the length, I thought it made the most since to use an index.

internal/commands/user/delete_test.go
```diff
@@ -0,0 +1,307 @@
...
+		},
+	}
+	for _, tc := range []struct {
+		description    string
+		outputs        []userOutput
+		expectedOutput string
+	}{
+		{
+			description:    "Should show indicate no users to delete",
+			outputs:        []userOutput{},
```
makesitgo, line 199: [nit] should be able to just remove this line?

internal/commands/user/flags.go
makesitgo: I think we can probably merge all of this in with `realm.UserState`, check out `realm.Location`/`realm.DeploymentModel` for similar cases (I think).  We do allow `LocationNil` and such there as well

internal/commands/user/input_test.go
```diff
@@ -0,0 +1,202 @@
...
+		{
+			ID: "user-3",
+			Identities: []realm.UserIdentity{
+				{ProviderType: providerTypeLocalUserPass},
+			},
+			Disabled: true,
+		},
+	}
+
+	t.Run("Setup should prompt for Users", func(t *testing.T) {
```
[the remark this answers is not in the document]
makesitgo: I don't know what my OCD did to you in the past Mr. Sedgwick, but I do apologize for it... 🙃 

Is that something we should slowly start adopting across the board here?

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,115 @@
...
+type usersInputs struct {
+	State         userStateType
+	ProviderTypes []string
+	Status        statusType
+	Users         []string
+}
+
+// ResolveUsers will use the provided Realm client to resolve the users specified by the realm.App through inputs
+func (i *usersInputs) ResolveUsers(ui terminal.UI, client realm.Client, app realm.App) ([]realm.User, error) {
+	var err error
```
makesitgo, line 28: can we remove this, and name/declare errors as they arise?

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if len(i.Users) > 0 {
+		var users []realm.User
+		for _, userID := range i.Users {
+			foundUser, err := client.FindUsers(app.GroupID, app.ID, realm.UserFilter{IDs: []string{userID}})
```
makesitgo, line 33: [question] why aren't we just simply creating a `realm.UserFilter` from the `usersInputs` here and sending that in to `realmClient.FindUsers`?  a lot of this code seems to be duplicating logic we already have defined in the realm client?

internal/commands/user/output.go
```diff
@@ -1,5 +1,7 @@
 package user
 
+import "github.com/10gen/realm-cli/internal/cloud/realm"
```
makesitgo, line 3: [nit] still get this import inside a block:

```go
import(
)

File: internal/utils/test/mock/terminal_ui.go [hunk 44]
```diff
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/AlecAivazis/survey/v2"
```
makesitgo, line 13: [nit] newline above this

internal/commands/user/delete.go
```diff
@@ -0,0 +1,169 @@
...
+
+func userDeleteTableRows(providerType string, outputs []deleteOutput) []map[string]interface{} {
+	userDeleteTableRows := make([]map[string]interface{}, 0, len(outputs))
+	for _, output := range outputs {
+		userDeleteTableRows = append(userDeleteTableRows, userDeleteTableRow(providerType, output))
+	}
+	return userDeleteTableRows
+}
+
+func userDeleteTableRow(providerType string, output deleteOutput) map[string]interface{} {
```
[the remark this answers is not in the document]
nathantfrank: terminal.NewTableLog takes in []map[string]interface{} as an arg

internal/utils/test/mock/terminal_ui.go
makesitgo: I'd rather we remove this mock ui for now.  I actually haven't written any tests that simulate errors coming from `survey` (as you can tell).  we might be able to do the same with some simulated key press (ctrl+c or something) and avoid the mock entirely.  thoughts?

makesitgo (commented):
another round of comments to go along with Tim's here

internal/utils/test/mock/terminal_ui.go
nathantfrank: Yeah if we are able to simulate or pass signals, I think pattern is a better fit for these tests.

internal/commands/user/delete_test.go
```diff
@@ -0,0 +1,307 @@
...
+
+		realmClient := mock.RealmClient{}
+		realmClient.FindAppsFn = func(filter realm.AppFilter) ([]realm.App, error) {
+			capturedAppFilter = filter
+			return []realm.App{testApp}, nil
+		}
+
+		realmClient.FindUsersFn = func(groupID, appID string, filter realm.UserFilter) ([]realm.User, error) {
+			capturedProjectID = groupID
+			capturedAppID = appID
```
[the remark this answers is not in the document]
nathantfrank: In this test case, `FindUser` will only be called once, as well as `DeleteUser` will only be called once. Otherwise I completely agree with you.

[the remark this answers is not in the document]
makesitgo: I hear you, but we could also accommodate and assert on these captured vars as a slice of one item just as easily.  that way, we still truer to the way the code actually operates, and if/when a test case ever is modified or added that does get us calling `Delete`/`Find` more than once, then we're already built to support that

tsedgwick (approved):
LGTM, once Nick's comments get addressed.

nice work!

internal/commands/user/delete.go
```diff
@@ -0,0 +1,152 @@
...
+	return nil
+}
+
+// Feedback is the command feedback
+func (cmd *CommandDelete) Feedback(profile *cli.Profile, ui terminal.UI) error {
+	if len(cmd.outputs) == 0 {
+		return ui.Print(terminal.NewTextLog("No users to delete"))
+	}
+
+	var outputByProviderType = map[string][]userOutput{}
```
makesitgo, line 86: [meganit] rename this to `outputsByProviderType`

internal/commands/user/delete.go
```diff
@@ -0,0 +1,152 @@
...
+	for _, output := range cmd.outputs {
+		for _, identity := range output.user.Identities {
+			outputByProviderType[identity.ProviderType] = append(outputByProviderType[identity.ProviderType], output)
+		}
+	}
+	logs := make([]terminal.Log, len(outputByProviderType))
```
makesitgo, line 92: to avoid the `logIndex`, you could use `make([]terminal.Log, 0, len(outputsByProviderType))` here and then use `append` in the for loop

internal/commands/user/delete.go
```diff
@@ -0,0 +1,152 @@
...
+func userDeleteTableRows(providerType string, outputs []userOutput) []map[string]interface{} {
+	userDeleteTableRows := make([]map[string]interface{}, 0, len(outputs))
+	for _, output := range outputs {
+		userDeleteTableRows = append(userDeleteTableRows, userDeleteTableRow(providerType, output))
+	}
+	return userDeleteTableRows
+}
+
+func userDeleteTableRow(providerType string, output userOutput) map[string]interface{} {
+	var msg string
```
makesitgo, line 135: can we align this name with the header const name associated with it?  can use `var details string` here, or something that achieves the same?  (if you go the "message" route, i'd avoid using `msg` at that point)

internal/commands/user/delete_test.go
```diff
@@ -0,0 +1,309 @@
...
+		assert.Equal(t, projectID, capturedFindProjectID)
+		assert.Equal(t, appID, capturedFindAppID)
+		assert.Equal(t, projectID, capturedDeleteProjectID)
+		assert.Equal(t, appID, capturedDeleteAppID)
+		assert.Equal(t, testUsers[0].ID, cmd.inputs.Users[0])
+		assert.Nil(t, cmd.outputs[0].err)
+		assert.Equal(t, cmd.outputs[0].user, testUsers[0])
+	})
+
+	t.Run("should save failed deletion errors", func(t *testing.T) {
```
makesitgo, line 93: can we test table these two tests?  I think you could achieve this pretty easily with

```go
struct{
  description string
  userDeleteErr error
  expectedOutputs []userOutput
}
```

also, use the asserts from the test above, they are a bit more specific

internal/commands/user/delete_test.go
```diff
@@ -0,0 +1,309 @@
...
+				[]string{
+					"01:23:45 UTC INFO  Provider type: local-userpass",
+					"  Email            ID      Type    Deleted  Details     ",
+					"  ---------------  ------  ------  -------  ------------",
+					"  user-1@test.com  user-1  type-1  false    client error",
+					"",
+				},
+				"\n",
+			),
+		},
```
makesitgo, line 219: can we have a "many" test case as well with multiple provider types, some error'd some not?

internal/commands/user/flags.go
```diff
@@ -9,22 +16,31 @@ const (
...
 )
 
 const (
-	providerTypeLocalUserPass = "local-userpass"
-	providerTypeAPIKey        = "api-key"
-	providerTypeFacebook      = "oauth2-facebook"
-	providerTypeGoogle        = "oauth2-google"
-	providerTypeAnonymous     = "anon-user"
-	providerTypeCustom        = "custom-token"
+	providerTypeLocalUserPass  = "local-userpass"
```
makesitgo, line 36: [nit] rename this to `providerTypeUserPassword`

internal/commands/user/flags.go
```diff
@@ -9,22 +16,31 @@ const (
...
+	providerTypeAPIKey         = "api-key"
+	providerTypeFacebook       = "oauth2-facebook"
+	providerTypeGoogle         = "oauth2-google"
+	providerTypeAnonymous      = "anon-user"
+	providerTypeCustom         = "custom-token"
```
makesitgo, line 41: [question] did you get this `Custom` name from somewhere?  wondering if we want it to be `CustomToken` to avoid any ambiguity with `CustomFunction`.  what do you think?

internal/commands/user/flags.go
```diff
@@ -35,5 +51,101 @@ var (
 		providerTypeGoogle,
 		providerTypeAnonymous,
 		providerTypeCustom,
+		providerTypeApple,
+		providerTypeCustomFunction,
+	}
+)
+
+// statusType enumset
```
makesitgo, line 59: two contradicting things here:

where did we land on bringing this stuff into the `realm` package?  there was a comment resolved about this, but no response and seemingly no movement here?

if we're keeping this stuff here, then I think we can remove a lot of these docstrings (at least on the unexported types and their functions, the most uninteresting ones)

internal/commands/user/flags.go
```diff
@@ -35,5 +51,101 @@ var (
...
+
+	if !isValidStatusType(newStatusType) {
+		return errInvalidStatusType
+	}
+
+	*st = newStatusType
+	return nil
+}
+
+// set of statusType(s)
```
makesitgo, line 80: [nit] can remove this or `// set of supported user status types` (stick to using the `type name` vs the `typeName`, if you catch my drift)

internal/commands/user/flags.go
```diff
@@ -35,5 +51,101 @@ var (
...
+const (
+	statusTypeEmpty     statusType = ""
+	statusTypeConfirmed statusType = "confirmed"
+	statusTypePending   statusType = "pending"
+)
+
+// statusType error msg
```
makesitgo, line 87: [nit] `// set of known user errors` (and consolidate the other errors into this `var` block)

internal/commands/user/flags.go
```diff
@@ -35,5 +51,101 @@ var (
...
+var (
+	errInvalidStatusType = func() error {
+		allStatusTypes := []string{statusTypeConfirmed.String(), statusTypePending.String()}
+		return fmt.Errorf("unsupported value, use one of [%s] instead", strings.Join(allStatusTypes, ", "))
+	}()
+)
+
+// isValidStatusType checks if a valid statusType
```
makesitgo, line 95: also feels like we could get rid of this docstring

internal/commands/user/input_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+	"github.com/Netflix/go-expect"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+func TestResolveUsersInputs(t *testing.T) {
+	testUsers := []realm.User{
+		{
+			ID: "user-1",
+			Identities: []realm.UserIdentity{
```
makesitgo, line 19: can we inline the `Identities` declaration to a one-liner, just to reduce some line heft here?

internal/commands/user/input_test.go
```diff
@@ -0,0 +1,176 @@
...
+		for _, tc := range []struct {
+			description   string
+			inputs        usersInputs
+			procedure     func(c *expect.Console)
+			users         []realm.User
+			expectedUsers []realm.User
+		}{
+			{
+				description: "with no input set",
+				inputs:      usersInputs{},
```
makesitgo, line 50: [meganit] remove this line

internal/commands/user/input_test.go
```diff
@@ -0,0 +1,176 @@
...
+				assert.Nil(t, consoleErr)
+				defer console.Close()
+
+				doneCh := make(chan (struct{}))
+				go func() {
+					defer close(doneCh)
+					tc.procedure(console)
+				}()
+
+				app := realm.App{
```
makesitgo, line 113: [question] do we even need real values for `ID` and `GroupID` here?  could this just be `var app realm.App`?

internal/commands/user/input_test.go
```diff
@@ -0,0 +1,176 @@
...
+	})
+	t.Run("setup should error", func(t *testing.T) {
+		for _, tc := range []struct {
+			description string
+			inputs      usersInputs
+			procedure   func(c *expect.Console)
+			expectedErr error
+			mockClient  func() realm.Client
+		}{
+			{
```
makesitgo, line 136: i don't think this wants to be it's own test table anymore?  only one here...

on this note though, you probably could add this right into the above table with some tweaking of your `description` strings.  just have fields for `users, usersErr, expectedUsers, expectedErr` in the test case struct

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,120 @@
...
+type usersInputs struct {
+	State         realm.UserState
+	ProviderTypes []string
+	Status        statusType
+	Users         []string
+}
+
+// ResolveUsers will use the provided Realm client to resolve the users specified by the realm.App through inputs
+func (i *usersInputs) ResolveUsers(ui terminal.UI, client realm.Client, app realm.App) ([]realm.User, error) {
+	if len(i.Users) > 0 {
```
makesitgo, line 28: where did we land on this filter logic being similar to that of the `realm.Client`'s own `FindUsers` function?  I imagined we would have been able to simply convert our input fields to a `realm.UserFilter` here and call it day?

seems like we could fetch all the users here up top, and then if `len(i.Users) > 0` we can just return that list, otherwise we can prompt the user with a multiselect

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,120 @@
...
+		foundUsers, err := client.FindUsers(app.GroupID, app.ID, userFilter)
+		if err != nil {
+			return nil, err
+		}
+		for _, user := range foundUsers {
+			selectableUsersSet[user.ID] = user
+		}
+	}
+
+	// Interactive User Selection
```
makesitgo, line 68: [nit] remove this line

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,120 @@
...
+	selectableUsers := map[string]realm.User{}
+	selectableUserOptions := make([]string, len(selectableUsersSet))
```
makesitgo, line 70: same thought here about initializing this to have 0 length but the right capacity, and use `append` below

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,120 @@
...
+	patternShort := "%s - %s"
+	patternLong := "%s - %s - %s"
+	userOptIndex := 0
+	for _, user := range selectableUsersSet {
+		var opt string
+		if len(user.Identities) == 0 {
```
makesitgo, line 76: seems like `opt` could be assigned just the values of `"Unknown", or "Anonymous", etc and the whole `fmt.Sprintf` bit can be moved down below

that or maybe even we can refactor this bit into a helper method to reduce some visual noise here, and then we can also add a few unit tests on that

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,120 @@
...
+				opt = fmt.Sprintf(patternShort, "Custom Function", user.ID)
+			default:
+				opt = fmt.Sprintf(patternShort, "Unknown", user.ID)
+			}
+		}
+		selectableUserOptions[userOptIndex] = opt
+		selectableUsers[opt] = user
+		userOptIndex++
+	}
+	selectedUsers := []string{}
```
makesitgo, line 104: [personal style] i prefer `var selectedUsers []string` for this type of initialization

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,120 @@
...
+	err := ui.AskOne(
```
makesitgo, line 105: you can inline this `err` to this if statement below just as is (even with the multiline).  that use a name here other than `err` as it sits in the function scope

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,120 @@
...
+		&survey.MultiSelect{
+			Message: "Which user(s) would you like to delete?",
+			Options: selectableUserOptions,
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+	users := make([]realm.User, len(selectedUsers))
+	for userIndex, userOption := range selectedUsers {
```
makesitgo, line 116: [meganit] `s/userIndex/i`, `s/userOption/user`

makesitgo (commented):
bunch of thoughts here, apologies for the super cosmetic/ocd stuff!  we can chat about some of the larger topics offline here for a bit first

lookin real good though!  nice work so far

internal/cloud/realm/user.go
```diff
@@ -92,11 +93,139 @@ type User struct {
 // UserIdentity is a Realm application user identity
 type UserIdentity struct {
 	UID          string                 `json:"id"`
-	ProviderType string                 `json:"provider_type"`
+	ProviderType ProviderType           `json:"provider_type"`
 	ProviderID   primitive.ObjectID     `json:"provider_id"`
 	ProviderData map[string]interface{} `json:"provider_data,omitempty"`
 }
 
+// ProviderType is a Realm application provider type
```
makesitgo, line 101: [nit/style] prefer to use "Realm app" when bringing up apps.  this was newly declared, so feel free to drive-by/update any incorrect instances you find (no need to seek them out though)

also, just within this package, let's rename this (and associated `const`s) to `AuthProviderType`

internal/cloud/realm/user.go
```diff
@@ -92,11 +93,139 @@ type User struct {
...
+type ProviderType string
+
+// set of supported provider type values
+const (
+	ProviderTypeEmpty          ProviderType = ""
+	ProviderTypeUserPassord    ProviderType = "local-userpass"
```
makesitgo, line 107: [typo] `Password`

internal/cloud/realm/user.go
```diff
@@ -92,11 +93,139 @@ type User struct {
...
+	ProviderTypeAPIKey         ProviderType = "api-key"
+	ProviderTypeFacebook       ProviderType = "oauth2-facebook"
+	ProviderTypeGoogle         ProviderType = "oauth2-google"
+	ProviderTypeAnonymous      ProviderType = "anon-user"
+	ProviderTypeCustomToken    ProviderType = "custom-token"
+	ProviderTypeApple          ProviderType = "oauth2-apple"
+	ProviderTypeCustomFunction ProviderType = "custom-function"
+)
+
+// slice of supported provider type values as string slice
```
makesitgo, line 117: either reword this to be more general: `// set of supported auth constants`, or move the comment down to be right above `ValidProviderTypes`, and make it something like `// ValidProviderTypes is the set of supported...`

I lean towards the former approach, but your call here

internal/cloud/realm/user.go
```diff
@@ -92,11 +93,139 @@ type User struct {
...
+		return "Custom JWT"
+	case ProviderTypeCustomFunction:
+		return "Custom Function"
+	default:
+		return "Unknown"
+	}
+}
+
+// IsValid validates the provided provider type
+func (pt ProviderType) IsValid() error {
```
makesitgo, line 159: [nit] the function name here makes me think the function should return `bool` here.  is the error actually needed/useful here (i can't find any usages here)

internal/cloud/realm/user.go
```diff
@@ -92,11 +93,139 @@ type User struct {
...
+		ProviderTypeCustomToken,
+		ProviderTypeCustomFunction:
+		return nil
+	default:
+		return errors.New("Invalid ProviderType")
+	}
+}
+
+// DisplayUser returns the display string for a user and provider type
+func (pt ProviderType) DisplayUser(user User) string {
```
makesitgo, line 176: any chance these "user display" functions (here and below) could live in the `commands/user` package as just unexported helper functions (assuming that's where they're used)?

internal/cloud/realm/user_test.go
```diff
@@ -85,7 +87,7 @@ func TestRealmUsers(t *testing.T) {
 			})
 
 			t.Run("And find a certain type of user", func(t *testing.T) {
-				users, err := client.FindUsers(groupID, app.ID, realm.UserFilter{Providers: []string{"local-userpass"}})
+				users, err := client.FindUsers(groupID, app.ID, realm.UserFilter{Providers: []realm.ProviderType{"local-userpass"}})
```
makesitgo, line 90: can you reference the `const` we have for these values now? (here and below)

internal/cloud/realm/user_test.go
```diff
@@ -140,3 +142,216 @@ func TestRealmUsers(t *testing.T) {
...
+		assert.Equal(t, realm.ProviderType("invalid type").IsValid(), errors.New("Invalid ProviderType"))
+	})
+}
+
+func TestProviderTypeDisplay(t *testing.T) {
+	for _, tc := range []struct {
+		pt             realm.ProviderType
+		expectedOutput string
+	}{
+		{
```
makesitgo, line 162: for the relatively simple test case struct here, could you clean up some line heft here and leave these defined inline? 

```go
  {realm.AuthProviderTypeAnonymou, "Anonymous"},
  {realm.AuthProviderTypeUserPassword, "User/Password"},
  // etc
```

internal/commands/user/delete.go
```diff
@@ -0,0 +1,149 @@
...
+		headerID,
+		headerType,
+		headerDeleted,
+		headerDetails,
+	)
+	return headers
+}
+
+func userDeleteTableRows(providerType realm.ProviderType, outputs []userOutput) []map[string]interface{} {
+	userDeleteTableRows := make([]map[string]interface{}, 0, len(outputs))
```
makesitgo, line 124: [meganit] rename to just `rows`

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,73 @@
...
+		Pending:   i.Pending,
+		Providers: realm.StringSliceToProviderTypes(i.ProviderTypes...),
+	}
+	foundUsers, usersErr := client.FindUsers(app.GroupID, app.ID, filter)
+	if usersErr != nil {
+		return nil, usersErr
+	}
+	if len(i.Users) > 0 {
+		if len(foundUsers) == 0 {
+			return nil, errors.New("No users found")
```
makesitgo, line 39: don't capitalize the error strings, we're often appending the messages and keep most things in lower case

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		return foundUsers, nil
+	}
+
+	selectableUsers := map[string]realm.User{}
+	selectableUserOptions := make([]string, len(foundUsers))
+	for idx, user := range foundUsers {
+		var pt realm.ProviderType
+		if len(user.Identities) == 0 {
+			pt = ""
```
makesitgo, line 49: shouldn't need this line (the zero value of `realm.AuthProviderType` is already this from the `var` declaration above

internal/commands/user/list.go
```diff
@@ -80,30 +80,30 @@ func (cmd *CommandList) Handler(profile *cli.Profile, ui terminal.UI) error {
...
 	for _, user := range cmd.users {
 		for _, identity := range user.Identities {
 			usersByProviderType[identity.ProviderType] = append(usersByProviderType[identity.ProviderType], user)
 		}
 	}
 
 	var logs []terminal.Log
-	for providerType, users := range usersByProviderType {
+	for _, pt := range realm.ValidProviderTypes {
+		providerType := realm.ProviderType(pt)
```
makesitgo, line 98: [question] why this change?  couldn't you keep just iterating over `usersByProviderType` here?

internal/commands/user/delete_test.go
```diff
@@ -0,0 +1,339 @@
...
+		userDeleteErr   error
+		expectedOutputs []userOutput
+	}{
+		{
+			description:   "should delete a user when a user id is provided",
+			userDeleteErr: nil,
+			expectedOutputs: []userOutput{
+				{
+					user: testUsers[0],
+					err:  nil,
```
makesitgo, line 55: [nit] remove this line

makesitgo (commented):
just a few final thoughts/questions here

internal/cloud/realm/user.go
nathantfrank: I prefer to use error in situations where the type has validation on it. If it was more optional I tend to agree on going with bool.

internal/commands/user/delete_test.go
nathantfrank: done, see below

internal/commands/user/inputs.go
nathantfrank: addressed

internal/commands/user/inputs.go
nathantfrank: addressed

internal/commands/user/list.go
nathantfrank: Maps, do not guarantee ordering. Was getting the occasional test failure.

nathantfrank: If we don't care about ordering, I could look to see about making the test work for any type of ordering

internal/cloud/realm/user.go
makesitgo: Point being if you feel the error is necessary, let's rename the function to something like `Validate`.  If you want to keep it as `IsValid`, then we should stick to returning a `bool` here.

But my bigger question is: what uses this / why is this needed?  I cannot find any usages of it outside of a test file.  Let's remove it all together if we don't need it.  Otherwise, it'd be nice to keep it similar to how we've approached this pattern with `UserState` above (and keep the `isValidUserState` function unexported and used only here)

internal/commands/user/list.go
makesitgo: ah completely forgot we were relying on deterministic output for the tests.  i like the purpose behind this change then, although I'd prefer we didnt do this `providerType := realm.ProviderType(pt)` on the first line of each loop.

what would it take to keep `realm.ValidProviderTypes` to be typed as `realm.AuthProvderTypes`?  a brief look on my end seems to indicate its possible, but just might mean we have to implement a `func (types AuthProviderTypes) Strings() []string` method, or adjust the `flags.EnumSetValue` to hold onto `validValues` as a `[]interface{}` instead of string array.  thoughts?

internal/cloud/realm/user.go
nathantfrank: talked offline and decided to remove

nathantfrank:
@makesitgo RFAL

internal/cloud/realm/user.go
```diff
@@ -89,14 +89,96 @@ type User struct {
...
+	if len(apts) == 0 {
+		return ""
+	}
+	bSep := []byte(sep)
+	out := make([]byte, 0, (1+len(bSep))*len(apts))
+	for _, s := range apts {
+		out = append(out, s...)
+		out = append(out, bSep...)
+	}
+	return string(out[:len(out)-len(sep)])
```
makesitgo, line 179: lets reduce some of the indexing logic here and use a `strings.Builder`:

```go
var sb strings.Builder
for i, apt := range apts {
  if i != 0 { sb.WriteString(sep) }
  sb.WriteString(apt.String())
}
return sb.String()
```

internal/cloud/realm/user_ext_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+		t.Run("And finding pending users should return an empty list", func(t *testing.T) {
+			users, err := client.FindUsers(groupID, app.ID, realm.UserFilter{Pending: true})
+			assert.Nil(t, err)
+			assert.Equal(t, []realm.User{}, users)
+		})
+	})
+}
+
+func TestAuthProviderTypeDisplay(t *testing.T) {
```
makesitgo, line 145: I would move this and `TestNewAuthProviderTypes` to the `user_test.go` file instead

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/AlecAivazis/survey/v2"
+)
+
+const (
+	userDataEmail = "email"
+	userDataName  = "name"
+)
+
+func validAuthProviderTypes() []interface{} {
```
makesitgo, line 18: [optional/future drive-by] might be worth having this be a pre-computed `var` eventually:

```go
var validAuthProviderTypes = func() []interface{} { /* make the []interface{} here */}()
```

since it's not changing much

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,110 @@
...
+	apts := make([]interface{}, 0, len(realm.ValidAuthProviderTypes))
+	for _, apt := range realm.ValidAuthProviderTypes {
+		apts = append(apts, apt)
+	}
+	return apts
+}
+
+func displayUser(apt realm.AuthProviderType, user realm.User) string {
```
makesitgo, line 26: i think we can simplify this and `displayUserData` methods here a bit.  something like:

```go
func displayUser(apt relam.AuthProviderType, user realm.User) string {
  var sb strings.Builder
  sb.WriteString(apt.Display() + " - ")
  if data := displayUserData(apt, user); user != "" {
    sb.WriteString(data + " - ") // probably also worth moving " - " into a `const` somewhere, maybe terminal.DelimiterInline
  }
  sb.WriteString(user.ID)
  return sb.String()
}

func displayUserData(apt realm.AuthProviderType, user realm.User) string {
  // return the "" if no match on provider type for now
}
```

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return ""
+	}
+	if displayUserData != "" {
+		display += sep + displayUserData
+	}
+	return display + sep + user.ID
+}
+
+func displayUserData(apt realm.AuthProviderType, user realm.User) (string, error) {
```
makesitgo, line 39: especially as this is just for display, I'd rather be a little resilient and just omit/display empty here

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,110 @@
...
+	var val interface{}
+	ok := false
```
makesitgo, line 41: [meganit] `var ok bool`

internal/commands/user/inputs.go
```diff
@@ -0,0 +1,110 @@
...
+	switch apt {
+	case realm.AuthProviderTypeUserPassword:
+		val, ok = user.Data["email"]
+	case realm.AuthProviderTypeAPIKey:
+		val, ok = user.Data["name"]
+	default:
```
makesitgo, line 47: if we go the `return ""` route, then this `default` case could be omitted

internal/utils/flags/flags.go
```diff
@@ -107,7 +108,14 @@ func (esv *EnumSetValue) validateAndRemoveDuplicates() error {
 }
 
 func (esv *EnumSetValue) errInvalidEnumValue() error {
-	return fmt.Errorf(`unsupported value, use one of ["%s"] instead`, strings.Join(esv.validValues, `", "`))
+	bSep := []byte(`", "`)
+	out := make([]byte, 0, (1+len(bSep))*len(esv.validValues))
```
makesitgo, line 112: same thing about removing the indexing logic here and just use a `strings.Builder` for concatenation here

makesitgo (approved):
alright, rest assured these are my last comments here 🙃  ... everything left is fairly straightforward so once that's up and evergreen is back green, this is good to go

🆒  beans

PR #50, opened 2021-01-19
REALMC-7169: Add push command

makesitgo:
~WIP: Need to add TODOs and write more tests~

Apologies for this one...


internal/app/structure.go
```diff
@@ -0,0 +1,136 @@
...
+	if err := unmarshalDirectoryInto(appDir.Path, NameTriggers, pkg); err != nil {
+		return nil, err
+	}
+	if err := unmarshalGraphQLInto(appDir, pkg); err != nil {
+		return nil, err
+	}
+	if err := unmarshalServicesInto(appDir, pkg); err != nil {
+		return nil, err
+	}
+	return pkg, nil
```
kpatel71716, line 135: [q] is it intentional that this code duplicates the readAppStructureV1 function above?(apologies if I totally missed something here)

kpatel71716 (commented):
Looks really good! I might ask you some questions offline just for my understanding but this makes sense to me

internal/app/structure.go
makesitgo: Yeah, so this code can definitely still be considered a WIP for now (I meant to add a TODO, so thanks for asking this).

For some context, the new config version exports some data in different spots now (e.g. `sync.dev_mode_enabled` was moved into a `sync/config.json`), but the import route expects a different structure when processing requests.

The code paths are split here for now to support how the old CLI did it ("V1") and this new path ("V2") where I'm not _quite_ sure where it's headed still.  We have [a ticket](https://jira.mongodb.org/browse/REALMC-7865) planned to work out a lot of this transformation logic on server-side, so once that settles we should be able to easily update this and/or merge the two paths all together

kpatel71716: Perfect, that works for me thanks!

internal/cloud/atlas/errors.go
```diff
@@ -0,0 +1,59 @@
...
+	if err.reason == "" {
+		return errCommonServerError
+	}
+	return fmt.Sprintf("%s: %s", errCommonServerError, err.reason)
+}
+
+// ErrMissingAuth is a missing auth error
+type ErrMissingAuth struct{}
+
+func (err ErrMissingAuth) Error() string { return "must provide auth details" }
```
UnicodeSnowman, line 36: same [q] here

UnicodeSnowman (approved):
LGTM

internal/app/inputs.go
```diff
@@ -63,6 +64,28 @@ func (i *ProjectInputs) Resolve(ui terminal.UI, wd string) error {
...
+	return errMsg
+}
+
+// ErrGroupNotFound is a group not found error
+type ErrGroupNotFound struct {
+}
+
+func (err ErrGroupNotFound) Error() string {
+	return "failed to find group"
+}
```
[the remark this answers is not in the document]
makesitgo: nope, just some copy/paste laziness here.  will switch both of these

PR #51, opened 2021-01-19
REALMC-7163 build out secrets list command

kpatel71716:


internal/commands/secrets/list.go
```diff
@@ -0,0 +1,112 @@
...
+	timeString := "n/a"
+	if secret.LastModified != 0 {
+		timeString = time.Unix(secret.LastModified, 0).UTC().String()
+	}
+	row := map[string]interface{}{
+		headerName:         secret.Name,
+		headerID:           secret.ID,
+		headerLastMofified: timeString,
+		headerSecret:       secret.Secret,
+	}
```
kpatel71716: I know the spec mentioned to output both `ID` and `Name`. I was wondering if there is any value to showing `Last Modified` and, more importantly, an indicator showing if the value is a secret or not. I assumed that we would show both here but I may have missed that

internal/cloud/realm/secrets.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Value is a value or secret stored in a Realm application
+type Value struct {
+	ID           string `json:"_id"`
+	Name         string `json:"name"`
+	LastModified int64  `json:"last_modified"`
+	Secret       bool   `json:"from_secret"`
+}
+
+func (c *client) FindValues(app App) ([]Value, error) {
```
kpatel71716, line 23: I used `Value` for the variable name and `FindValues` for the function name since the endpoint retrieves all values. I maintained this throughout the list command as well. Wanted to get thoughts on this since I felt secret just another type of value in this scenario and I am displaying both

internal/commands/secrets/list.go
```diff
@@ -0,0 +1,112 @@
...
+func secretTableRow(value realm.Value) map[string]interface{} {
+	timeString := "n/a"
+	if value.LastModified != 0 {
+		timeString = time.Unix(value.LastModified, 0).UTC().String()
+	}
+	row := map[string]interface{}{
+		headerName:         value.Name,
+		headerID:           value.ID,
+		headerLastMofified: timeString,
+		headerSecret:       value.Secret,
```
kpatel71716, line 101: The spec mentions to display `Name` and `ID` but was wondering if there was value in the other two fields. Most importantly the `Secret` bool indicating if a value is a secret or not. My understanding was that this command could list both 'Values' and 'Secrets'. Though I am not completely sure if the difference means anything to a user. 

I do think it might be strange to display an `Is Secret` field to a user if secrets and values aren't that different (and would make sense if that is why the current cli does not do so)

internal/cloud/realm/secrets.go
makesitgo: yeah, I like `Value` here for consistency with how we're calling it on the api/server side

internal/cloud/realm/secrets.go
```diff
@@ -0,0 +1,42 @@
...
+	res, resErr := c.do(
+		http.MethodGet,
+		fmt.Sprintf(valuesPathPattern, app.GroupID, app.ID),
+		api.RequestOptions{},
+	)
+	if resErr != nil {
+		return nil, resErr
+	}
+
```
makesitgo, line 32: with the recent refactor, we actually have a standardized approach to this scenario now:

```go
res, resErr := c.do(...)
if resErr != nil {
  return nil, resErr
}
  return nil, api.ErrUnexpectedStatusCode{"get values", res.StatusCode}
}
defer res.Body.Close()
```

kpatel71716: Sounds good, I can rebase once the push PR goes in. I imagine there will other standardizations to integrate as well

internal/commands/commands.go
```diff
@@ -76,4 +77,21 @@ var (
...
+		Aliases:     []string{"secret"},
+		Description: "Manage the secrets of your MongoDB Realm app",
+		Help:        "secret", // TODO(REALMC-7429): add help text description
+		SubCommands: []cli.CommandDefinition{
+			cli.CommandDefinition{
+				Use:         "list",
+				Aliases:     []string{"ls"},
+				Description: "List the names of secrets in your Realm app",
+				// TODO(REALMC-7429): add help text description
+				Help:    "Displays a list of data tables. Each data table displays the Name and ID of one (1) secret.",
```
makesitgo, line 92: check out [here](https://docs.google.com/document/d/14fBZaN47HRe3-2Pk1uSfYcCxX3Dd95wK0tprS3WE1vk/edit) for better starter values with `Help` and `Description`

internal/commands/commands.go
```diff
@@ -76,4 +77,21 @@ var (
...
+	Secrets = cli.CommandDefinition{
+		Use:         "secrets",
+		Aliases:     []string{"secret"},
+		Description: "Manage the secrets of your MongoDB Realm app",
+		Help:        "secret", // TODO(REALMC-7429): add help text description
+		SubCommands: []cli.CommandDefinition{
+			cli.CommandDefinition{
+				Use:         "list",
+				Aliases:     []string{"ls"},
+				Description: "List the names of secrets in your Realm app",
```
makesitgo, line 90: you'll want `Display` set here as well, to `secrets list`

internal/commands/secrets/list.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// Handler is the command handler
+func (cmd *CommandList) Handler(profile *cli.Profile, ui terminal.UI) error {
+	app, appErr := app.Resolve(ui, cmd.realmClient, cmd.inputs.Filter())
+	if appErr != nil {
+		return appErr
+	}
+
+	secrets, FindSecretsErr := cmd.realmClient.FindSecrets(app)
```
makesitgo, line 48: [nit] just rename this error to `findErr` or `secretsErr`

internal/cloud/realm/client.go
```diff
@@ -31,6 +31,8 @@ type Client interface {
 	DeleteApp(groupID, appID string) error
 	FindApps(filter AppFilter) ([]App, error)
 
+	FindSecrets(app App) ([]Secret, error)
```
makesitgo, line 34: change this signature to accept `groupID, appID string` (similar to the other method signatures here for this interface)

internal/commands/secrets/list.go
```diff
@@ -0,0 +1,98 @@
...
+	var logs []terminal.Log
+	logs = append(logs, terminal.NewTableLog(
+		fmt.Sprintf("Found %d secrets", len(cmd.secrets)),
+		secretTableHeaders(),
+		secretTableRows(cmd.secrets)...,
+	))
+	return ui.Print(logs...)
+}
+
+func secretTableHeaders() []string {
```
makesitgo, line 71: this is fine here for now, since we don't have any other list commands.  but considering we're in the `secrets` package here, lets just rename these to be `listSecretsTableHeaders` and `listSecretsTableRows`.

---

also, for the headers in particular here, it doesn't need to be a function, just have:

```go
var (
  listSecretsTableHeaders = []string{headerID, headerName}
)
```

internal/commands/secrets/list.go
```diff
@@ -0,0 +1,98 @@
...
+	headers = append(
+		headers,
+		headerID,
+		headerName,
+	)
+	return headers
+}
+
+func secretTableRows(secrets []realm.Secret) []map[string]interface{} {
+	secretTableRows := make([]map[string]interface{}, 0, len(secrets))
```
makesitgo, line 82: [nit] rename this to just `rows`

internal/commands/secrets/list.go
```diff
@@ -0,0 +1,98 @@
...
+	for _, secret := range secrets {
+		secretTableRows = append(secretTableRows, map[string]interface{}{
+			headerName: secret.Name,
+			headerID:   secret.ID,
+		})
+	}
+	return secretTableRows
+}
+
+func (i *listInputs) Resolve(profile *cli.Profile, ui terminal.UI) error {
```
makesitgo, line 92: just looking at this, and now having seen this pattern occur quite a bit, what do you think about maybe tweaking the signature of `ProjectInputs.Resolve` to match `InputResolver.Resolve`?

If we do that, then we can omit all the cases where we define an `inputs` struct just to hold `ProjectInputs`.  I'm sure there was a reason I opted to not do that in the first place, but I can't quite recall.  Do you mind trying that out and seeing what kinda of "gotchas" we run into?

internal/commands/secrets/list_test.go
```diff
@@ -0,0 +1,181 @@
...
+		ID:          appID,
+		GroupID:     projectID,
+		ClientAppID: "eggcorn-abcde",
+		Name:        "eggcorn",
+	}
+
+	t.Run("Should find app secrets", func(t *testing.T) {
+		testSecrets := []realm.Secret{
+			{
+				ID:   "secret1",
```
makesitgo, line 40: [meganit] just inline these definitions to help reduce the visual load of this test

internal/commands/secrets/list_test.go
```diff
@@ -0,0 +1,181 @@
...
+		{
+			description:    "Should indicate no secrets found when none are found",
+			secrets:        []realm.Secret{},
+			expectedOutput: "01:23:45 UTC INFO  No available secrets to show\n",
+		},
+		{
+			description: "Should display all found secrets",
+			secrets: []realm.Secret{
+				{
+					ID:   "60066e14734d0b6c336ffc23",
```
makesitgo, line 141: [meganit] inline these definitions to help reduce visual load here

makesitgo (commented):
nothing major from me, but here's a first pass of a few comments/questions for ya.  nice work so far!

internal/commands/secrets/list.go
kpatel71716: @makesitgo I was able to make the switch without running into any noticeable gotchas. The biggest reason against it would probably just be needing to import the cli package in internal/app/inputs.go. The latest commit in the PR has the specific changes if you want to take a look at what I modified

internal/commands/commands.go
kpatel71716: The file should have the text from the linked google doc. I wasn't sure if the text was final and I can remove the TODO if so (left it as a placeholder for now)

makesitgo: I see...yeah feel free to remove the TODO then, it's the best we got for now.  thanks!

internal/commands/secrets/list.go
makesitgo: that looks right as far as the refactor goes, but let's actually keep this as it was.  we won't always be resolving app data from the `WorkingDirectory`, so that `string` will be useful.  thanks for testing all this out here! glad we reached some sanity here

internal/cloud/realm/client.go
```diff
@@ -31,6 +31,8 @@ type Client interface {
 	DeleteApp(groupID, appID string) error
 	FindApps(filter AppFilter) ([]App, error)
 
+	FindSecrets(groupID, appID string) ([]Secret, error)
```
makesitgo, line 34: I usually reserve the `Find` word when a call has some form of options/filter to it, and generally prefixing a method with `Get` is not very go-like...  let's go with `ListSecrets` here

makesitgo (commented):
just a few last things, and after you revert the app inputs Resolve refactor this should be good to go

internal/commands/secrets/list.go
kpatel71716: Done! Reverted. I think the string is useful here as well

internal/cloud/realm/client.go
kpatel71716: Changed to `Secrets` as discussed offline

internal/commands/secrets/list.go
```diff
@@ -0,0 +1,80 @@
...
+	logs = append(logs, terminal.NewTableLog(
+		fmt.Sprintf("Found %d secrets", len(cmd.secrets)),
+		listSecretsTableHeaders,
+		listSecretTableRows(cmd.secrets)...,
+	))
+	return ui.Print(logs...)
+}
+
+var (
+	listSecretsTableHeaders = []string{headerID, headerName}
```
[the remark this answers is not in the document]
kpatel71716: I agree with this also. I prefer removing the redundancy as we keep that code in the secrets package anyways. As more is added, I will adjust this

makesitgo (approved):
honestly this is good to go... don't worry about the rebase, i'll merge your stuff in on my end and make the appropriate changes

nice work here!

🆒  beans

tsedgwick (approved):
lgtm

PR #52, opened 2021-01-21
REALMC-7161: Buildout command: user disable

nathantfrank:


internal/commands/user/user_state.go
```diff
@@ -0,0 +1,184 @@
...
+
+// NewCommandEnable returns a *CommandUserState which is the `user enable` command
+func NewCommandEnable() *CommandUserState {
+	return &CommandUserState{userEnable: true}
+}
+
+// CommandUserState is the `user enable/disable` command
+type CommandUserState struct {
+	userEnable  bool
+	inputs      userStateInputs
```
makesitgo, line 29: [question] are we collecting flags other than `--user` here?

internal/commands/user/user_state.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	return nil
+}
+
+// Flags is the command flags
+func (cmd *CommandUserState) Flags(fs *pflag.FlagSet) {
+	cmd.inputs.Flags(fs)
+	var flagUsage string
+	if cmd.userEnable {
```
makesitgo, line 51: after seeing how simple these commands end up being (and how similar they are to your previous `user delete` work), I'm leaning towards ditching the abstraction here all together and just defining/implementing `CommandEnable`, `CommandDisable`, `CommandDelete` as their own structs (with their own inputs).

along that same line, we can probably start to abstract away some of the `Feedback` logic we share across many of these commands.  I think that `[]userOutput` => `map` => `[]terminal.Log` flow is fairly similar across the commands

makesitgo (commented):
question/thought for you about these command implementations, also should add an EnableUser test in `realm/user_test.go` (after `"And find specific user using all filter options"`, unfortunately test ordering matters in that file)

internal/commands/user/user_state.go
kpatel71716: This is a good point that I noticed with user revoke as well. I had a couple of comments [in that PR](https://github.com/10gen/realm-cli/pull/54) pointing out the similarities and am happy to adjust anything there as well

nathantfrank: I agree with the `Feedback` refactor to be reusable throughout the command set. 
For ditching the abstraction, I'm not quite in alignment. The main thought I have is following DRY and I'm not seeing any red flags from the abstractions made so far to warrant not following DRY.

internal/commands/user/user_state.go
nathantfrank: AuthProviderTypes for filtering when interactive selecting users.

makesitgo: Do we need the interactive selection?  I'm not sure that spec'd out...for `enable`/`disable`/`revoke`, let's just keep the inputs to `--user` for now

internal/commands/user/user_state.go
makesitgo: this particular DRY effort/abstraction just is not very go-like, if we were working in a language with better generic support this type of approach would fit a little more nicely

internal/commands/user/disable.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Handler is the command handler
+func (cmd *CommandDisable) Handler(profile *cli.Profile, ui terminal.UI) error {
+	app, err := app.Resolve(ui, cmd.realmClient, cmd.inputs.Filter())
+	if err != nil {
+		return err
+	}
+
+	users, usersErr := cmd.inputs.resolveUsers(ui, cmd.realmClient, app)
```
makesitgo, line 70: I don't think we need the extra function here, this could just be:

```go
users, usersErr := client.FindUsers(app.GroupID, app.ID, realm.UserFilter{IDs: cmd.inputs.Users})
```

internal/commands/user/disable.go
```diff
@@ -0,0 +1,121 @@
...
+		cmd.outputs = append(cmd.outputs, userOutput{user: user, err: err})
+	}
+	return nil
+}
+
+// Feedback is the command feedback
+func (cmd *CommandDisable) Feedback(profile *cli.Profile, ui terminal.UI) error {
+	if len(cmd.outputs) == 0 {
+		msg := "No users to disable"
+		return ui.Print(terminal.NewTextLog(msg))
```
makesitgo, line 86: [nit] remove the `msg` var here and one-line everything

internal/commands/user/disable.go
```diff
@@ -0,0 +1,121 @@
...
+			fmt.Sprintf("Provider type: %s", apt.Display()),
+			append(userTableHeaders(apt), headerEnabled, headerDetails),
+			userTableRows(apt, outputs, userDisableRow)...,
+		))
+	}
+	return ui.Print(logs...)
+}
+
+func userDisableRow(output userOutput, row map[string]interface{}) {
+	if output.err != nil && !output.user.Disabled {
```
makesitgo, line 111: same thing here with defining `var`s for `enabled bool` and `details string`.  save the row/map assignments to one place at the end of the func

internal/commands/user/enable.go
```diff
@@ -0,0 +1,121 @@
...
+			fmt.Sprintf("Provider type: %s", apt.Display()),
+			append(userTableHeaders(apt), headerEnabled, headerDetails),
+			userTableRows(apt, outputs, userEnableRow)...,
+		))
+	}
+	return ui.Print(logs...)
+}
+
+func userEnableRow(output userOutput, row map[string]interface{}) {
+	if output.err != nil && output.user.Disabled {
```
makesitgo, line 111: same thing here with defining `var`s for `disabled bool` and `details string`.  save the row/map assignments to one place at the end of the func

internal/commands/user/disable_test.go
```diff
@@ -0,0 +1,324 @@
...
+	"strings"
+	"testing"
+
+	"github.com/10gen/realm-cli/internal/app"
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/10gen/realm-cli/internal/utils/test/mock"
+)
+
+func TestDisableSetup(t *testing.T) {
```
makesitgo, line 14: [nit] keep the test func naming pattern consistent here and use `TestUserDisable*`

internal/commands/user/disable_test.go
```diff
@@ -0,0 +1,324 @@
...
+		profile.SetRealmBaseURL("http://localhost:8080")
+		cmd := &CommandDisable{inputs: disableInputs{}}
+
+		assert.Nil(t, cmd.realmClient)
+		assert.Nil(t, cmd.Setup(profile, nil))
+		assert.NotNil(t, cmd.realmClient)
+	})
+}
+
+func TestDisableHandler(t *testing.T) {
```
makesitgo, line 26: [nit] keep the test func naming pattern consistent here and use `TestUserDisable*`

internal/commands/user/disable_test.go
```diff
@@ -0,0 +1,324 @@
...
+		ID:          appID,
+		GroupID:     projectID,
+		ClientAppID: "eggcorn-abcde",
+		Name:        "eggcorn",
+	}
+	testUsers := []realm.User{
+		{
+			ID:         "user-1",
+			Identities: []realm.UserIdentity{{ProviderType: realm.AuthProviderTypeAnonymous}},
+			Disabled:   false,
```
makesitgo, line 39: [meganit] remove this

internal/commands/user/disable_test.go
```diff
@@ -0,0 +1,324 @@
...
+
+		assert.Nil(t, cmd.Handler(nil, nil))
+		assert.Equal(t, realm.AppFilter{App: appID, GroupID: projectID}, capturedAppFilter)
+		assert.Equal(t, projectID, capturedFindProjectID)
+		assert.Equal(t, appID, capturedFindAppID)
+		assert.Equal(t, projectID, capturedDisableProjectID)
+		assert.Equal(t, appID, capturedDisableAppID)
+		assert.Equal(t, testUsers[0].ID, cmd.inputs.Users[0])
+		assert.Nil(t, cmd.outputs[0].err)
+		assert.Equal(t, cmd.outputs[0].user, testUsers[0])
```
makesitgo, line 86: [nit] can consolidate these last two asserts with:

```go
assert.Equal(t, userOutput{testUsers[0], nil}
```

if you don't go that route, at least swap the position here so its:

```go
assert.Equal(t, testUsers[0], cmd.outputs[0].user)
```

internal/commands/user/disable_test.go
```diff
@@ -0,0 +1,324 @@
...
+	})
+
+	t.Run("should save failed disable errors", func(t *testing.T) {
```
makesitgo, line 89: it seems like this test and the one above could be tabled with just

```go
struct {
  description string
  disableUserErr error
}
```

internal/commands/user/disable_test.go
```diff
@@ -0,0 +1,324 @@
...
+		outputs        []userOutput
+		expectedOutput string
+	}{
+		{
+			description:    "should show indicate no users to disable",
+			outputs:        []userOutput{},
+			expectedOutput: "01:23:45 UTC INFO  No users to disable\n",
+		},
+		{
+			description: "should show 1 failed user, command: user disable",
```
makesitgo, line 218: avoid punctuation in test descriptions (it reads funny in `go test` output) (also the part following the `,` here seems unnecessary?)

internal/commands/user/disable_test.go
```diff
@@ -0,0 +1,324 @@
...
+					"  Email            ID      Type    Enabled  Details     ",
+					"  ---------------  ------  ------  -------  ------------",
+					"  user-1@test.com  user-1  type-1  true     client error",
+					"",
+				},
+				"\n",
+			),
+		},
+		{
+			description: "should show 2 failed users",
```
makesitgo, line 234: [nit] update description to be something more along the lines of "should show failures to disable a user amongst successful results across different auth provider types"

internal/commands/user/disable_test.go
```diff
@@ -0,0 +1,324 @@
...
+		{
+			ID:         "user-5",
+			Identities: []realm.UserIdentity{{ProviderType: realm.AuthProviderTypeCustomToken}},
+			Type:       "type-3",
+		},
+	}
+	for _, tc := range []struct {
+		description    string
+		outputs        []userOutput
+		expectedOutput string
```
makesitgo, line 210: lets try not to overload the `output` word here, I've been using variants of the word `contents` for any terminal display asserts

internal/commands/user/disable_test.go
```diff
@@ -0,0 +1,324 @@
...
+			assert.Equal(t, tc.expectedOutput, out.String())
+		})
+	}
+}
+
+func TestUserDisableRow(t *testing.T) {
+	for _, tc := range []struct {
+		description      string
+		authProviderType realm.AuthProviderType
+		output           userOutput
```
makesitgo, line 278: something's off with this test case struct here, I think ultimately you want to remove `output` field here?

currently, it doesn't look like `tc.authProviderType`, but if you keep that and add `data map[string]interface{}` and `outputErr error` fields, then the `userOutput` struct can be built inside of the test function (rather than here in the `tc` initialization)

internal/commands/user/enable.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Handler is the command handler
+func (cmd *CommandEnable) Handler(profile *cli.Profile, ui terminal.UI) error {
+	app, err := app.Resolve(ui, cmd.realmClient, cmd.inputs.Filter())
+	if err != nil {
+		return err
+	}
+
+	users, usersErr := cmd.inputs.resolveUsers(ui, cmd.realmClient, app)
```
makesitgo, line 70: same thing here about not needing the extra `resolveUsers` func here

internal/commands/user/enable.go
```diff
@@ -0,0 +1,121 @@
...
+		cmd.outputs = append(cmd.outputs, userOutput{user: user, err: err})
+	}
+	return nil
+}
+
+// Feedback is the command feedback
+func (cmd *CommandEnable) Feedback(profile *cli.Profile, ui terminal.UI) error {
+	if len(cmd.outputs) == 0 {
+		msg := "No users to enable"
+		return ui.Print(terminal.NewTextLog(msg))
```
makesitgo, line 86: just inline everything here

internal/commands/user/enable.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	var outputsByProviderType = map[realm.AuthProviderType][]userOutput{}
+	for _, output := range cmd.outputs {
```
makesitgo, line 89: its probably worth defining a func on a `type userOutputs []userOutput`  type alias that can return itself as a `map[realm.AuthProviderType]userOutputs` by now

internal/commands/user/enable_test.go
```diff
@@ -0,0 +1,327 @@
...
+		assert.Equal(t, projectID, capturedFindProjectID)
+		assert.Equal(t, appID, capturedFindAppID)
+		assert.Equal(t, projectID, capturedEnableProjectID)
+		assert.Equal(t, appID, capturedEnableAppID)
+		assert.Equal(t, testUsers[0].ID, cmd.inputs.Users[0])
+		assert.Nil(t, cmd.outputs[0].err)
+		assert.Equal(t, cmd.outputs[0].user, testUsers[0])
+	})
+
+	t.Run("should save failed enable errors", func(t *testing.T) {
```
makesitgo, line 88: i think a lot of the comments i made in `disable_test` are going to end up applying here (so i'll avoid repeating a few of the nits), but this test looks like it can be tabled with the one above

internal/commands/user/list.go
```diff
@@ -169,3 +124,12 @@ func (i *listInputs) Resolve(profile *cli.Profile, ui terminal.UI) error {
 
 	return nil
 }
+
+func userListRow(output userOutput, row map[string]interface{}) {
+	row[headerEnabled] = !output.user.Disabled
```
makesitgo, line 129: [nit] move this down to the other row/map setters

internal/commands/user/output_test.go
```diff
@@ -0,0 +1,84 @@
...
+			},
+			expectedRow: map[string]interface{}{
+				"ID":    "user-1",
+				"Email": "user-1@test.com",
+				"Type":  "type-1",
+			},
+		},
+	} {
+		t.Run(tc.description, func(t *testing.T) {
+			assert.Equal(t, tc.expectedRow, userTableRow(tc.authProviderType, tc.output, func(uo userOutput, m map[string]interface{}) {}))
```
makesitgo, line 81: [meganit] add `row := userTableRow(tc.authProviderType, tc.output, func(uo userOutput, m map[string]interface{}) {})` here just to break up the assert a bit

internal/commands/user/output_test.go
```diff
@@ -0,0 +1,84 @@
...
+		})
+	}
+}
+
+func TestUserTableRow(t *testing.T) {
+	for _, tc := range []struct {
+		description      string
+		userEnable       bool
+		authProviderType realm.AuthProviderType
+		output           userOutput
```
makesitgo, line 38: can probably get by with removing this field and adding a `userData map[string]interface{}` field for the test case struct

makesitgo (commented):
i like a lot of the refactors you have going on here for our `user` commands.  just a bunch of small stuff here from a first pass... really nice work!

internal/cloud/realm/user_ext_test.go
```diff
@@ -123,6 +123,21 @@ func TestRealmUsers(t *testing.T) {
 				assert.Equal(t, []realm.User{email3}, users)
 			})
 
+			t.Run("And enable users", func(t *testing.T) {
+				assert.Nil(t, client.EnableUser(groupID, testApp.ID, email1.ID))
```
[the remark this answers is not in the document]
nathantfrank: Does the test case on line 131 not satisfy this?

internal/commands/user/delete_test.go
```diff
@@ -259,38 +382,15 @@ func TestUserDeleteFeedback(t *testing.T) {
...
+func TestUserDeleteRow(t *testing.T) {
 	for _, tc := range []struct {
 		description      string
 		authProviderType realm.AuthProviderType
 		output           userOutput
 		expectedRow      map[string]interface{}
 	}{
 		{
-			description:      "should show name for apikey type user",
+			description:      "should show successful disable user row",
```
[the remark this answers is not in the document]
nathantfrank: Yes, mistake in the copy pasta

internal/cloud/realm/user_ext_test.go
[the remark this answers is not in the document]
makesitgo: yeah, I meant to elaborate more here (my reviews are getting wonky with this bad internet...)

anyways, let's avoid mutating the test data (i.e. setting Disabled field on email1 and email3). instead, I'd just do explicit finds for each ID and assert that the data coming back is as expected

internal/commands/user/disable.go
```diff
@@ -0,0 +1,106 @@
...
+
+// Setup is the command setup
+func (cmd *CommandDisable) Setup(profile *cli.Profile, ui terminal.UI) error {
+	cmd.realmClient = profile.RealmAuthClient()
+	return nil
+}
+
+// Handler is the command handler
+func (cmd *CommandDisable) Handler(profile *cli.Profile, ui terminal.UI) error {
+	app, err := app.Resolve(ui, cmd.realmClient, cmd.inputs.Filter())
```
makesitgo, line 54: [nit] use `app, appErr`

internal/commands/user/disable.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	users, usersErr := cmd.realmClient.FindUsers(app.GroupID, app.ID, realm.UserFilter{IDs: cmd.inputs.Users})
+	if usersErr != nil {
+		return usersErr
+	}
+
+	for _, user := range users {
+		err := cmd.realmClient.DisableUser(app.GroupID, app.ID, user.ID)
+		cmd.outputs = append(cmd.outputs, userOutput{user: user, err: err})
```
makesitgo, line 66: [meganit] the `user:` and `err:` field declarations are unnecessary here

internal/commands/user/disable.go
```diff
@@ -0,0 +1,106 @@
...
+			fmt.Sprintf("Provider type: %s", apt.Display()),
+			append(userTableHeaders(apt), headerEnabled, headerDetails),
+			userTableRows(apt, outputs, userDisableRow)...,
+		))
+	}
+	return ui.Print(logs...)
+}
+
+func userDisableRow(output userOutput, row map[string]interface{}) {
+	var (
```
makesitgo, line 94: [meganit] just use `var enabled bool` and `var details string` on separate lines here.  50% LOC savings!

internal/commands/user/disable.go
```diff
@@ -0,0 +1,106 @@
...
+		enabled bool
+		details string
+	)
+	if output.err != nil && !output.user.Disabled {
+		enabled = true
+	}
+	if output.err != nil {
+		details = output.err.Error()
+	}
+	row[headerEnabled] = enabled
```
makesitgo, line 104: i didn't notice this before, but shouldn't we use a positive `"Disabled"` header for the `user disable` output?

seems more intuitive to print a true/yes for a successful "disabling" when running this particular command.  thoughts?

based on how that question is answered, there may be an issue with your first if statement above (shouldn't it be `if err == nil`?)

i think ultimately im expecting something like this:

```go
var disabled bool
var details string
if output.err == nil {
  disabled = true
} else {
  details = output.err.Error()
}
row[headerDisabled] = disabled
row[headerDetails] = details
```

internal/commands/user/disable_test.go
```diff
@@ -0,0 +1,270 @@
...
+			Identities: []realm.UserIdentity{{ProviderType: realm.AuthProviderTypeAnonymous}},
+			Disabled:   true,
+		},
+	}
+
+	for _, tc := range []struct {
+		description    string
+		disableUserErr error
+	}{
+		{description: "should disable a user when a user id is provided"},
```
makesitgo, line 51: meganit, if you want to in-line these test case declarartions, omit the field names to reduce text-noise:

```go
{"should disable a user when a user id is provided", nil},
{"should save failed disable errors", errors.New("client error")},
```

internal/commands/user/disable_test.go
```diff
@@ -0,0 +1,270 @@
...
+		},
+		{
+			ID:         "user-5",
+			Identities: []realm.UserIdentity{{ProviderType: realm.AuthProviderTypeCustomToken}},
+			Type:       "type-3",
+		},
+	}
+	for _, tc := range []struct {
+		description     string
+		outputs         []userOutput
```
makesitgo, line 175: [meganit] use the new type here: `userOutputs`

internal/commands/user/disable_test.go
```diff
@@ -0,0 +1,270 @@
...
+		expectedContent string
+	}{
+		{
+			description:     "should show no users to disable",
+			outputs:         []userOutput{},
```
makesitgo, line 180: [meganit] probably should be able to just remove this (i hope out code doesn't care b/w a nil or empty array here)

internal/commands/user/disable_test.go
```diff
@@ -0,0 +1,270 @@
...
+			expectedContent: "01:23:45 UTC INFO  No users to disable\n",
+		},
+		{
+			description: "should show 1 failed user",
+			outputs: []userOutput{
```
makesitgo, line 185: [meganit] use the new type here: `userOutputs`, and the `user:` and `err:` field declarations can go away in the line below

internal/commands/user/disable_test.go
```diff
@@ -0,0 +1,270 @@
...
+					"  ---------------  ------  ------  -------  ------------",
+					"  user-1@test.com  user-1  type-1  true     client error",
+					"",
+				},
+				"\n",
+			),
+		},
+		{
+			description: "should show failures to disable 2 users amongst successful results across different auth provider types",
+			outputs: []userOutput{
```
makesitgo, line 201: [meganit] use the new type here: `userOutputs`, and reduce some text-noise with removing the `user:` and `err:` field declarations below this

internal/commands/user/enable.go
```diff
@@ -0,0 +1,106 @@
...
+
+// Setup is the command setup
+func (cmd *CommandEnable) Setup(profile *cli.Profile, ui terminal.UI) error {
+	cmd.realmClient = profile.RealmAuthClient()
+	return nil
+}
+
+// Handler is the command handler
+func (cmd *CommandEnable) Handler(profile *cli.Profile, ui terminal.UI) error {
+	app, err := app.Resolve(ui, cmd.realmClient, cmd.inputs.Filter())
```
makesitgo, line 54: [nit] `app, appErr` here

internal/commands/user/enable.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	users, usersErr := cmd.realmClient.FindUsers(app.GroupID, app.ID, realm.UserFilter{IDs: cmd.inputs.Users})
+	if usersErr != nil {
+		return usersErr
+	}
+
+	for _, user := range users {
+		err := cmd.realmClient.EnableUser(app.GroupID, app.ID, user.ID)
+		cmd.outputs = append(cmd.outputs, userOutput{user: user, err: err})
```
makesitgo, line 66: [nit] unnecessary `user:` and `err:` declarations here

internal/commands/user/enable.go
```diff
@@ -0,0 +1,106 @@
...
+			fmt.Sprintf("Provider type: %s", apt.Display()),
+			append(userTableHeaders(apt), headerEnabled, headerDetails),
+			userTableRows(apt, outputs, userEnableRow)...,
+		))
+	}
+	return ui.Print(logs...)
+}
+
+func userEnableRow(output userOutput, row map[string]interface{}) {
+	var (
```
makesitgo, line 94: similar thought here as `user disable`.  I'm confused why we have such verbose logic for determining if enable was successful or not, shouldn't it just be the case that if `err` is `nil` then we know the user successfully disabled?

internal/commands/user/enable_test.go
```diff
@@ -0,0 +1,274 @@
...
+			ID:         "user-2",
+			Identities: []realm.UserIdentity{{ProviderType: realm.AuthProviderTypeAnonymous}},
+		},
+	}
+
+	for _, tc := range []struct {
+		description   string
+		enableUserErr error
+	}{
+		{description: "should enable a user when a user id is provided"},
```
makesitgo, line 51: similar in-line clean up here

internal/commands/user/enable_test.go
```diff
@@ -0,0 +1,274 @@
...
+		{
+			ID:         "user-5",
+			Identities: []realm.UserIdentity{{ProviderType: realm.AuthProviderTypeCustomToken}},
+			Type:       "type-3",
+			Disabled:   true,
+		},
+	}
+	for _, tc := range []struct {
+		description     string
+		outputs         []userOutput
```
makesitgo, line 180: [meganit] use the new type here: `userOutputs` and below

internal/commands/user/list_test.go
```diff
@@ -103,158 +99,82 @@ func TestUserListHandler(t *testing.T) {
...
-	}
-}
-
 func TestUserListFeedback(t *testing.T) {
 	for _, tc := range []struct {
-		description    string
-		users          []realm.User
-		expectedOutput string
+		description     string
+		outputs         []userOutput
```
makesitgo, line 115: [meganit] use the new type: `userOutputs` here

internal/commands/user/list_test.go
```diff
@@ -103,158 +99,82 @@ func TestUserListHandler(t *testing.T) {
...
+			outputs: []userOutput{
 				{
-					ID:                     "id1",
-					Identities:             []realm.UserIdentity{{ProviderType: realm.AuthProviderTypeUserPassword}},
-					Type:                   "type1",
-					Disabled:               false,
-					Data:                   map[string]interface{}{"email": "myEmail1"},
-					CreationDate:           1111111111,
-					LastAuthenticationDate: 1111111111,
+					user: realm.User{
```
makesitgo, line 126: save a line and move this inline with `outputs: userOutputs{user: realm.User{` (consolidate the end of this declaration to so it ends with `}},`

internal/commands/user/output.go
```diff
@@ -16,6 +16,18 @@ const (
 	headerDetails                = "Details"
 )
 
+type userOutputs []userOutput
+
+func (uo userOutputs) outputsByProviderType() map[realm.AuthProviderType]userOutputs {
```
makesitgo, line 21: [nit] rename this to `mapByProviderType`

internal/commands/user/output.go
```diff
@@ -16,6 +16,18 @@ const (
...
+	var outputsByProviderType = map[realm.AuthProviderType]userOutputs{}
+	for _, output := range uo {
```
makesitgo, line 23: can we make this read `for _, output := range outputs {`?  and change `outputsByProviderType` to just `out` or `outputsM` or something?

internal/commands/user/output.go
```diff
@@ -26,3 +38,44 @@ func getUserOutputComparerBySuccess(outputs []userOutput) func(i, j int) bool {
...
+		headers = append(headers, headerName)
+	case realm.AuthProviderTypeUserPassword:
+		headers = append(headers, headerEmail)
+	}
+	headers = append(
+		headers,
+		headerID,
+		headerType,
+	)
+	return headers
```
makesitgo, line 57: [nit] we can probably just use `return append(headers, headerID, headerType)` all in one line here

internal/commands/user/output.go
```diff
@@ -26,3 +38,44 @@ func getUserOutputComparerBySuccess(outputs []userOutput) func(i, j int) bool {
...
+}
+
+func userTableRows(authProviderType realm.AuthProviderType, outputs []userOutput, tableRowModifier userTableRowModifier) []map[string]interface{} {
```
makesitgo, line 60: use the new type here: `userOutputs`

makesitgo (commented):
submitting these comments ahead of our chat... should be good to go after

internal/commands/user/disable.go
nathantfrank: discussed offline

internal/cloud/realm/user_ext_test.go
```diff
@@ -99,17 +100,22 @@ func TestRealmUsers(t *testing.T) {
...
 				assert.Nil(t, client.DisableUser(groupID, testApp.ID, email1.ID))
-				assert.Nil(t, client.DisableUser(groupID, testApp.ID, email3.ID))
-			})
-
-			t.Run("And find all disabled users", func(t *testing.T) {
-				users, err := client.FindUsers(groupID, testApp.ID, realm.UserFilter{State: realm.UserStateDisabled})
-				assert.Nil(t, err)
+				users1, users1Err := client.FindUsers(groupID, testApp.ID, realm.UserFilter{IDs: []string{email1.ID}})
+				assert.Nil(t, users1Err)
+				email1Disabled := users1[0]
```
makesitgo, line 105: you should put an `assert.Equal(t, 1, len(users1))` here, otherwise the test may just fail to a panic (do this for the ones below too)

internal/cloud/realm/user_ext_test.go
```diff
@@ -99,17 +100,22 @@ func TestRealmUsers(t *testing.T) {
...
+				assert.True(t, email1Disabled.Disabled, fmt.Sprintf("expected %s to be disabled", email1Disabled.Data["email"]))
```
makesitgo, line 106: [optional] this test data doesn't live around too long, so the id values aren't too too helpful here.  feel free to leave as is, but having "expected user to be disabled" could suffice

internal/cloud/realm/user_ext_test.go
```diff
@@ -120,7 +126,53 @@ func TestRealmUsers(t *testing.T) {
 				}
 				users, err := client.FindUsers(groupID, testApp.ID, filter)
 				assert.Nil(t, err)
-				assert.Equal(t, []realm.User{email3}, users)
+				assert.Equal(t, 1, len(users))
+				assert.Equal(t, email3.ID, users[0].ID)
```
makesitgo, line 130: [question] if you nested this test in the one just above here, would you again have access to `email3Disabled`, and could that work here for an assert like `assert.Equal(t, []realm.User{email3Disabled}, users)`?  if not, this is totally fine to leave as is

internal/cloud/realm/user_ext_test.go
```diff
@@ -120,7 +126,53 @@ func TestRealmUsers(t *testing.T) {
...
+						}
+					}
+					assert.Equal(t, []realm.User{email1Enabled, email2, email3Enabled}, emailUsers)
+					assert.Equal(t, []string{apiKey1.ID, apiKey2.ID}, apiKeyIDs)
+				})
+			})
+
+			t.Run("And find enabled user/password users", func(t *testing.T) {
+				filter := realm.UserFilter{State: realm.UserStateEnabled, Providers: []realm.AuthProviderType{realm.AuthProviderTypeUserPassword}, IDs: []string{email1.ID, email2.ID, email3.ID}}
+				users, err := client.FindUsers(groupID, testApp.ID, filter)
```
makesitgo, line 171: these users are probably back equal to their original states now (after being disabled and enabled)?  would asserting equality on `[]realm.User{email1, email2, email3}` suffice here?

internal/commands/user/delete_test.go
```diff
@@ -9,16 +9,164 @@ import (
...
+					c.ExpectString("Which user(s) would you like to delete?")
+					c.Send("user-2")
+					c.SendLine(" ")
+					c.ExpectEOF()
+				},
+				users:         testUsers[1:],
+				expectedUsers: []realm.User{testUsers[1]},
+			},
+			{
+				description: "with status set",
```
makesitgo, line 80: [qeustion] were we going to consolidate some of these test cases here?

makesitgo (approved):
last few questions/comment from me, but then once the commits and evergreen are in line, this is good to go!

seriously, nice stuff with this.  thank you for dealing with me 🙃 

🆒  beans

internal/commands/user/delete_test.go
nathantfrank: You are totally right, I remembered I forgot to do this late last night

internal/cloud/realm/user_ext_test.go
nathantfrank: Yeah, that would work

PR #53, opened 2021-01-27
REALMC-7166: Add pull command

makesitgo:


internal/commands/app/init_inputs.go
```diff
@@ -14,7 +14,7 @@ const (
 	flagProjectUsage = "the MongoDB cloud project id"
 
 	flagFrom      = "from"
-	flagFromShort = "s"
+	flagFromShort = "a"
```
kpatel71716, line 17: [minor thought] I am wondering if, once the commands are implemented, we want to take a look at all of the flags and their shorthands to see if anything could be more intuitive here

kpatel71716 (approved):
LGTM! Nice stuff here, don't have anything besides that one comment

internal/commands/app/init_inputs.go
nathantfrank: On that same thought, with a flag as short as `--from` it might not necessarily warrant a shorthand, unless it's the most common used flag (in which case it probably should get prioritized to get the `-f`). This thought goes towards our flags that are global or across multiple commands, unless they are to be used often removing their shorthand might make more sense if there is flag that is going to be used more even if it is local to a single command.

internal/commands/pull/command_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+	cmd := &Command{}
+	assert.Nil(t, cmd.realmClient)
+
+	assert.Nil(t, cmd.Setup(profile, nil))
+	assert.NotNil(t, cmd.realmClient)
+}
+
+func TestPullHandler(t *testing.T) {
+	t.Run("Should return an error if the command fails to resolve from", func(t *testing.T) {
```
nathantfrank, line 29: [nit] keep test descriptions lowercase for ease of use when running a test

internal/commands/pull/inputs.go
```diff
@@ -0,0 +1,118 @@
...
+const (
+	flagFrom      = "from"
+	flagFromShort = "a"
+	flagFromUsage = "specify the app to pull changes down from"
+
+	flagProject      = "project"
+	flagProjectUsage = "the MongoDB cloud project id"
+
+	flagAppVersion      = "app-version"
+	flagAppVersionUsage = "specity the app config version to pull changes down as"
```
nathantfrank, line 23: [minor] typo, 
```suggestion
	flagAppVersionUsage = "specify the app config version to pull changes down as"
```

internal/commands/pull/inputs.go
```diff
@@ -0,0 +1,118 @@
...
+	flagIncludeDependenciesShort = "d"
+	flagIncludeDependenciesUsage = "specify the command to push Realm app dependencies changes as well"
+
+	flagIncludeHosting      = "include-hosting"
+	flagIncludeHostingShort = "s"
+	flagIncludeHostingUsage = "specify the command to push Realm app hosting changes as well"
+
+	flagDryRun      = "dry-run"
+	flagDryRunShort = "x"
+	flagDryRunUsage = "specify the command to run without writing any changes to the file system"
```
nathantfrank, line 39: [q] These flags all start with `specify the command to` even though they are bools. Is there something missing or something I don't understand.

internal/commands/pull/inputs_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	u "github.com/10gen/realm-cli/internal/utils/test"
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/10gen/realm-cli/internal/utils/test/mock"
+
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+func TestPullInputsResolve(t *testing.T) {
+	t.Run("Should return an error if run from outside a project directory and no app-dir is set", func(t *testing.T) {
```
nathantfrank, line 18: [nit] keep test descriptions lowercase for ease of use when running a test

nathantfrank (changes requested):
LGTM, I think you are almost there, just had a question and a thought to add

internal/commands/app/init_inputs.go
makesitgo: yeah, in general polish is needed across the board for command flags and usages.  for now, we'll stick with whatever works, and get product/design to chime in on the final say once we've got more of the command suite built out (and we can really see what our common options are, etc)

internal/commands/pull/inputs.go
makesitgo: you're not missing anything, im just bad with words.  i won't worry too much about the particular language in the flags' usages just yet, but the intention here was that a `bool` flag is used to "specify the command to" run/push/do stuff

nathantfrank: KO, just to have my thoughts put down, I think the language around bool flags might be better suited to be `include to <behavior>`

nathantfrank (approved):
LGTM

internal/commands/pull/inputs.go
makesitgo: i like that, updated the `bool` flags for `push` and `pull` here

PR #54, opened 2021-01-28
REALMC-7160 build out user revoke command

kpatel71716:


internal/commands/user/flags.go
```diff
@@ -19,4 +19,5 @@ const (
 	flagUserShort       = "u"
 	flagUserUsage       = `set the user ids for which to filter the list of app users with`
 	flagUserDeleteUsage = `set the user ids for which to delete from the app`
+	flagUserRevokeUsage = `set the user ids for which to revoke from the app`
```
kpatel71716, line 22: [q] is there text from design for what should go here. In the [spec](https://docs.google.com/document/d/15Y0_jCxU2OOVx5sspQZjdewaUbFtkvWsxZUDrnkk0vc/edit#heading=h.cyglq0hem105) I saw `Select (many) from list of (filtered) app users` in the user flag prompt but I didn't see similar text in other places so wasn't sure

internal/commands/user/inputs.go
```diff
@@ -93,7 +93,7 @@ func (i *usersInputs) ResolveUsers(ui terminal.UI, client realm.Client, app real
 	askErr := ui.AskOne(
 		&selectedUsers,
 		&survey.MultiSelect{
-			Message: "Which user(s) would you like to delete?",
+			Message: "Which user(s) would you like to select?",
```
kpatel71716, line 96: made this text more generic since revoke reuses the same function

internal/commands/user/revoke.go
```diff
@@ -0,0 +1,151 @@
...
+
+func userRevokeTableRow(authProviderType realm.AuthProviderType, output userOutput) map[string]interface{} {
+	var details string
+	if output.err != nil {
+		details = output.err.Error()
+	}
+	success := "no"
+	if output.err == nil {
+		success = "yes"
+	}
```
kpatel71716, line 137: Thoughts on outputting yes/no instead of true/false in the cli feedback? I felt it would make for sense to the user
Screen Shot 2021-01-27 at 10 37 54 PM

internal/commands/user/revoke.go
```diff
@@ -0,0 +1,151 @@
...
+		headerDetails: details,
+	}
+	switch authProviderType {
+	case realm.AuthProviderTypeAPIKey:
+		row[headerName] = output.user.Data[userDataName]
+	case realm.AuthProviderTypeUserPassword:
+		row[headerEmail] = output.user.Data[userDataEmail]
+	}
+	return row
+}
```
kpatel71716, line 151: I pretty much just took this from `user delete`. I don't mind refactoring and reusing here though I actually don't mind keeping the code in the commands separate in this case. Any strong preferences for reusing?

internal/commands/user/revoke_test.go
```diff
@@ -0,0 +1,337 @@
...
+		},
+	}
+	for _, tc := range []struct {
+		description    string
+		outputs        []userOutput
+		expectedOutput string
+	}{
+		{
+			description:    "should show no users to revoke",
+			expectedOutput: "01:23:45 UTC INFO  No users to revoke, try changing the --user input\n",
```
kpatel71716, line 200: I took this text from the spec as well

internal/commands/user/inputs.go
```diff
@@ -71,10 +71,10 @@ func (i *usersInputs) ResolveUsers(ui terminal.UI, client realm.Client, app real
...
 		return nil, usersErr
 	}
-	if len(i.Users) > 0 {
-		if len(foundUsers) == 0 {
-			return nil, errors.New("no users found")
-		}
+	if len(i.Users) > 0 && len(foundUsers) == 0 {
+		return nil, errors.New("no users found")
+	}
+	if len(i.Users) > 0 || len(foundUsers) == 0 {
```
kpatel71716, line 77: I modified the logic here to prevent an unexpected error case when an app has no users. I tried some different refactoring but this one worked the best for both manual and unit tests

internal/commands/user/revoke.go
```diff
@@ -0,0 +1,151 @@
...
+		err := cmd.realmClient.RevokeUserSessions(app.GroupID, app.ID, user.ID)
+		cmd.outputs = append(cmd.outputs, userOutput{user: user, err: err})
+	}
+	return nil
+}
+
+// Feedback is the command feedback
+func (cmd *CommandRevoke) Feedback(profile *cli.Profile, ui terminal.UI) error {
+	if len(cmd.outputs) == 0 {
+		return ui.Print(terminal.NewTextLog("No users to revoke, try changing the --user input"))
```
kpatel71716, line 79: I came across an interesting scenario here from my own user error (though I could see another dev doing the same). When selecting the users with a multi-select, I tend to press enter instead of space and accidentally select no users. I modified this message here as sort of an indicator that maybe the user did not select a user. We use text like `No users to revoke` elsewhere but I wanted to get some thoughts on the messaging here since the shorter text could not be registered as a user did not select any users and just pressed enter.

kpatel71716:
I wanted to get some thoughts on the comments I dropped in the PR before fixing the broken tests. Feel free to reach out to discuss anything here further

internal/commands/user/revoke.go
nathantfrank: I personally prefer it after seeing it, but I don't know if there are any norms we have for displaying information

internal/commands/user/revoke.go
nathantfrank: I think we should definitely start to refactor these feedback helper functions since there is just tons of lines that are identical.

internal/commands/user/inputs.go
nathantfrank: I'm debating whether this should be passed in as an arg, or keeping it generic. Thoughts?

internal/commands/user/revoke.go
nathantfrank: I like the change, I think we can catch 0 users selected when in interactive mode and print `No users were selected to revoke` and end the command there

nathantfrank (commented):
Just had a few thoughts I want to keep discussing

internal/commands/user/revoke.go
kpatel71716: That's a good idea. I will take a look around and think about this some more. The one tricky part is that there is a difference between no users found and no users selected and only the common inputs.go code would be able to tell the difference. I know we previously discussed having each command have it's own logic on what to do with the found users (which I still think is a good approach). But maybe the `ResolveUsers` function should be split between finding users and selecting users? That would let the individual commands dictate the flow

internal/commands/user/revoke.go
kpatel71716: I was thinking the same, I will let @makesitgo chime in here as well

internal/commands/user/inputs.go
kpatel71716: I think it would definitely be 'best' to be more descriptive. In terms or code maintainability and simplicity I like just using 'select' here personally. Though if anyone with a strong preference wants an arg for the message, I am not opposed to it

internal/commands/user/revoke.go
kpatel71716: Yeah, the super simple way would be to change `headerRevoked` to something generic like `Successful` and then we could reuse the code across commands  pretty easily. Not sure if we have a standard on this or not.

internal/commands/user/flags.go
makesitgo: flag text is not quite formalized yet... just go with mostly what you have here, just update the end to be `"...for which to revoke sessions from"`

internal/commands/user/inputs.go
makesitgo: we can be descriptive here... let's add an arg to `ResolveUsers` here in the form of `action string` (place it up front, before `ui`) and rename the function to `reoslveUsersFor` (is there still a reason this needs to be exported?)

that way, its usage ends up reading like:

```go
users, usersErr := cmd.inputs.resolveUsersFor("delete", ui, cmd.realmClient, app)
```

internal/commands/user/revoke.go
makesitgo: I like the look and feel of this as well... do you think this might be something we want for _all_ booleans?  im inclined to think yes, and if so, we could probably make this change in `terminal/table.go`'s `parseValue` function.

what do you all think?

internal/commands/user/revoke.go
makesitgo: my thinking here is that a lot of this code isn't necessarily a bad thing ... it's explicit about what it does for each command.  that being said, we can definitely spruce it up a bit and reduce a bit of the repetition here.

i think its about time we start with some kind of `TableBuilder` concept, to hopefully alleviate some of the pains and awkward code here.  It's could just construct the `[]string` and `[]map[string]interface{}` pairs for printing tables.  We can chat about this offline in more detail

internal/commands/user/revoke.go
kpatel71716: I tested out the change in table.go in this PR (seemed to only affect the user commands). I liked the result

internal/commands/user/revoke.go
kpatel71716: I decided to stick with the message in feedback since I wasn't sure if we wanted to display an error or just a log like this if a user selects no users. Wanted to get some more thoughts here, but if we stick with what is in the PR, I can add the `try changing the -user input` text to the other commands

internal/commands/user/delete.go
```diff
@@ -60,11 +55,18 @@ func (cmd *CommandDelete) Handler(profile *cli.Profile, ui terminal.UI) error {
...
-		return usersErr
+
+	resolvedUsers, resolveErr := cmd.inputs.resolveUsers(cmd.realmClient, app.GroupID, app.ID)
+	if resolveErr != nil {
+		return resolveErr
+	}
+
+	selectedUsers, selectErr := cmd.inputs.selectUsers(ui, resolvedUsers, "delete")
+	if selectErr != nil {
+		return selectErr
```
nathantfrank, line 66: [nit] keep the prefix consistent
`resolvedUsers, resolveErr` -> `resolvedUsers, resolvedErr`
`selectedUsers, selectErr ` -> `selectedUsers, selectedErr`

internal/commands/user/disable.go
```diff
@@ -47,12 +47,17 @@ func (cmd *CommandDisable) Handler(profile *cli.Profile, ui terminal.UI) error {
 		return appErr
 	}
 
-	users, usersErr := cmd.realmClient.FindUsers(app.GroupID, app.ID, realm.UserFilter{IDs: cmd.inputs.Users})
-	if usersErr != nil {
-		return usersErr
+	resolvedUsers, resolveErr := cmd.inputs.resolveUsers(cmd.realmClient, app.GroupID, app.ID)
```
nathantfrank, line 50: [nit] consistent prefix

internal/commands/user/enable.go
```diff
@@ -47,12 +47,17 @@ func (cmd *CommandEnable) Handler(profile *cli.Profile, ui terminal.UI) error {
 		return appErr
 	}
 
-	users, usersErr := cmd.realmClient.FindUsers(app.GroupID, app.ID, realm.UserFilter{IDs: cmd.inputs.Users})
-	if usersErr != nil {
-		return usersErr
+	resolvedUsers, resolveErr := cmd.inputs.resolveUsers(cmd.realmClient, app.GroupID, app.ID)
```
nathantfrank, line 50: [nit] consistent prefix

internal/commands/user/enable.go
```diff
@@ -47,12 +47,17 @@ func (cmd *CommandEnable) Handler(profile *cli.Profile, ui terminal.UI) error {
...
+	if resolveErr != nil {
+		return resolveErr
 	}
 
-	for _, user := range users {
+	selectedUsers, selectErr := cmd.inputs.selectUsers(ui, resolvedUsers, "enable")
```
nathantfrank, line 55: [nit] consistent prefix

internal/commands/user/inputs.go
```diff
@@ -47,3 +57,56 @@ func displayUserData(apt realm.AuthProviderType, user realm.User) string {
...
+	return realm.UserFilter{
+		IDs:       i.Users,
+		State:     i.State,
+		Pending:   i.Pending,
+		Providers: realm.NewAuthProviderTypes(i.ProviderTypes...),
+	}
+}
+
+func (i multiUserInputs) resolveUsers(realmClient realm.Client, groupID, appID string) ([]realm.User, error) {
+	foundUsers, findErr := realmClient.FindUsers(groupID, appID, i.filter())
```
nathantfrank, line 71: [nit] consistent prefix

internal/terminal/table.go
```diff
@@ -131,6 +131,12 @@ func parseValue(value interface{}) string {
 	case nil: // leave zero-value
 	case string:
 		parsed = v
+	case bool:
+		if value == true {
```
nathantfrank, line 135: [Q] do we need to do the equate? Could it just be 
`if value {`

internal/commands/user/revoke.go
nathantfrank: My preference is leaning towards the display an error, but I also don't have a super strong opinion if @makesitgo feels otherwise.

nathantfrank (commented):
GJ, just a few nits and a Q

internal/commands/user/disable.go
makesitgo: [optional] to add to this, maybe just stick to `resolved`/`selected` for the names of the user arrays?

internal/commands/user/input_test.go
```diff
@@ -1,11 +1,14 @@
...
 
 import (
+	"errors"
 	"fmt"
 	"testing"
 
 	"github.com/10gen/realm-cli/internal/cloud/realm"
 	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/10gen/realm-cli/internal/utils/test/mock"
+	"github.com/Netflix/go-expect"
```
makesitgo, line 11: newline here

internal/commands/user/input_test.go
```diff
@@ -96,3 +99,191 @@ func TestProviderTypeDisplayUser(t *testing.T) {
...
+			users         []realm.User
+			expectedUsers []realm.User
+		}{
+			{
+				description:   "with no input set",
+				users:         testUsers,
+				expectedUsers: testUsers,
+			},
+			{
+				description:   "with all input set, except users",
```
makesitgo, line 156: avoid punctuation in these descriptions, just go with `"with all input set except users"` here

---

also, i'm not sure what this test adds more than the one previous to this does?  for this `resolveUsers` function, id expect to see test cases for:

  - [x] when client errors, we error
  - [ ] when client returns no users, and inputs specified no user ids, we return an empty array
  - [x] when client returns no users, and inputs specified user ids, we error
  - [x] when a client returns users, we return those users

all other ones here should be able to be omitted, and you are probably left here with a test table of size 3 and one extra/error test.

internal/commands/user/input_test.go
```diff
@@ -96,3 +99,191 @@ func TestProviderTypeDisplayUser(t *testing.T) {
...
+			Data:       map[string]interface{}{"email": "user-3@test.com"},
+		},
+	}
+
+	t.Run("should prompt for users", func(t *testing.T) {
+		for _, tc := range []struct {
+			description   string
+			inputs        multiUserInputs
+			procedure     func(c *expect.Console)
+			users         []realm.User
```
makesitgo, line 242: it doesn't seem like `tc.users` is a necessary test case parameter here, right?

internal/commands/user/inputs.go
```diff
@@ -47,3 +57,56 @@ func displayUserData(apt realm.AuthProviderType, user realm.User) string {
...
+		return nil, findErr
+	}
+	if len(i.Users) > 0 && len(foundUsers) == 0 {
+		return nil, errors.New("no users found")
+	}
+	return foundUsers, nil
+}
+
+func (i multiUserInputs) selectUsers(ui terminal.UI, resolvedUsers []realm.User, action string) ([]realm.User, error) {
+	if len(i.Users) > 0 {
```
makesitgo, line 82: looking at this now, we could probably turn this if check into `if len(resolvedUsers) == 0` and just return `nil, nil` if thats the case.  in that case, we don't need this function to be defined on `multiUserInputs` either, it could just be:

```go
func selectUsers(ui terminal.UI, users []realm.User, action string) ([]realm.User, error) {
  if len(users) == 0 { return nil, nil }
  
  selected := make([]realm.User, len(selections))
  // do rest of select logic here...
  return select, nil
}
```

internal/commands/user/input_test.go
```diff
@@ -96,3 +99,191 @@ func TestProviderTypeDisplayUser(t *testing.T) {
...
+					c.ExpectString("Which user(s) would you like to delete?")
+					c.Send("user-1")
+					c.SendLine(" ")
+					c.ExpectEOF()
+				},
+				users:         testUsers,
+				expectedUsers: []realm.User{testUsers[0]},
+			},
+			{
+				description: "with all input set, except users",
```
makesitgo, line 257: does this test really do much different than the one before it?  I think what we need a test case here for is when `selectUsers` returns early and never prompts the user for input

internal/commands/user/output.go
```diff
@@ -14,6 +14,7 @@ const (
 	headerType                   = "Type"
 	headerDeleted                = "Deleted"
 	headerDetails                = "Details"
+	headerRevoked                = "Revoked"
```
makesitgo, line 17: let's make this `"Session Revoked"`

internal/commands/user/revoke.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/10gen/realm-cli/internal/utils/flags"
+	"github.com/spf13/pflag"
```
makesitgo, line 11: newline here

internal/commands/user/revoke.go
makesitgo: i think ideally we aren't ever printing any `ERROR` messages in command code (and let those be entirely system driven).  in this case, i think just printing `No users to revoke sessions for` (or `No users found...`) is just fine

internal/terminal/table.go
makesitgo: `value` is still seen as an `interface{}`, but you should be able to use the reference to `v` for that:

```go
if v { parsed = "yes" } else { parsed = "no" }

**Review by makesitgo — COMMENTED** [review]
here's a first pass from me, i like how that `multiUserInputs` refactor looks

File: internal/commands/user/inputs.go (hunk 12 above)
Comment (kpatel71716) [reply internal/commands/user/inputs.go]: changed to `foundUsers, err` since I liked the tense of `found` here

File: internal/commands/user/input_test.go (hunk 16 above)
Comment (kpatel71716) [reply internal/commands/user/input_test.go:246]: With the latest changes it is necessary (i.e. one test case needs that param to be nil while the other needs it to have a full slice of users that gets filtered out and checked against expectedUsers), so I left it in.

File: internal/commands/user/input_test.go (hunk 15 above)
Comment (kpatel71716) [reply internal/commands/user/input_test.go]: I cleaned up the test cases a bit. I _think_ the test cases present now account for the right subset of scenarios. I can add one for testing an error when selecting users. Looking at how to trigger that rn

File: internal/commands/user/input_test.go [hunk 21]
```diff
@@ -96,3 +100,201 @@ func TestProviderTypeDisplayUser(t *testing.T) {
...
+					c.ExpectEOF()
+				},
+				users:         testUsers,
+				expectedUsers: []realm.User{testUsers[0]},
+			},
+			{
+				description: "unless no users are found",
+				procedure: func(c *expect.Console) {
+					console, _ := c.ExpectEOF()
+					assert.Equal(t, "", console)
```
kpatel71716, line 264: added this check here to ensure that we aren't prompting the user (though it did not come across that pattern anywhere else). Asserting nil on the error failed in evergreen from a random error, so I just left this as is: `*os.PathError{read /dev/ptmx: input/output error}`

internal/commands/user/inputs.go
kpatel71716: The issue came across here was a scenario where the user flag is passed in, and those users are found before this function calls. If we don't user the multiUserInputs.Users field, we will prompt the user for selection even though that step isn't necessary. I updated the logic to check both the multiUserInputs.Users and users param so we can account for all scenarios

internal/commands/user/inputs.go
```diff
@@ -47,3 +57,56 @@ func displayUserData(apt realm.AuthProviderType, user realm.User) string {
...
+func (i multiUserInputs) filter() realm.UserFilter {
+	return realm.UserFilter{
+		IDs:       i.Users,
+		State:     i.State,
+		Pending:   i.Pending,
+		Providers: realm.NewAuthProviderTypes(i.ProviderTypes...),
+	}
+}
+
+func (i multiUserInputs) resolveUsers(realmClient realm.Client, groupID, appID string) ([]realm.User, error) {
```
makesitgo, line 70: [nit] let's rename this to `findUsers`

internal/commands/user/disable.go
```diff
@@ -47,12 +47,16 @@ func (cmd *CommandDisable) Handler(profile *cli.Profile, ui terminal.UI) error {
 		return appErr
 	}
 
-	users, usersErr := cmd.realmClient.FindUsers(app.GroupID, app.ID, realm.UserFilter{IDs: cmd.inputs.Users})
-	if usersErr != nil {
-		return usersErr
+	resolved, resolvedErr := cmd.inputs.resolveUsers(cmd.realmClient, app.GroupID, app.ID)
```
makesitgo, line 50: now that this original function will be renamed to `findUsers`, each one of these `disableInputs`/`revokeInputs`/etc structs should then implement their own:

```go
func (i disableInputs) resolveUsers(ui terminal.UI, realmClient realm.Client, app realm.App) ([]realm.User, error) {
  found, foundErr := i.findUsers(realmClient, app)
  // ...
  
  return i.selectUsers(ui, users, "disable")
}
```

with this approach, you can go back to calling these `users, usersErr` in the command code where you call `cmd.inputs.resolveUsers(ui, cmd.realmClient, app)`

makesitgo (commented):
one last tweak here... and then we'll get the test cases in line from there.  thanks for this!

internal/commands/user/inputs.go
kpatel71716: Discussed offline, leaving as is

internal/terminal/table.go
```diff
@@ -131,6 +131,12 @@ func parseValue(value interface{}) string {
 	case nil: // leave zero-value
 	case string:
 		parsed = v
+	case bool:
+		if v {
+			parsed = "yes"
+		} else {
+			parsed = "no"
+		}
```
tsedgwick, line 139: was this a product decision? seems like true/false is pretty standard among developers.

internal/commands/user/revoke.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// CommandRevoke is the `user revoke` command
+type CommandRevoke struct {
+	inputs      revokeInputs
+	outputs     userOutputs
+	realmClient realm.Client
+}
+
+type revokeInputs struct {
```
tsedgwick, line 22: [nit] should this just follow the same pattern as the command and be `inputsRevoke`?

internal/commands/user/delete.go
```diff
@@ -60,10 +55,12 @@ func (cmd *CommandDelete) Handler(profile *cli.Profile, ui terminal.UI) error {
 	if appErr != nil {
 		return appErr
 	}
-	users, usersErr := cmd.inputs.resolveUsers(ui, cmd.realmClient, app)
-	if usersErr != nil {
-		return usersErr
+
+	users, userErr := cmd.inputs.resolveUsers(ui, cmd.realmClient, app)
```
makesitgo, line 59: [meganit] i think this diff goes away if you just rename this back to `usersErr`...

internal/commands/user/disable.go
```diff
@@ -47,9 +47,9 @@ func (cmd *CommandDisable) Handler(profile *cli.Profile, ui terminal.UI) error {
 		return appErr
 	}
 
-	users, usersErr := cmd.realmClient.FindUsers(app.GroupID, app.ID, realm.UserFilter{IDs: cmd.inputs.Users})
-	if usersErr != nil {
-		return usersErr
+	users, userErr := cmd.inputs.resolveUsers(ui, cmd.realmClient, app)
```
makesitgo, line 50: same thing here with the unnecessary diff... probably copy/pasta across these modified commands if you want to revert those back

internal/commands/user/input_test.go
```diff
@@ -96,3 +100,174 @@ func TestProviderTypeDisplayUser(t *testing.T) {
...
+		}{
+			{
+				description:   "should return found all users with no input users ",
+				users:         testUsers,
+				expectedUsers: testUsers,
+			},
+			{
+				description:   "should error with input users and no found users",
+				inputs:        multiUserInputs{Users: []string{"user-1"}},
+				expectedUsers: nil,
```
makesitgo, line 160: [nit] can omit this

internal/commands/user/input_test.go
```diff
@@ -96,3 +100,174 @@ func TestProviderTypeDisplayUser(t *testing.T) {
...
+				expectedErr:   errors.New("no users found"),
+			},
+
+			{
+				description:   "should return an empty users slice with no found users or input users",
+				expectedUsers: nil,
```
makesitgo, line 166: [nit] can omit this

internal/commands/user/input_test.go
```diff
@@ -96,3 +100,174 @@ func TestProviderTypeDisplayUser(t *testing.T) {
...
+			{
+				description: "should prompt with no input set",
+				procedure: func(c *expect.Console) {
+					c.ExpectString("Which user(s) would you like to delete?")
+					c.Send("user-1")
+					c.SendLine(" ")
+					c.ExpectEOF()
+				},
+				users:         testUsers,
+				expectedUsers: []realm.User{testUsers[0]},
```
makesitgo, line 231: [nit] could get away with just `expectedUsers: testUsers[0:1]` here

internal/commands/user/input_test.go
```diff
@@ -96,3 +100,174 @@ func TestProviderTypeDisplayUser(t *testing.T) {
...
+			},
+			{
+				description: "should not prompt if no users are found",
```
makesitgo, line 234: I missed your comment on this before... when have tests that assert the absence of prompts, we don't need to write them in this style where we're using a pseudo-terminal and all.  we probably want to move things around here to be:

```go
t.Run("should prompt with no input set", func(t *testing.T) {
  _, console, _, ui, consoleErr := mock.NewVT10XConsole()
  assert.Nil(t, consoleErr)
  defer console.Close()

  doneCh := make(chan (struct{}))
  go func() {
    defer close(doneCh)
 
    console.ExpectString("Which user(s) would you like to delete?")
    console.Send("user-1")
    console.SendLine(" ")
    console.ExpectEOF()
  }()

  var i multiUserInputs
  users, err := i.selectUsers(ui, testUsers, "delete")

  console.Tty().Close() // flush the writers
  <-doneCh              // wait for procedure to complete

  assert.Nil(t, err)
  assert.Equal(t, testUsers[0:1], users)
})

t.Run("should not prompt the user", func(t *testing.T) {
for _, tc := range []struct{
  description string
  inputs multiUserInputs
  users []realm.User
}{
  {"with no inputs set and no users found"},
  {
    "with user inputs set and some users found",
    multiUserInputs{Users: []string{"user-1"}},
    testUsers,
  },
}{
  t.Run(tc.description, func(t *testing.T) {
    users, err := tc.inputs.selectUsers(nil, tc.users, "")
    assert.Nil(t, err)
    assert.Equal(t, tc.users, users)
  })
}
})
```

internal/commands/user/enable.go
```diff
@@ -96,3 +96,12 @@ func userEnableRow(output userOutput, row map[string]interface{}) {
...
 	row[headerDetails] = details
 }
+
+func (i enableInputs) resolveUsers(ui terminal.UI, realmClient realm.Client, app realm.App) ([]realm.User, error) {
+	found, foundErr := i.findUsers(realmClient, app.GroupID, app.ID)
+	if foundErr != nil {
+		return nil, foundErr
+	}
+
+	return i.selectUsers(ui, found, "enable")
```
tsedgwick, line 106: seems like this is repeated for all actions, if we're going to abstract out `selectUsers` and `findUsers`, why not just remove this `resolveUsers` func and call `selectUsers`?

makesitgo (approved):
just one last test re-organization, but this is all good by me!  nice work

🆒  beans

internal/terminal/table.go
makesitgo: I just threw a dart here... I think initially it was showing as `TRUE`/`FALSE` and became a bit visually jarring, but we can for sure roll with `"true"`/`"false"` here if you think that's more natural  (i'll take an informal poll with product/design as well)

internal/commands/user/delete.go
tsedgwick: [meganit] idiomatic go prefers `errXXX`, or simply, `err` 😎

internal/terminal/table.go
tsedgwick: I believe it's more natural and we shouldn't muddy the waters of changing a boolean to a yes/no personally.

internal/commands/user/enable.go
makesitgo: blame me for that one!  feel free to go with tim's route here

internal/commands/user/revoke.go
makesitgo: we discussed this one a while back together... we kept inputs in the style of `xxxInputs` to match the corresponding file `xxx_inputs.go` (ensuring inputs and command code sat next to each other in the package)

internal/commands/user/delete.go
kpatel71716: I went with err for all 3 calls in the Handlers of the commands (and I even refactored user list a bit to line up), lmk of any nits on the naming here :)

makesitgo (approved):
:cool:  beans (again)

PR #55, opened 2021-01-29
REALMC-7865: Support config versions when reading/writing local app data

makesitgo:
Purpose of this refactor:
  - eliminate `app` package
    - `app.Config` is now moved to `local.App`
    - `app.Reolve`/`app.ResolveGroupID`/`app.ProjectInputs` all now live in `cli` package
    - take back `app` var name in code
  - tackle this config version thing head on: introducing the `local` package
    - `local.App` is constructed one of a few ways:
      1. with a rootDir, name, location, and deploymentModel (as a default, `ConfigVersion20210101`* app)
      1. loaded from a rootDir, where it's searched upwards for the project root and identified `realm_config.json` first, if present, then `config.json`, then `stitch.json`
      1. loaded from a rootDir with its respective config parsed and data set
      1. loaded from a rootDir with its entire data loaded from disk
    - `local.App` marshals JSON for two reasons:
      1. to write its config contents to disk in a file at the root directory
      1. to send as the request body for import requests

**NOTE:** `ConfigVersion20210101` is now broken with respect to `baas`/`master` (we are hitting the v1 import route handler with a request it doesn't understand).  that is why `realm.DefaultAppConfigVersion` remains pinned at `ConfigVersion20200603` for now.  [REALMC-7880](https://jira.mongodb.org/browse/REALMC-7880) will remedy that situation from the server-side first, at which point we can then begin [REALMC-7653](https://jira.mongodb.org/browse/REALMC-7653)


internal/local/data.go
```diff
@@ -0,0 +1,176 @@
...
+	if readErr != nil {
+		return fmt.Errorf("failed to read file at %s: %s", opts.path, readErr)
+	}
+
+	if len(data) == 0 {
+		if opts.failOnEmpty {
+			return fmt.Errorf("no file contents at %s", opts.path)
+		}
+		return nil
+	}
```
tsedgwick, line 174: what if we just make a helper func for this section? seems like all we need `out` for is to pass it directly to `json.Unmarshal`. I would suggest we create a helper function that returns `[]byte, error` from the file options. 

```
data, err := fileHelperBetterNamed(opts)
if err != nil { // handle }
err = json.Unmarshal(data, out)
if err != nil { // handle }
```

Maybe go one step further - we could just make the helper a default:
```
data, err := defaultFileHelperBetterNamed(path)
if err != nil { // handle }
err = json.Unmarshal(data, out)
if err != nil { // handle }
```

internal/local/structure_v1.go
```diff
@@ -0,0 +1,237 @@
...
+	return nil
+}
+
+func (a *AppDataV1) unmarshalServices(rootDir string) error {
+	dir := filepath.Join(rootDir, NameServices)
+
+	dw := directoryWalker{path: dir, onlyDirs: true}
+	if walkErr := dw.walk(func(file os.FileInfo, path string) error {
+		var out serviceStructure
+
```
tsedgwick, line 158: [mega-nit] extra line

internal/local/structure_v2.go
```diff
@@ -0,0 +1,236 @@
...
+	return nil
+}
+
+func (a *AppDataV2) unmarshalSecrets(rootDir string) error {
+	opts := defaultFileOptions(rootDir, FileSecrets)
+	if _, statErr := os.Stat(opts.path); statErr == nil {
+		if err := unmarshalJSONInto(&a.Secrets, opts); err != nil {
+			return err
+		}
+	}
```
tsedgwick, line 75: [similar-mega-nit] flip conditional for happy path on left

tsedgwick (approved):
mostly nits / suggestions from my end.

internal/local/structure_v1.go
```diff
@@ -0,0 +1,237 @@
...
+func (a *AppDataV1) unmarshalFunctions(rootDir string) error {
+	functions, err := unmarshalFunctionsV1(filepath.Join(rootDir, NameFunctions))
+	if err != nil {
+		return err
+	}
+	a.Functions = functions
+	return nil
+}
+
+func unmarshalFunctionsV1(path string) ([]map[string]interface{}, error) {
```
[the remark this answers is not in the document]
makesitgo: so this isn't an "app version" per the server standards, but basically it represents the fork in logic pre- and post- Application Structure project.

In the past, even with new config versions we had the some config file structure (just different names).  Now we have a new name and config structure, with new ways to process the file system. So this `V1` and `V2` is meant to reflect that

The lack of real "types" for these structures is that simply, we don't need em (this code path is ultimately only hit when sending an import request, and this builds the request body).  For a lot of this `v1` logic, I really wanted to be as close to the original CLI in implementation as possible.

You'll find in `v2` we start to add a few more structs, granted everything is still more or less a generic map/interface

internal/local/structure_v1.go
```diff
@@ -0,0 +1,237 @@
...
+	opts := defaultFileOptions(rootDir, FileSecrets)
+	if _, statErr := os.Stat(opts.path); statErr == nil {
+		if err := unmarshalJSONInto(&a.Secrets, opts); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (a *AppDataV1) unmarshalAuthProviders(rootDir string) error {
```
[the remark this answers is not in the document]
makesitgo: so the reason I put all of these unmarshal funcs here in the first place was so the `LoadData` could very cleanly outline the loading strategy we use.  none of these are really used elsewhere, it was really just a battle of having `LoadData` be in the order of magnitude of 100  lines of code vs 10 sorta thing.

I went for normal `func`s first and not defined on the pointer receiver, but then the naming got a little hairy b/c `v2` was going to want similarly-named-but-differently-implemented functions.  so here we are with `func (a *AppDataV1)` for everything

what are your thoughts?  just one large function?

internal/local/data.go
```diff
@@ -0,0 +1,176 @@
...
+		case extJSON:
+			var o map[string]interface{}
+			if err := unmarshalJSONInto(&o, fileOptions{path: path}); err != nil {
+				return err
+			}
+			out = append(out, o)
+		}
+		return nil
+	}); err != nil {
+		return nil, nil
```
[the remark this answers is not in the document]
makesitgo: 👻

internal/local/structure_v1.go
makesitgo: fwiw, that was intentional... but i'll tighten it up in here

internal/local/data.go
makesitgo: yeah i like this, went for `unmarshalJSONFile` and `unmarshalJSONFileWithOptions`

internal/cloud/realm/import_export_test.go
```diff
@@ -121,6 +123,74 @@ func TestRealmImportExport(t *testing.T) {
...
+			client := newAuthClient(t)
+
+			groupID := u.CloudGroupID()
+
+			app, appErr := client.CreateApp(groupID, "importexport-test", realm.AppMeta{})
+			assert.Nil(t, appErr)
+
+			appData := tc.importData(app)
+
+			assert.Nil(t, client.Import(groupID, app.ID, appData))
```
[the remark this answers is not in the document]
makesitgo: whats weird here is `gofmt` isnt changing anything, and it looks fine in my editor.  /shrug

internal/local/app.go
```diff
@@ -0,0 +1,163 @@
...
+		for _, config := range allConfigFiles {
+			ok, err := isFileExist(filepath.Join(wd, config.String()))
+			if err != nil {
+				return App{}, false, err
+			}
+			if ok {
+				return App{RootDir: wd, Config: config}, true, nil
+			}
+		}
+		if wd == "/" {
```
tsedgwick, line 156: what is this protecting us from? this won't work on windows for what it's worth.

internal/local/structure_v1.go
```diff
@@ -0,0 +1,276 @@
...
+	ConfigVersion        realm.AppConfigVersion   `json:"config_version"`
+	ID                   string                   `json:"app_id,omitempty"`
+	Name                 string                   `json:"name"`
+	Location             realm.Location           `json:"location"`
+	DeploymentModel      realm.DeploymentModel    `json:"deployment_model"`
+	Security             map[string]interface{}   `json:"security"`
+	Hosting              map[string]interface{}   `json:"hosting,omitempty"`
+	CustomUserDataConfig map[string]interface{}   `json:"custom_user_data_config"`
+	Sync                 map[string]interface{}   `json:"sync"`
+	Environment          interface{}              `json:"environment,omitempty"` // TODO realm.Environment
```
tsedgwick, line 24: should this TODO have a ticket attached?

tsedgwick (approved):
last few nits / q's - overall LGTM

internal/local/app.go
makesitgo: carry-over implementation from the previous `cli` ... seemingly there to try and prevent reading further up than the root directory?  i'm inclined to keep it as is just because we haven't had any issues with it yet, but is there a better way to approach this?

internal/local/structure_v1.go
makesitgo: nope, i had this as a reminder for me to go and figure out what this was in the backend.  we can use just `string` for now

PR #56, opened 2021-01-29
REALMC-7162 build out secrets create command

kpatel71716:


internal/commands/secrets/create_inputs.go
```diff
@@ -0,0 +1,29 @@
...
+			return err
+		}
+	}
+	if i.Value == "" {
+		if err := ui.AskOne(&i.Value, &survey.Input{Message: "Secret Value"}); err != nil {
+			return err
+		}
+	}
+
+	return nil
```
kpatel71716, line 28: Out of curiosity, should we re-prompt if a user passes in an empty string? I don't believe we do elsewhere, and I actually like the idea of us not maintaining logic at this level of code. I just wasn't sure if it would be weird to perform the following steps:
1. Run `secrets create`
2. Press enter when prompted for `Secret Name: `(i.e. passing the empty string)
3. Press enter when prompted for `Secret Value`
4. Select an app if one is not provided
5. `secrets create` fails because the name and value are empty

internal/commands/secrets/flags.go
```diff
@@ -0,0 +1,11 @@
+package secrets
+
+const (
+	flagName      = "secret-name"
+	flagNameShort = "n"
+	flagNameUsage = `the name of the secret to add to your Realm App`
+
+	flagValue      = "secret-value"
+	flagValueShort = "v"
+	flagValueUsage = `the value of the secret to add to your Realm App`
```
kpatel71716, line 10: I know the spec has `-name` and `-value` as the shorthand here but since we can only support one character in the shorthand I used `n` and `v`

internal/cloud/realm/secrets.go
```diff
@@ -37,3 +37,25 @@ func (c *client) Secrets(groupID, appID string) ([]Secret, error) {
 	}
 	return secrets, nil
 }
+
+type createSecretRequest struct {
+	Name  string `json:"name"`
+	Value string `json:"value"`
+}
+
+func (c *client) CreateSecret(groupID, appID, name, value string) error {
```
kpatel71716, line 46: Thoughts on having some sort of e2e test here for secrets? Once we have secrets delete we can probably do something like:
1. Create App
2. Test secrets create
3. Test secrets list
4. test secrets delete
5. Delete app

internal/commands/secrets/create_inputs.go
makesitgo: no we should fail in that scenario... we want a frictionless experience, but not some endless loop chasing certain inputs from the user.

if they are prompted for name and value of a secret and hit enter to leave them blank, we should respond with an error indicating "failed to create secret"

internal/cloud/realm/secrets.go
makesitgo: yes, you should add that test in this PR, at least the parts of creating and listing.  Check out `realm/user_ext_test.go` for examples on these types of tests

internal/cloud/realm/secrets.go
```diff
@@ -37,3 +37,25 @@ func (c *client) Secrets(groupID, appID string) ([]Secret, error) {
...
+		http.MethodPost,
+		fmt.Sprintf(secretsPathPattern, groupID, appID),
+		createSecretRequest{name, value},
+		api.RequestOptions{},
+	)
+	if resErr != nil {
+		return resErr
+	}
+	if res.StatusCode != http.StatusCreated {
+		return api.ErrUnexpectedStatusCode{"secrets create", res.StatusCode}
```
makesitgo, line 57: the pattern for these is `{action} {resource}`... use `create secret` here

internal/commands/secrets/create.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Handler is the command handler
+func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI) error {
+	app, appErr := app.Resolve(ui, cmd.realmClient, cmd.inputs.Filter())
+	if appErr != nil {
+		return appErr
+	}
+
+	if app.ID == "" {
```
makesitgo, line 44: what scenario is this guarding against?  when could we resolve an app without an id?

internal/commands/secrets/create.go
```diff
@@ -0,0 +1,58 @@
...
+		return nil
+	}
+
+	secretsErr := cmd.realmClient.CreateSecret(app.GroupID, app.ID, cmd.inputs.Name, cmd.inputs.Value)
+	if secretsErr != nil {
```
makesitgo, line 49: [nit] just return the `cmd.realmClient` call to end the handler here

internal/commands/secrets/create_inputs.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+func (i *createInputs) Resolve(profile *cli.Profile, ui terminal.UI) error {
+	if i.Name == "" {
+		if err := ui.AskOne(&i.Name, &survey.Input{Message: "Secret Name"}); err != nil {
+			return err
+		}
+	}
+	if i.Value == "" {
+		if err := ui.AskOne(&i.Value, &survey.Input{Message: "Secret Value"}); err != nil {
```
makesitgo, line 23: ok this is the right idea (if input is blank, ask the question), but we can handle this a little differently with the way out `inputs` work.  check out the way `login/inputs.go` works, where we collect a slice of `*survey.Question`s and then prompt with:

```go
if err := ui.Ask(i, questions...); err != nil {
  return err
}
```

They key is that `survey.Question` takes a `Name` field which maps to field names found on `i`, binding the answers per question to each field

internal/commands/secrets/create_inputs_test.go
```diff
@@ -0,0 +1,102 @@
+package secrets
+
+import (
+	"testing"
+
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/10gen/realm-cli/internal/utils/test/mock"
+	"github.com/Netflix/go-expect"
```
makesitgo, line 9: newline here

internal/commands/secrets/create_inputs_test.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+func TestCreateInputs(t *testing.T) {
+	for _, tc := range []struct {
+		description    string
+		createInputs   createInputs
```
makesitgo, line 15: [meganit] just name this inputs

internal/commands/secrets/create_inputs_test.go
```diff
@@ -0,0 +1,102 @@
...
+		prepareProfile func(p *cli.Profile)
```
makesitgo, line 16: is `prepareProfile` even used in this test block?  might be able to get rid of it

internal/commands/secrets/create_inputs_test.go
```diff
@@ -0,0 +1,102 @@
...
+	}{
+		{
+			description: "should prompt for secret name when not provided",
+			createInputs: createInputs{
+				Value: "value",
+			},
+			prepareProfile: func(p *cli.Profile) {},
+			procedure: func(c *expect.Console) {
+				c.ExpectString("Secret Name")
+				c.SendLine("name")
```
makesitgo, line 28: im still a little muddy myself about the whole `ExpectEOF` stuff, but I'm surprised this test works here without it?  I'm gonna play around with this a little more locally

internal/commands/secrets/create_inputs_test.go
```diff
@@ -0,0 +1,102 @@
...
+				c.ExpectString("Secret Value")
+				c.SendLine("value")
+				c.ExpectEOF()
+			},
+			test: func(t *testing.T, i createInputs) {
+				assert.Equal(t, "name", i.Name)
```
makesitgo, line 60: [nit] could make this `assert.Equal(t, createInputs{"name",  "value"}, i)`  (and could likewise do the same for the above tests... just gives a little more assurance we aren't modifying more things than we expect

internal/commands/secrets/create_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+		assert.Nil(t, cmd.Setup(profile, nil))
+		assert.NotNil(t, cmd.realmClient)
+	})
+}
+
+func TestSecretsCreateHandler(t *testing.T) {
+	projectID := "projectID"
+	appID := "appID"
+	testApp := realm.App{
```
makesitgo, line 29: reminder to "take back the `app` name" after rebasing off #55

internal/commands/secrets/create_test.go
```diff
@@ -0,0 +1,118 @@
...
+		Name:        "eggcorn",
+	}
+
+	t.Run("should create app secrets", func(t *testing.T) {
+		realmClient := mock.RealmClient{}
+		realmClient.FindAppsFn = func(filter realm.AppFilter) ([]realm.App, error) {
+			return []realm.App{testApp}, nil
+		}
+
+		realmClient.CreateSecretFn = func(groupID, appID, name, value string) error {
```
makesitgo, line 42: can we add capture `var`s here and then assert below that we are correctly passing in inputs like we expect here?

internal/commands/secrets/flags.go
```diff
@@ -0,0 +1,11 @@
+package secrets
```
makesitgo, line 1: this file can probably go away ... just put these `const` declarations into the `input` file for this command

makesitgo (commented):
couple of initial thoughts here, nice turnaround on this

internal/commands/secrets/flags.go
kpatel71716: I was thinking it could be reused for `secrets update` and `secrets delete`. Though you bring up a good point that maybe we could have a generic inputs file for all of those instead of create_inputs, update_inputs, etc?

internal/commands/secrets/create_inputs_test.go
kpatel71716: Taking a look at this too. I can add it back in since it is probably best practice from what I see? Will play around with this and see what happens

internal/commands/secrets/flags.go
jo-tse-mongo: Possibly, but for `secrets delete` I'm planning on having either the `ID` or `Name` as a flag, which might not be reusable with other `delete_inputs`

kpatel71716: Ah ok I see that now. IMO the best route would to follow what Nick said and each command will have it's own inputs, even if update and delete could reuse that code. Keeps things separate so I like that trade off

makesitgo: yeah, with the nature of some of these commands, we are just inevitably going to see a lot of repetition.  but we should embrace that!  especially because things like flags and what not aren't especially tricky concepts, we can afford that duplication to keep things decoupled

internal/commands/secrets/create.go
kpatel71716: That's correct, I removed

internal/cloud/realm/secrets_ext_test.go
```diff
@@ -0,0 +1,60 @@
...
+	t.Run("with an active session ", func(t *testing.T) {
+		client := newAuthClient(t)
+		groupID := u.CloudGroupID()
+
+		secretName := "secretName"
+		secretValue := "secretValue"
+
+		testApp, appErr := client.CreateApp(groupID, "secrets-test", realm.AppMeta{})
+		assert.Nil(t, appErr)
+		defer client.DeleteApp(groupID, testApp.ID)
```
kpatel71716, line 33: [q] I added a defer here but if we want to be more explicit here, especially if we wanted to log any error that could come up, I can refactor this. I felt this was fine since there isn't much we can do here, the impact would be a floating app that we can easily attribute to this test, and we have e2e tests already covering app delete

internal/cloud/realm/secrets.go
```diff
@@ -37,3 +37,25 @@ func (c *client) Secrets(groupID, appID string) ([]Secret, error) {
...
+		http.MethodPost,
+		fmt.Sprintf(secretsPathPattern, groupID, appID),
+		createSecretRequest{name, value},
+		api.RequestOptions{},
+	)
+	if resErr != nil {
+		return resErr
+	}
+	if res.StatusCode != http.StatusCreated {
+		return api.ErrUnexpectedStatusCode{"create secrets", res.StatusCode}
```
makesitgo, line 57: [meganit] `"create secret"` (singular)

internal/cloud/realm/secrets.go
```diff
@@ -37,3 +37,25 @@ func (c *client) Secrets(groupID, appID string) ([]Secret, error) {
...
+	}
+
```
makesitgo, line 59: [nit] remove this newline

internal/cloud/realm/secrets_ext_test.go
makesitgo: yeah, honestly I haven't been a good citizen here in regards to test cleanup yet, so let's make a helper here in the `realm_test` package, something like:

```go
func setupTestApp(t *testing.T, groupID, name string) (realm.App, func()) {
  t.Helper()
  // create app with empty meta here
  // assert create err was nil
  // teardown := func() { /* call delete app, you can use `t.Log` to print back any errors */ }
}
```

makesitgo: yeah, honestly I haven't been a good citizen here in regards to test cleanup yet, so let's make a helper here in the `realm_test` package, something like:

```go
func setupTestApp(t *testing.T, groupID, name string) (realm.App, func()) {
  t.Helper()
  // create app with empty meta here
  // assert create err was nil
  // teardown := func() { /* call delete app, you can use `t.Log` to print back any errors */ }
}
```

internal/cloud/realm/secrets_ext_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+		assert.Nil(t, client.Import(groupID, testApp.ID, map[string]interface{}{
+			app.NameAuthProviders: []map[string]interface{}{
```
makesitgo, line 36: [question] do you need this import call?  why are auth providers necessary to be present when working with app secrets?

internal/cloud/realm/secrets_ext_test.go
```diff
@@ -0,0 +1,60 @@
...
+			assert.Nil(t, secretsErr)
+			assert.Equal(t, 0, len(secrets))
+		})
+		t.Run("should create a secret", func(t *testing.T) {
+			assert.Nil(t, client.CreateSecret(groupID, testApp.ID, secretName, secretValue))
+		})
+		t.Run("should list a created secret", func(t *testing.T) {
+			secrets, secretsErr := client.Secrets(groupID, testApp.ID)
+			assert.Nil(t, secretsErr)
+			assert.Equal(t, 1, len(secrets))
```
makesitgo, line 56: these two asserts can go away with just:

```go
assert.Equal(t, []Secret{{secretName, secretValue}}, secrets)
```

internal/commands/commands.go
```diff
@@ -106,6 +106,13 @@ var (
 		Description: "Manage the secrets of your MongoDB Realm app",
 		Help:        "secrets",
 		SubCommands: []cli.CommandDefinition{
+			cli.CommandDefinition{
+				Use:         "create",
+				Display:     "secrets create",
+				Description: "Creates a secret for a Realm application",
```
makesitgo, line 112: was this from the design team's page?  I think this should be `Realm app` according to their standards

internal/commands/secrets/create_inputs.go
```diff
@@ -0,0 +1,52 @@
+package secrets
+
+import (
+	"github.com/10gen/realm-cli/internal/app"
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/AlecAivazis/survey/v2"
```
makesitgo, line 7: newline

internal/commands/secrets/create_inputs.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+const (
+	flagName      = "secret-name"
```
makesitgo, line 11: ignore my spec now that we can't have `-` flags with more than one letter to them... the long forms of these flags should be `--name` and `--value` here (since `secret` is already known from the command itself)

internal/commands/secrets/create_inputs.go
```diff
@@ -0,0 +1,52 @@
...
+	flagNameShort = "n"
+	flagNameUsage = `the name of the secret to add to your Realm App`
+
+	flagValue      = "secret-value"
+	flagValueShort = "v"
+	flagValueUsage = `the value of the secret to add to your Realm App`
+
+	createInputFieldSecretName  = "Name"
```
makesitgo, line 19: [meganit] these field consts actually should be able to be lowercased (but still snakeCased) and survey recognizes them

internal/commands/secrets/create_inputs.go
```diff
@@ -0,0 +1,52 @@
...
+
+	if i.Value == "" {
+		questions = append(questions, &survey.Question{
+			Name:   createInputFieldSecretValue,
+			Prompt: &survey.Password{Message: "Secret Value"},
+		})
+	}
+
+	if len(questions) > 0 {
+		if err := ui.Ask(i, questions...); err != nil {
```
makesitgo, line 47: [nit] just `return ui.Ask(i, questions...)` here

makesitgo (commented):
just a few questions/suggestions for now, lookin good!

internal/cloud/realm/secrets_ext_test.go
kpatel71716: The `client.Secrets()` call returns an object with an ID and Name so I went with the 2-liner here since we don't check the id. We could modify `client.CreateSecret` to return an ID, capture the ID in an earlier test and then check the assert here. I felt CreateSecret was fine just returning an error if we don't display the ID of the created secret in the command output

internal/cloud/realm/secrets_ext_test.go
kpatel71716: This was my bad from skimming over it while copy/paste lol. Removing

internal/cloud/realm/secrets_ext_test.go
kpatel71716: Done! Added a helpers_test.go file. I also added the realm.Client as an argument

internal/cloud/realm/secrets_ext_test.go
makesitgo: sorry, I meant turn
```go
assert.Equal(t, 1, len(secrets))
assert.Equal(t, secretName, secrets[0].Name)
```
into
```go
assert.Equal(t, []Secret{{secretName, secretValue}}, secrets)
```

internal/cloud/realm/secrets_ext_test.go
makesitgo: instead of `helpers_test.go`, just place it in the existing `app_test.go` file

internal/cloud/realm/secrets_ext_test.go
kpatel71716: I may have misinterpreted but I was referring to those lines as well. the realm.Secret object has `ID` and 'Name` as opposed to `Name` and `Value` so there would be an assert error on `assert.Equal(t, []Secret{{secretName, secretValue}}, secrets)` since `secretValue` won't match the ID field returned bt the `Secrets()` call

makesitgo: I gotcha now...sorry should have read closer (I was getting request/response mixed up therE).  this is fine as is then!

kpatel71716: No worries, all good!

internal/cloud/realm/secrets_ext_test.go
kpatel71716: Done! I left app_test as is (i.e. didn't use the helper there). I am fine with modifying app_test if we want to reuse the helper inside of there (I was a bit split on that). I also don't mind updating the other tests in a follow-up/this PR

internal/cloud/realm/app_test.go
```diff
@@ -50,3 +51,16 @@ func TestRealmApps(t *testing.T) {
...
 	})
 }
+
+func setupTestApp(t *testing.T, client realm.Client, groupID, name string) (realm.App, func()) {
+	t.Helper()
+	app, err := client.CreateApp(groupID, name, realm.AppMeta{})
+	assert.Nil(t, err)
+	teardown := func() {
+		deleteErr := client.DeleteApp(groupID, app.ID)
+		if deleteErr != nil {
```
makesitgo, line 61: [nit] just inline this: `if err := client...; err != nil {`

internal/cloud/realm/app_test.go
```diff
@@ -50,3 +51,16 @@ func TestRealmApps(t *testing.T) {
...
+			t.Log(fmt.Sprintf("failed to delete test app with groupID: %v and ID: %v", groupID, app.ID))
```
makesitgo, line 62: should be able to just use `t.Logf` here... lets rework this message too:

```go
t.Logf("warning: failed to delete test app (id: %s): %s", app.ID, err)
```

internal/cloud/realm/secrets_ext_test.go
makesitgo: ok last bit on this: update `CreateSecret` to return `(Secret, error)` (since that's actually what the API returns to us).  from there, you can assert on `.Name` being correct in the "create" test, and you can assert on a whole `[]Secret` slice in the ensuing "find all created" test

internal/cloud/realm/secrets.go
```diff
@@ -37,3 +37,24 @@ func (c *client) Secrets(groupID, appID string) ([]Secret, error) {
 	}
 	return secrets, nil
 }
+
+type createSecretRequest struct {
+	Name  string `json:"name"`
+	Value string `json:"value"`
+}
+
+func (c *client) CreateSecret(groupID, appID, name, value string) error {
```
makesitgo, line 46: should return `(Secret, error)` here

makesitgo (approved):
last few tweaks here, but after those and once the commits and evergreen are all in line ... this will be good to go.  nice stuff!

🆒  beans

internal/cloud/realm/app_test.go
kpatel71716: Perfect, learned something new today!

internal/cloud/realm/secrets_ext_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	t.Run("with an active session ", func(t *testing.T) {
+		client := newAuthClient(t)
+		groupID := u.CloudGroupID()
+
+		secretName := "secretName"
+		secretValue := "secretValue"
+
+		testApp, teardown := setupTestApp(t, client, groupID, "secrets-test")
+		defer teardown()
```
kpatel71716, line 31: @nathantfrank FYI once this is merged, these 2 lines can be used in the user_ext test. If you merge first, I can add them in there for `user revoke`

PR #58, opened 2021-02-03
REALMC-7653: Make config version 20210101 the default cli option for import/export

makesitgo:


internal/utils/test/assert/assert.go
```diff
@@ -29,47 +29,52 @@ func RegisterOpts(t reflect.Type, opts ...cmp.Option) {
...
 // Equal compares x and y for equality
 // and fails with the test if not
-func Equal(t testing.TB, x, y interface{}) {
+func Equal(t testing.TB, expected, actual interface{}) {
 	t.Helper()
-	switch x.(type) {
+	switch expected.(type) {
 	case string:
-		Equalf(t, x, y, "expected %q to equal %q", y, x)
+		Equalf(t, expected, actual, "failed to assert equals ( actual, expected )\n\t%q\n\t%q", actual, expected)
```
makesitgo, line 36: ⚠️ the changes in this file are aesthetic... the diffs of expected vs actual are now lined up separately, making it a bit easier to visually compare right off the bat

internal/local/structure_v2.go
```diff
@@ -127,21 +109,32 @@ func (a *AppDataV2) unmarshalAuth(rootDir string) error {
 	}
 
 	var auth AuthStructure
-
 	cfg, cfgErr := readFile(filepath.Join(dir, FileConfig.String()))
 	if cfgErr != nil {
 		return cfgErr
 	}
+	if err := unmarshalJSON(cfg, &auth.Config); err != nil {
+		fmt.Println("failed to read config", string(cfg))
```
UnicodeSnowman, line 117: is there a better way to write to stdout here via some CLI util vs. `fmt`? or is this just for debugging?

UnicodeSnowman (approved):

internal/local/structure_v2.go
makesitgo: just debug code, thanks for catching

tsedgwick (approved):

PR #60, opened 2021-02-05
REALMC-7165 Add secrets delete command to realm-cli 

jo-tse-mongo:
[WIP] in regards to tests


internal/commands/secrets/delete.go
```diff
@@ -0,0 +1,137 @@
...
+	sort.SliceStable(cmd.outputs, secretOutputComparerBySuccess(cmd.outputs))
+	logs := terminal.NewTableLog(
+		secretDeleteMessage,
+		secretHeaders(),
+		secretTableRows(cmd.outputs, secretDeleteRow)...,
+	)
+	return ui.Print(logs)
+}
+
+func (i *deleteInputs) resolveDelete(args []string, secrets []realm.Secret, ui terminal.UI) ([]realm.Secret, error) {
```
makesitgo, line 80: can we change this function signature to be:

```go
func (i *deleteInputs) resolveSecrets(ui terminal.UI, client realm.Client, app realm.App) ([]realm.Secret, error) {
  // fetch the appSecrets via the realm.Client, you probably want these put into a map[name]Secret

  // if len(i.Secrets) == 0, then we'll want to prompt the user to select (many) from the list of app secrets

  // else iterate over each i.Secret flag and figure out its corresponding ID/realm.Secret

  // return that resolved array of realm.Secret back to the command handler
}
```

basically just shifting around the bits of logic you already have.  and we shouldn't be using `args` for anything here, just rely on the flags setting values

---

also this should move to the `delete_inputs.go` file

internal/commands/secrets/delete.go
```diff
@@ -0,0 +1,137 @@
...
+			toDelete[i] = s
+		}
+	}
+
+	return toDelete, nil
+}
+
+func secretDeleteRow(output secretOutput, row map[string]interface{}) {
+	if output.err != nil {
+		row[headerDetails] = output.err.Error()
```
makesitgo, line 133: I'm sure tests will prove this out, but let's make sure having `nil` values for `details`/`deleted` is ok when the table renders as output text.  Check with @kpatel71716 and @nathantfrank on where we landed with rendering `bool` values in the table (was it as "yes"/"no" or something else?), and then make sure we're consistent here

internal/commands/secrets/delete.go
```diff
@@ -0,0 +1,137 @@
...
+func (cmd *CommandDelete) Inputs() cli.InputResolver {
+	return &cmd.inputs
+}
+
+func (cmd *CommandDelete) Flags(fs *pflag.FlagSet) {
+	cmd.inputs.Flags(fs)
+	fs.StringSliceVarP(&cmd.inputs.secrets, flagSecret, flagSecretShort, []string{}, flagSecretUsage)
+}
+
+func (i *deleteInputs) Resolve(profile *cli.Profile, ui terminal.UI) error {
```
makesitgo, line 27: this should move to the `delete_inputs.go` file

internal/commands/secrets/output.go
```diff
@@ -1,6 +1,56 @@
...
+
+type secretTableRowModifier func(secretOutput, map[string]interface{})
+
+func secretOutputComparerBySuccess(outputs secretOutputs) func(i, j int) bool{
+	return func(i, j int) bool {
+		return outputs[i].err != nil && outputs[j].err == nil
+	}
+}
+
+func secretHeaders() []string{
```
makesitgo, line 32: check out the latest code in the `user` package with how they're handling `headers` and `rows`.  I think it super close to the way you have it here, but can we adjust this approach so it can be similarly used in `secrets list`?

makesitgo (commented):
just a quick pass here... just a few small things for now.  lookin real good so far!

internal/commands/secrets/delete.go
jo-tse-mongo: From what I remember about `parseValue` is that it can take in nil, but it'll leave it as an empty string

makesitgo: i think i wrote this comment on the wrong line, but I was more concerned with the `bool` vs `nil` treatment.  I think we want `true`/`false` to render as `"yes"`/`"no"`?

internal/cloud/realm/secrets.go
```diff
@@ -27,10 +28,11 @@ func (c *client) Secrets(groupID, appID string) ([]Secret, error) {
 	if resErr != nil {
 		return nil, resErr
 	}
+
 	if res.StatusCode != http.StatusOK {
 		return nil, api.ErrUnexpectedStatusCode{"secrets", res.StatusCode}
 	}
-	defer res.Body.Close()
```
makesitgo, line 33: we definitely still want this here, as we read from the `res.Body` next here

internal/cloud/realm/secrets.go
```diff
@@ -63,3 +65,22 @@ func (c *client) CreateSecret(groupID, appID, name, value string) (Secret, error
...
+		fmt.Sprintf(secretPathPattern, groupID, appID, secretID),
+		api.RequestOptions{},
+	)
+
+	if resErr != nil {
+		return resErr
+	}
+
+	if res.StatusCode != http.StatusNoContent {
+		defer res.Body.Close()
```
makesitgo, line 81: and you won't need the `defer` here, since we never actually attempt to read the `res.Body` anywhere here

internal/cloud/realm/secrets_test.go
```diff
@@ -46,5 +47,28 @@ func TestSecrets(t *testing.T) {
 				assert.Equal(t, []realm.Secret{secret}, secrets)
 			})
 		})
+
+		t.Run("should create secrets for deletion", func(t *testing.T) {
+			testLen := 3
+			testSecrets := make([]realm.Secret, testLen)
+			for i := 0; i < testLen; i++ {
```
makesitgo, line 54: we probably don't even need all of this here (though it's great testing!).  for the most part, leave the file as is, and just add a few extra tests... it should read something like:

```go
t.Run("should fail without an auth client", func(t *testing.T) {
  
  t.Run("with an active session ", func(t *testing.T) {

    // FIXME: actually we can just delete this test case now, its effectively duplicated with the new delete stuff you'll be adding
    t.Run("should have no secrets upon app initialization", func(t *testing.T) {

    t.Run("should create a secret", func(t *testing.T) {
      t.Run("and list all app secrets", func(t *testing.T) {})
    })

    t.Run("should delete a secret", func(t *testing.T) {

      // TODO: delete the created test secret here

      t.Run("and list all app secrets", func(t *testing.T) {

        // TODO: find all test secrets again and assert it's empty

      })
    })
    })
  })
})
```

internal/commands/secrets/delete.go
```diff
@@ -0,0 +1,89 @@
...
+	if appErr != nil {
+		return appErr
+	}
+
+	secretList, secretListErr := cmd.realmClient.Secrets(app.GroupID, app.ID)
+	if secretListErr != nil {
+		return secretListErr
+	}
+
+	toDelete, resolveErr := cmd.inputs.resolveSecrets(ui, secretList)
```
makesitgo, line 49: NOTE: once @kpatel71716 merges his `user revoke` work, we'll have an established pattern for commands structured like this going forward.  i think it'll be worth your time here to make the necessary refactors to line up with that pattern, so going forward `users` AND `secrets` can be considered "scope-complete"

internal/commands/secrets/output.go
```diff
@@ -1,6 +1,56 @@
...
+
 const (
-	headerID   = "ID"
-	headerName = "Name"
+	headerID      = "ID"
+	headerName    = "Name"
+	headerDeleted = "Deleted"
+	headerDetails = "Details"
+
+	secretDeleteMessage = "Deleted Secrets"
```
makesitgo, line 14: is this used anywhere?

internal/commands/secrets/output.go
```diff
@@ -1,6 +1,56 @@
...
+type secretTableHeaderModifier func() []string
+type secretTableRowModifier func(secretOutput, map[string]interface{})
+
+func secretOutputComparerBySuccess(outputs secretOutputs) func(i, j int) bool {
+	return func(i, j int) bool {
+		return outputs[i].err != nil && outputs[j].err == nil
+	}
+}
+
+func secretHeaders(modifier secretTableHeaderModifier) []string {
```
makesitgo, line 33: [question] could this just be `func secretHeaders(additionalHeaders ...string) []string`?

internal/commands/secrets/delete_inputs.go
```diff
@@ -0,0 +1,83 @@
...
+	if err := i.ProjectInputs.Resolve(ui, profile.WorkingDirectory); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (i *deleteInputs) resolveSecrets(ui terminal.UI, allSecrets []realm.Secret) ([]realm.Secret, error) {
+	var toDelete []realm.Secret
+
+	if len(i.secrets) == 0 {
```
makesitgo, line 32: can we flip this structure a bit for readability purposes? something like:

```
func (i *deleteInputs) resolveSecrets(ui terminal.UI, secrets []realm.Secret) ([]realm.Secret, error) {
  if len(i.secrets) > 0 {
    resolved := make([]realm.Secret, len(i.secrets))
    // ...
    return resolved, nil
  }

  // no "else" needed here since we return above, less indentation for the next code path
}
```

internal/commands/secrets/delete_inputs.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		var selectedSecrets []string
+		askErr := ui.AskOne(
+			&selectedSecrets,
+			&survey.MultiSelect{
+				Message: "Which secret(s) would you like to delete?",
+				Options: selectableSecretOptions,
+			},
+		)
+
```
makesitgo, line 49: [stylenit] omit this newline, also optionally rename this to `err` and/or inline the if-call

internal/commands/secrets/delete_inputs.go
```diff
@@ -0,0 +1,83 @@
...
+		ids := make(map[string]realm.Secret, len(allSecrets))
+		names := make(map[string]realm.Secret, len(allSecrets))
+		for _, secret := range allSecrets {
+			ids[secret.ID] = secret
+			names[secret.Name] = secret
+		}
+
+		for i, arg := range i.secrets {
+			if _, ok := names[arg]; ok {
```
makesitgo, line 73: the first arg (`_`_ here) is actually the value itself, so you can use that instead of `names[arg]` a 2nd time here (same for `id below)

internal/commands/secrets/delete_inputs_test.go
```diff
@@ -0,0 +1,98 @@
...
+			selectedSecrets: []string{"secret_name_2"},
+			expectedOutput:  []realm.Secret{secrets[2]},
+		},
+		{
+			description:     "select by name and ID",
+			selectedSecrets: []string{"secret_name_2", "secret_id_4", "secret_name_1"},
+			expectedOutput:  []realm.Secret{secrets[2], secrets[4], secrets[1]},
+		},
+	} {
+		t.Run("should return the realm.Secret values from deleteInputs: "+tc.description, func(t *testing.T) {
```
makesitgo, line 43: a note about these `t.Run` strings: try to keep things lowercase and avoid using punctation like `.` or `:` in  (it ends up looking a bit wonky in the resulting `go test` output)

something like `"should return the secrets when specified by"+tc.description` would suffice here

internal/commands/secrets/delete_inputs_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+			_, console, _, ui, consoleErr := mock.NewVT10XConsole()
+			assert.Nil(t, consoleErr)
+			defer console.Close()
+
+			doneCh := make(chan struct{})
+			go func() {
+				defer close(doneCh)
+				console.ExpectString("Which secret(s) would you like to delete?")
+				for _, selected := range tc.selectedSecrets {
```
makesitgo, line 81: really neat! 🔥

internal/commands/secrets/delete_test.go
```diff
@@ -0,0 +1,310 @@
...
+		description    string
+		testInput      []string
+		expectedOutput []secretOutput
+		expectedErr    error
+	}{
+		{
+			description: "should return successful outputs for proper secret inputs",
+			testInput:   []string{"secret_id_1"},
+			expectedOutput: []secretOutput{
+				{
```
makesitgo, line 44: [stylenit] these are small enough to be defined inline: `{secrets[1], nil}` (and this particular `[]secretOutput` array can probably fit all in-line

internal/commands/secrets/delete_test.go
```diff
@@ -0,0 +1,310 @@
...
+			assert.Equal(t, projectID, capturedDeleteProjectID)
+			assert.Equal(t, appID, capturedDeleteAppID)
+		})
+	}
+
+	t.Run("should return an error", func(t *testing.T) {
+		successfulAppFn := func(filter realm.AppFilter) ([]realm.App, error) {
+			return []realm.App{app}, nil
+		}
+		successfulSecretsFn := func(groupID, appID string) ([]realm.Secret, error) {
```
makesitgo, line 120: [question] are these used anywhere else? they probably can just be defined right in `clientSetup`

internal/commands/secrets/delete_test.go
```diff
@@ -0,0 +1,310 @@
...
+						SecretsFn: successfulSecretsFn,
+					}
+				},
+				expectedErr: errors.New("Something went wrong with the app"),
+			},
+		} {
+			t.Run(tc.description, func(t *testing.T) {
+				realmClient := tc.clientSetup()
+
+				cmd := &CommandDelete{
```
makesitgo, line 156: [stylenit] could make this all one line

makesitgo (commented):
bravo!  going to comment here and wait for the last few changes here, but nothing blocking by me.  this looks great!

internal/cloud/realm/secrets.go
jo-tse-mongo: oops this was a mistake

internal/commands/secrets/delete_test.go
jo-tse-mongo: they're used multiple times in the different test cases for this set

internal/commands/secrets/output.go
jo-tse-mongo: The table displays some message before the actual table itself, so it's used there
It's called "message" so I'm not sure what to name it or if it should be a const

makesitgo: we can remove the `const` and just leave the string value as an arg to `terminal.NewTable(...`

internal/commands/commands.go
```diff
@@ -149,6 +149,13 @@ Realm app will be exported as well.`,
 				Help:        "Displays a list of data tables. Each data table displays the Name and ID of one (1) secret.",
 				Command:     &secrets.CommandList{},
 			},
+			cli.CommandDefinition{
+				Use:         "delete",
+				Display:     "secret delete",
```
kpatel71716, line 154: [nit] We should probably use `secrets delete` here to be consistent with pluralization

internal/cloud/realm/secrets.go
```diff
@@ -63,3 +67,21 @@ func (c *client) CreateSecret(groupID, appID, name, value string) (Secret, error
 	}
 	return secret, nil
 }
+
+func (c *client) DeleteSecret(groupID, appID, secretID string) error {
+	res, resErr := c.do(
```
kpatel71716, line 72: [nit/reminder] that we are going with just `err` here for naming standards

kpatel71716 (commented):
Looks pretty good to me! Just some really small things

internal/cloud/realm/secrets_test.go
```diff
@@ -46,5 +47,20 @@ func TestSecrets(t *testing.T) {
 				assert.Equal(t, []realm.Secret{secret}, secrets)
 			})
 		})
+
+		t.Run("should delete the app secret", func(t *testing.T) {
```
makesitgo, line 51: [optional/style] personally I'd just nest this further under `should create a secret` and avoid the `var secret realm.Secret` declaration at the top

makesitgo (approved):
ibueno!

🆒  beans

makesitgo:
we will wait for @kpatel71716 to merge in #54 so you can rebase off his test fixes, and then evergreen should be good to go for this

PR #62, opened 2021-02-08
REALMC-8006: Include http_endpoints to new import request

makesitgo:
to be merged after #61

### DIFF
--- a/internal/local/structure.go
+++ b/internal/local/structure.go
@@ -1,0 +1,219 @@
+package local
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// GraphQLStructure represents the Realm app graphql structure
+type GraphQLStructure struct {
+	Config          map[string]interface{}   `json:"config,omitempty"`
+	CustomResolvers []map[string]interface{} `json:"custom_resolvers,omitempty"`
+}
+
+// SecretsStructure represents the Realm app secrets
+type SecretsStructure struct {
+	AuthProviders map[string]map[string]string `json:"auth_providers,omitempty"`
+	Services      map[string]map[string]string `json:"services,omitempty"`
+}
+
+// ServiceStructure represents the Realm app service structure
+type ServiceStructure struct {
+	Config           map[string]interface{}   `json:"config,omitempty"`
+	IncomingWebhooks []map[string]interface{} `json:"incoming_webhooks"`
+	Rules            []map[string]interface{} `json:"rules"`
+}
+
+func parseEnvironments(rootDir string) (map[string]map[string]interface{}, error) {
+	out := map[string]map[string]interface{}{}
+
+	dw := directoryWalker{
+		path:      filepath.Join(rootDir, NameEnvironments),
+		onlyFiles: true,
+	}
+	if err := dw.walk(func(file os.FileInfo, path string) error {
+		o, err := parseJSON(path)
+		if err != nil {
+			return err
+		}
+		out[file.Name()] = o
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	if len(out) == 0 {
+		return nil, nil
+	}
+	return out, nil
+}
+
+func parseFunctions(rootDir string) ([]map[string]interface{}, error) {
+	if _, err := os.Stat(rootDir); err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	var out []map[string]interface{}
+
+	dw := directoryWalker{path: rootDir, onlyDirs: true}
+	if walkErr := dw.walk(func(file os.FileInfo, path string) error {
+		if strings.Contains(path, nameNodeModules) {
+			return nil // skip node_modules
+		}
+
+		config, configErr := parseJSON(filepath.Join(path, FileConfig.String()))
+		if configErr != nil {
+			return configErr
+		}
+
+		src, srcErr := parseJavascript(path, FileSource)
+		if srcErr != nil {
+			return srcErr
+		}
+
+		out = append(out, map[string]interface{}{
+			NameConfig: config,
+			NameSource: src,
+		})
+		return nil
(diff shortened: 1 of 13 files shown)


internal/local/structure.go
```diff
@@ -0,0 +1,222 @@
...
+	src, err := ioutil.ReadFile(filepath.Join(rootDir, file.String()))
+	if err != nil {
+		return "", err
+	}
+	return string(src), nil
+}
+
+func parseJavascriptSource(rootDir string) (string, error) {
+	return parseJavascript(rootDir, FileSource)
+}
```
tsedgwick, line 122: [nit] I'm always against a one line function abstraction

internal/local/structure.go
```diff
@@ -0,0 +1,222 @@
...
+
+	return GraphQLStructure{config, customResolvers}, true, nil
+}
+
+func parseJavascript(rootDir string, file File) (string, error) {
+	src, err := ioutil.ReadFile(filepath.Join(rootDir, file.String()))
+	if err != nil {
+		return "", err
+	}
+	return string(src), nil
```
tsedgwick, line 117: more of a question - do we actually need this as a string? or can we keep this as a []byte?

internal/local/structure.go
```diff
@@ -0,0 +1,222 @@
...
+	var out map[string]interface{}
+	if err := unmarshalJSON(data, &out); err != nil {
+		return nil, err
+	}
+	return out, nil
+}
+
+func parseJSONConfig(rootDir string) (map[string]interface{}, error) {
+	return parseJSON(filepath.Join(rootDir, FileConfig.String()))
+}
```
tsedgwick, line 139: same nit as above

internal/local/structure.go
```diff
@@ -0,0 +1,222 @@
...
+	if dataErr != nil {
+		return nil, dataErr
+	}
+
+	var out map[string]interface{}
+	if err := unmarshalJSON(data, &out); err != nil {
+		return nil, err
+	}
+	return out, nil
+}
```
tsedgwick, line 135: do we really need this func? doesn't seem like we gain too much other than abstracting away 9 lines of code 🤷🏻

internal/local/structure.go
```diff
@@ -0,0 +1,222 @@
...
+
+func parseJSONSecrets(rootDir string) (*SecretsStructure, error) {
+	path := filepath.Join(rootDir, FileSecrets.String())
+
+	if ok, err := fileExists(path); err != nil {
+		return nil, err
+	} else if !ok {
+		return nil, nil // if secrets.json does not exist, continue
+	}
```
tsedgwick, line 173: wouldn't fileExists throw a FileNotFound error?

internal/local/structure.go
```diff
@@ -0,0 +1,222 @@
...
+		out = append(out, o)
+		return nil
+	}); walkErr != nil {
+		return nil, walkErr
+	}
+
+	return out, nil
+}
+
+func parseJSONSecrets(rootDir string) (*SecretsStructure, error) {
```
tsedgwick, line 166: [nit] and for all the other funcs - can we just call this parseSecrets? I can't imagine we ever support anything other than JSON. Plus, with the abstraction of this func the caller doesn't need to be aware of what format the data is in.

tsedgwick: Also - I would probably opt for this to return an io.Reader or []byte and then pull the json.Marshal up into the calling func. this could simply become`readSecrets`

tsedgwick (commented):
mostly questions / nits

internal/local/structure.go
makesitgo: wasn't ever a fan of the word always/never myself 🙃 ... will remove this, had it in place when I was a little uncertain about how functions/webhooks were going to play out

internal/local/structure.go
makesitgo: we do actually need this as a string in our current server implementation, i was burned by this a few weeks back in a nasty test

internal/local/structure.go
makesitgo: you *really* don't want me to have this helper (it's been removed and added back now about 2-3 different times in various forms).

we use this function/chunk of code currently 24 times in our app structure/parsing code, and that's not yet even our cap.  it really does have an impact on the code heft/readability in this case

internal/local/structure.go
makesitgo: not if im understanding the implementation correctly:

```go
func fileExists(path string) (bool, error) {
	_, err := os.Stat(path)
	if err == nil {
		return true, nil
	}
	if os.IsNotExist(err) {
		return false, nil
	}
	return false, err
}
```

the motivation for even creating this helper was simply to flatten out the error handling from `os.Stat` so there wasn't that nested check for `os.IsNotExist`

internal/local/structure.go
tsedgwick: 😅  alright I'll stop commenting on this. carry on...

internal/local/structure.go
makesitgo: yeah, i can change that name to remove `JSON` from it, but im going to stick with read/marshaling alongside each other in these funcs.  its logic that's always executed together within this CLI context and eventually.  i also like this breakdown for the inevitable `functions` unit tests I'm going to be writing, making sure we're parsing and collecting data from that directory properly

internal/local/structure.go
tsedgwick: Seems like we just replaced one check with another though, no? We actually double up the check by using this helper since the helper func checks the IsNotExists error.

`os.Stat -> error -> IsNotExist (true) -> outer func checks error (nil) -> outer func checks ok (false) -> on to next step`

where it could just become 

`os.Stat -> error -> IsNotExist (true) -> on to next step`

This has been in there so I'm fine without the refactor in this PR, I just wanted to call it out now that I've thought it through more.

tsedgwick (approved):
remaining nits on my end aren't blockers. LGTM

UnicodeSnowman (approved):

PR #64, opened 2021-02-08
REALMC-7168: Buildout command: app create

nathantfrank:


internal/commands/app/new_app_inputs.go
```diff
@@ -0,0 +1,71 @@
...
+	if i.From != "" {
+		if i.Project == "" {
+			groupID, groupErr := cli.ResolveGroupID(ui, ac)
+			if groupErr != nil {
+				return from{}, groupErr
+			}
+			i.Project = groupID
+		}
+
+		app, err := cli.ResolveApp(ui, rc, realm.AppFilter{GroupID: i.Project, App: i.From})
```
makesitgo, line 62: I think we're conflating `Project` and `From` here a little bit

`Project` is meant to be the `GroupID` we use to specify/make changes to a Realm app, the target group if you will.  it also serves as a first "filter" for when we fetch available apps to the user (as mongo employees, we have a ton of available apps to us).

`From` is simply an app that the user has access to and wants to read data from (in this case to create a similarly structured app).  we can use the `i.Project` flag as an optional filter, but we don't want to "resolve` it until we need to.

for instance, `app init` never _needs_ to resolve the project/group id to write to.
however, `app create`, `push` and `pull` will.

knowing that, i think the `if i.Project == ""` chunk of logic above can probably be moved to its own function on `(i *newAppInputs)`, so that those commands can call it directly when need be in their own handlers/resolves

makesitgo (commented):
first pass here: wanted to get this in before diving in to the test files

internal/commands/app/create_inputs_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+func TestAppCreateInputsResolveAppName(t *testing.T) {
+	testApp := realm.App{
+		ID:          primitive.NewObjectID().Hex(),
+		GroupID:     primitive.NewObjectID().Hex(),
+		ClientAppID: "test-app-abcde",
+		Name:        "test-app",
+	}
+
+	for _, tc := range []struct {
```
makesitgo, line 95: this is a loaded test case `struct`... and im not sure we actually need the interactive console here, right?  (every `procedure` here is an empty func...)

I think the `"should return an error when client fails to find apps"` should be its own `t.Run` block.  you can table the other two together if you'd like, and the test case struct should end up being much simpler

internal/commands/app/create_inputs_test.go
```diff
@@ -0,0 +1,272 @@
...
+				c.ExpectEOF()
+			},
+			test: func(t *testing.T, i createInputs) {
+				assert.Equal(t, "test-app", i.Name)
+				assert.Equal(t, flagDeploymentModelDefault, i.DeploymentModel)
+				assert.Equal(t, flagLocationDefault, i.Location)
+			},
+		},
+		{
+			description: "with a name flag set should prompt for nothing else and set realm.Location and deployment model to defaults",
```
makesitgo, line 40: for tests that "don't prompt" things, lets write them in the style where we aren't even using the interactive/pseudo terminal.  they can simply just be unit tested with `nil` or simple `ui`s

internal/commands/app/create_inputs_test.go
```diff
@@ -0,0 +1,272 @@
...
+			description: "should prompt for project if project is not set",
+			procedure: func(c *expect.Console) {
+				c.ExpectString("Atlas Project")
+				c.Send(testApp.GroupID)
+				c.SendLine(" ")
+				c.ExpectEOF()
+			},
+			expectedProject: testApp.GroupID,
+		},
+		{
```
makesitgo, line 189: same thing about loaded test case `struct`s and not needing pseudo-terminals for tests where we aren't prompting for anything.  these three tests all seem easy enough to test in their own `t.Run` blocks

internal/commands/app/create_inputs_test.go
```diff
@@ -0,0 +1,272 @@
...
+			assert.Equal(t, tc.expectedErr, err)
+			assert.Equal(t, tc.expectedProject, projectID)
+		})
+	}
+}
+
+func TestAppCreateInputsResolveDirectory(t *testing.T) {
+	t.Run("should return path of wd with app name appended", func(t *testing.T) {
+		profile, teardown := mock.NewProfileFromTmpDir(t, "app_create_test")
+		defer teardown()
```
makesitgo, line 226: we don't actually need a temporary directory created here, might be worth adding `mock.NewProfileFromWd` and have that use `os.Getwd` to set `WorkingDirectory` field.  no need for the `teardown` func either then

internal/commands/app/create_inputs_test.go
```diff
@@ -0,0 +1,272 @@
...
+		inputs := createInputs{}
+		appName := "test-app"
+
+		dir, err := inputs.resolveDirectory(profile.WorkingDirectory, appName)
+
+		assert.Nil(t, err)
+		assert.Equal(t, path.Join(profile.WorkingDirectory, appName), dir)
+	})
+
+	t.Run("should return path of wd with directory appended when directory is set", func(t *testing.T) {
```
makesitgo, line 237: can we add a test case (and in this one, you may _need_ that `NewProfileFromTmpDir` where we have a directory that matches an existing file name and still return that path?

internal/commands/app/new_app_input_test.go
```diff
@@ -0,0 +1,110 @@
...
+		inputs         newAppInputs
+		procedure      func(c *expect.Console)
+		findAppErr     error
+		groupsErr      error
+		expectedFrom   from
+		expectedFilter realm.AppFilter
+		expectedErr    error
+	}{
+		{
+			description:    "should return the app id and group id of specified app when project and from are set",
```
makesitgo, line 42: let's reorganize this test file a bit... assuming we make the change to have two separate functions for resolving `Project` and resolving `From`, that'll be the first divide of test cases.  after that, within those groups lets have tests (or test tables if need be) for the ones that are interactive and need a pseudo terminal and then ones that aren't interactive and can roll with a `nil`/simple ui

internal/commands/app/create_test.go
```diff
@@ -0,0 +1,191 @@
...
+		assert.Nil(t, cmd.realmClient)
+		assert.Nil(t, cmd.atlasClient)
+		assert.Nil(t, cmd.Setup(profile, nil))
+		assert.NotNil(t, cmd.realmClient)
+		assert.NotNil(t, cmd.atlasClient)
+	})
+}
+
+func TestAppCreateHandler(t *testing.T) {
+	t.Run("should create bare bone project when no from type is specified", func(t *testing.T) {
```
makesitgo, line 37: [nit] switch "bare bone" with "a minimal"

internal/commands/app/create_test.go
```diff
@@ -0,0 +1,191 @@
...
+			GroupID: "test-project",
+			Name:    "test-app",
+			AppMeta: realm.AppMeta{
+				Location:        realm.LocationVirginia,
+				DeploymentModel: realm.DeploymentModelGlobal,
+			},
+		}, createdApp)
+	})
+
+	t.Run("should create a templated app when from is set", func(t *testing.T) {
```
makesitgo, line 88: avoid "templated" word here, as that will likely be an entirely different feature in the near future.  `"should create a new app with a structure based on the specified from app"` or something similar should suffice

internal/commands/app/create_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+		testApp := realm.App{
+			ID:          primitive.NewObjectID().Hex(),
+			GroupID:     primitive.NewObjectID().Hex(),
+			ClientAppID: "from-app-abcde",
+			Name:        "from-app",
+		}
+
+		var zipCloser *zip.ReadCloser
+		defer func() { zipCloser.Close() }()
```
makesitgo, line 100: instead of this, just have

```go
zipPkg, err := zip.OpenReader("testdata/project.zip")
assert.Nil(t, err)
```

and then in your mocked `ExportFn`, just `return ", &zipPkg.Reader, nil`

internal/commands/app/create_test.go
```diff
@@ -0,0 +1,191 @@
...
+				Project:         testApp.GroupID,
+				Location:        realm.LocationIreland,
+				DeploymentModel: realm.DeploymentModelGlobal,
+			}},
+			realmClient: client,
+		}
+
+		assert.Nil(t, cmd.Handler(profile, nil))
+
+		data, readErr := ioutil.ReadFile(filepath.Join(profile.WorkingDirectory, cmd.inputs.From, local.FileRealmConfig.String()))
```
makesitgo, line 137: could we use `local.LoadApp` here instead?  and the ensuing assert could just be `assert.Equal(t, local.AppRealmConfigJSON{...}, localApp.AppData)`

internal/commands/app/create_test.go
```diff
@@ -0,0 +1,191 @@
...
+		}}}, config)
+		assert.Equal(t, realm.App{
+			GroupID: testApp.GroupID,
+			Name:    testApp.Name,
+			AppMeta: realm.AppMeta{
+				Location:        realm.LocationIreland,
+				DeploymentModel: realm.DeploymentModelGlobal,
+			},
+		}, createdApp)
+	})
```
makesitgo, line 156: they could be argued as redundant, but can we add tests here for the client failure cases, just one for each client call we make to ensure we're still returning that error approrpriately?

makesitgo (commented):
alright done with the test review as well.  going to leave these as just comments for now, even though there's nothing blocking from my end.  lets just let the dust settle here and then it should be good to go!  nice work here

internal/commands/app/create.go
```diff
@@ -0,0 +1,199 @@
...
+	// if writeErr := local.WriteZip(zipPkgPath, zipPkg); writeErr != nil {
+	// 	return writeErr
+	// }
+
+	importErr := cmd.realmClient.Import(
+		newApp.GroupID,
+		newApp.ID,
+		loadedApp,
+	)
+	if importErr != nil {
```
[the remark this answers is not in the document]
nathantfrank: the length causes gofmt to make multiple lines so I'm gonna keep it that way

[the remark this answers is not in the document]
makesitgo: we dont have those rules applied to our codebase? we definitely are over line lengths everywhere 😬 

that's fine, but also don't be afraid of the multi-line, in-line if statement (if that makes any sense).  it at least reduces polluting the larger scope if/when possible

internal/commands/app/create.go
```diff
@@ -0,0 +1,199 @@
...
+		var projectIDErr error
+		projectID, projectIDErr = cmd.inputs.resolveProject(ui, cmd.atlasClient)
+		if projectIDErr != nil {
+			return projectIDErr
+		}
+	} else {
+		projectID = from.GroupID
+	}
+
+	appName, appNameErr := cmd.inputs.resolveAppName(ui, cmd.realmClient, from)
```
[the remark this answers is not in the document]
nathantfrank: I think it makes sense to keep them separate, just so we have smaller chunks for testing that we know behave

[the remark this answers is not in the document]
makesitgo: i figured that was the intention, no complaints from me here on that.  also ignore my "value receiver" comment here, i think that later comment around `resolveName` being mutative is more impactful than anything here

internal/commands/app/init_test.go
```diff
@@ -82,13 +83,16 @@ func TestAppInitHandler(t *testing.T) {
 		}
 
 		cmd := &CommandInit{
-			inputs:      initInputs{From: "test"},
+			inputs: initInputs{newAppInputs{
+				From:    "test",
+				Project: "test-project",
```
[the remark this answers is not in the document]
nathantfrank: Misunderstanding on my part, it is not needed

internal/commands/app/create_inputs.go
```diff
@@ -0,0 +1,128 @@
...
+		app, err := cli.ResolveApp(ui, client, realm.AppFilter{GroupID: f.GroupID, App: f.AppID})
+		if err != nil {
+			return "", err
+		}
+		return app.Name, nil
+	}
+	return i.Name, nil
+}
+
+func (i *createInputs) resolveProject(ui terminal.UI, client atlas.Client) (string, error) {
```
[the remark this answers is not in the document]
nathantfrank: Keeping it separate for testing

kpatel71716 (approved):
LGTM! Nice stuff

PR #65, opened 2021-02-09
REALMC-8112: Add environments to import request payload

andrewzlchen:
## Work Done

Adding support for environments in the old realm-cli while new realm-cli is under development

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -32,6 +32,7 @@ const (
 	rulesName            = "rules"
 	secretsName          = "secrets"
 	servicesName         = "services"
+	environmentsName     = "environments"
 	sourceName           = "source"
 	valuesName           = "values"
 	graphQLName          = "graphql"
@@ -222,8 +223,13 @@ func UnmarshalFromDir(path string) (map[string]interface{}, error) {
 	if err != nil {
 		return app, err
 	}
-
 	app[servicesName] = services
+
+	environments, err := unmarshalJSONFilesWithFilenames(filepath.Join(path, environmentsName))
+	if err != nil {
+		return app, err
+	}
+	app[environmentsName] = environments
 
 	return app, nil
 }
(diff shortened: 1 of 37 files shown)


utils/utils.go
```diff
@@ -252,6 +258,30 @@ func unmarshalJSONFiles(path string, ignoreDirErr bool) ([]interface{}, error) {
 	return files, nil
 }
 
+func unmarshalJSONFilesWithFilenames(path string) (map[string]interface{}, error) {
+	fileInfos, err := ioutil.ReadDir(path)
+	if err != nil {
+		return map[string]interface{}{}, err
```
makesitgo, line 264: could this just be `return nil, err` here with your changes to the unmarshaling logic on the backend?  (same question on the error handling for json.Unmarshal below)

utils/utils_test.go
```diff
@@ -63,5 +63,19 @@ func TestAppLoadFromDirectory(t *testing.T) {
 			u.So(t, svcMap["rules"], gc.ShouldNotBeEmpty)
 			u.So(t, svcMap["incoming_webhooks"], gc.ShouldNotBeEmpty)
 		}
+
+		u.So(t, app["environments"], gc.ShouldHaveLength, 5)
+		for _, env := range app["environments"].(map[string]interface{}) {
```
makesitgo, line 68: its probably best to stick to the `environments, ok := app["environments"].(map[string]interface{})` syntax here and have calls to `u.So(t, ok, gc.ShouldBeTrue)` (we seem a bit inconsistent on that in general in this file)

just if/when that fails, we'll at least get a clean test failure rather than a panic

makesitgo (approved):
looks good to me...

🆒  beans

utils/utils_test.go
andrewzlchen: makes sense. I'll do that for all other instances in this file.

PR #66, opened 2021-02-10
REALMC-7868: Add support for dependencies

makesitgo:
⚠️  A significant chunk of the line/file count here is due to adding a `node_modules` directory to the `local/testdata` package


internal/cloud/realm/client.go
```diff
@@ -35,6 +38,7 @@ type Client interface {
 	DeployDraft(groupID, appID, draftID string) (AppDeployment, error)
 	DiffDraft(groupID, appID, draftID string) (AppDraftDiff, error)
 	DiscardDraft(groupID, appID, draftID string) error
+	Deployments(groupID, appID string) ([]AppDeployment, error)
```
makesitgo, line 41: ⚠️ added this method solely for the purposes of testing

In order to wait for the `DEPLOY_DEPENDENCIES` job to complete, I needed to get all deployments to find the right `ID`

makesitgo:
I borked something here.... standby please

PR #67, opened 2021-02-10
REALMC-7868: Add support for dependencies

makesitgo:
I promise it's < 20 files here, but I included a `node_modules` directory in `local/testdata` to prove we can import from that format


internal/cloud/realm/client.go
```diff
@@ -35,6 +38,7 @@ type Client interface {
 	DeployDraft(groupID, appID, draftID string) (AppDeployment, error)
 	DiffDraft(groupID, appID, draftID string) (AppDraftDiff, error)
 	DiscardDraft(groupID, appID, draftID string) error
+	Deployments(groupID, appID string) ([]AppDeployment, error)
```
makesitgo, line 41: ⚠️   added this method solely for the purposes of testing

In order to wait for the `DEPLOY_DEPENDENCIES` job to complete, I needed to get all deployments to find the right id

internal/cloud/realm/dependencies_test.go
```diff
@@ -0,0 +1,96 @@
...
+	u.SkipUnlessRealmServerRunning(t)
+
+	client := newAuthClient(t)
+
+	groupID := u.CloudGroupID()
+
+	app, appErr := client.CreateApp(groupID, "importexport-test", realm.AppMeta{})
+	assert.Nil(t, appErr)
+
+	t.Run("Should successfully import a zip node_modules archive", func(t *testing.T) {
```
nathantfrank, line 26: [nit] lowercase test and the tests below

internal/cloud/realm/deployment_test.go
```diff
@@ -30,6 +30,12 @@ func TestRealmDeployments(t *testing.T) {
 		draft, draftErr := client.CreateDraft(groupID, app.ID)
 		assert.Nil(t, draftErr)
 
+		t.Run("Should initially find no deployments", func(t *testing.T) {
```
nathantfrank, line 33: [nit] lowercase test and tests below

internal/local/archive_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+)
+
+func TestArchiveReaderNew(t *testing.T) {
+	wd, wdErr := os.Getwd()
+	assert.Nil(t, wdErr)
+
+	testRoot := filepath.Join(wd, "testdata/dependencies")
+
+	t.Run("Should return an error when a project has an unsupported node_modules archive", func(t *testing.T) {
```
nathantfrank, line 17: [nit] lowercase test and tests below

internal/local/dependencies_test.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+)
+
+func TestDependenciesFind(t *testing.T) {
+	wd, wdErr := os.Getwd()
+	assert.Nil(t, wdErr)
+
+	testRoot := filepath.Join(wd, "testdata/dependencies")
+
+	t.Run("Should return an empty data when a project has no node_modules archive", func(t *testing.T) {
```
nathantfrank, line 21: [nit] lowercase test and test below

nathantfrank (commented):
Just nits from me

internal/cloud/realm/dependencies.go
```diff
@@ -0,0 +1,87 @@
...
+
+const (
+	dependenciesPathPattern        = appPathPattern + "/dependencies"
+	dependenciesArchivePathPattern = dependenciesPathPattern + "/archive"
+
+	paramFile = "file"
+)
+
+func (c *client) ImportDependencies(groupID, appID, uploadPath string) error {
+	file, fileErr := os.Open(uploadPath)
```
kpatel71716, line 24: [nit] Reminder to use `err` over `xxErr`here and throughout this function from our recent standardization

kpatel71716 (commented):
Looks pretty good to me! Just a reminder from my end

.evg.yml
```diff
@@ -16,6 +16,29 @@ functions:
           curl --silent -OL ${yarn_url}
           tar zvxf latest.tar.gz
           mv yarn-* yarn
+  "fetch_baas_artifacts": &fetch_baas_artifacts
+    - command: shell.exec
+      params:
+        shell: "bash"
+        script: |
+          git ls-remote git@github.com:10gen/baas.git | grep refs/heads/v20 | sort -k2 | awk '{ print "baas_revision: \""$1"\"" }' | tail -n 1 > baas.yml
```
tsedgwick, line 24: [nit] let's put a comment as to what we're doing here for future us

.evg.yml
```diff
@@ -261,11 +273,12 @@ buildvariants:
...
     atlas_server_url: "https://cloud-dev.mongodb.com"
     baas_supportlib_url: "https://s3.amazonaws.com/stitch-artifacts/stitch-support/stitch-support-rhel-70-4.3.2-721-ge791a2e-patch-5e2a6ad2a4cf473ae2e67b09.tgz"
-    go_url: "https://dl.google.com/go/go1.14.10.linux-amd64.tar.gz"
-    libmongo_url: "https://s3.amazonaws.com//stitch-artifacts/stitch-mongo-libs/stitch_mongo_libs_linux_64_patch_752d3ad5fa6c559e22ba1215acbe80b5c9567753_5fa1b8110305b929df995244_20_11_03_20_05_38/libmongo.so"
+    go_url: "https://dl.google.com/go/go1.14.13.linux-amd64.tar.gz"
+    libmongo_url: "https://s3.amazonaws.com//stitch-artifacts/stitch-mongo-libs/stitch_mongo_libs_linux_64_patch_9d94a997898b85fada11e619139c648e0b4be73f_5ff760db3627e064e1499e92_21_01_07_19_28_28/libmongo.so"
     mongodb_url: "https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-4.0.2.tgz"
     realm_server_url: "http://localhost:8080"
     yarn_url: "https://s3.amazonaws.com/stitch-artifacts/yarn/latest.tar.gz"
+    baas_revision: v20210210
```
tsedgwick, line 281: I don't believe we don't need this any longer

internal/cloud/realm/app.go
```diff
@@ -54,10 +54,10 @@ func (c *client) CreateApp(groupID, name string, meta AppMeta) (App, error) {
 	if resErr != nil {
 		return App{}, resErr
 	}
-	defer res.Body.Close()
 	if res.StatusCode != http.StatusCreated {
-		return App{}, parseResponseError(res)
+		return App{}, api.ErrUnexpectedStatusCode{"create app", res.StatusCode}
 	}
+	defer res.Body.Close()
```
tsedgwick, line 60: why did these defers move?

internal/cloud/realm/dependencies.go
```diff
@@ -0,0 +1,87 @@
...
+	if resErr != nil {
+		return resErr
+	}
+	if res.StatusCode != http.StatusNoContent {
+		return api.ErrUnexpectedStatusCode{"import dependencies", res.StatusCode}
+	}
+	return nil
+}
+
+func (c *client) ExportDependencies(groupID, appID string) (string, io.ReadCloser, error) {
```
tsedgwick, line 67: Can we get a comment as to why we are exporting this as an `io.ReadCloser` instead of a struct? I see we simply use that to write to a file, but let's call that out here since other client calls return a struct. Or - I wouldn't mind returning a struct with FileName and Contents.

internal/cloud/realm/dependencies.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+const (
+	dependenciesPathPattern        = appPathPattern + "/dependencies"
+	dependenciesArchivePathPattern = dependenciesPathPattern + "/archive"
+
+	paramFile = "file"
+)
+
+func (c *client) ImportDependencies(groupID, appID, uploadPath string) error {
```
tsedgwick, line 23: Is the linter not enforcing comments on exported funcs?

internal/local/archive.go
```diff
@@ -0,0 +1,207 @@
...
+		return newTarReader(file), nil
+	}
+
+	if ext == extTgz || strings.HasSuffix(path, extTarGz) {
+		return newGzipReader(file)
+	}
+
+	if ext == "" {
+		return newDirReader(path)
+	}
```
tsedgwick, line 55: ~[nit] this could just become a switch statement~ very disappointed we do this check `strings.HasSuffix(path, extTarGz)` which prevents us from doing a switch statement 😭

internal/local/archive.go
```diff
@@ -0,0 +1,207 @@
...
+	return nil, errUnknownArchiveExtension(path)
+}
+
+func errUnknownArchiveExtension(path string) error {
+	return fmt.Errorf(
+		"failed to read archive file at %s: unsupported format, use one of [%s] instead",
+		path,
+		strings.Join(supportedExts, ", "),
+	)
+}
```
tsedgwick, line 66: [nit] this seems unnecessary. Is the sole purpose of this func for tests?

tsedgwick (approved):
overall LGTM

internal/cloud/realm/app.go
makesitgo: with the new `api.ErrUnexpectedStatusCode`, we don't attempt to read from the `res.Body` until after we've determined we received the expected status code (previously, `parseResponseError` did read from `res.Body`)

the was brought about by handling automatic auth refresh in the client api calls.  we only return `res` from `client.do` when its status code is between 200 and 299.  so "parsing" the response for a server error was never really going to work in that scenario anyways (it wasn't a server error, just an unexpected 2XX code)

internal/cloud/realm/dependencies.go
makesitgo: the linter does enforce that.  ive noticed this in the past, but some exported funcs seem to not require a docstring/comment.  for instance ,`String` is definitely one.  i think for this case, there's something about implementing an interface method defined in the same package that you don't need the comment?

internal/local/archive.go
makesitgo: agreed with you on that one

internal/local/archive.go
makesitgo: the sole purpose of this func was for reducing the visual load when looking at the calling code... i can move it back, these things just tend to help me

internal/cloud/realm/dependencies.go
makesitgo: so this is likely just my inexperience with low-level i/o stuff, but we do it this way in the old cli so i presumed that was an optimization to avoid having to "read" the contents twice?

if you have a suggested route, i'm all ears.  i imagine its something to do with:

```go
// please ignore the name...
type MyExportedDependencies struct {
  filename string
  contents []byte // is this then populated with the return value of ioutil.ReadAll ???
}
```

makesitgo:
needs to be merged after https://github.com/10gen/baas/pull/3817

PR #69, opened 2021-02-17
REALMC-7164 Add secrets update to realm-cli 

jo-tse-mongo:


internal/commands/secrets/inputs.go
```diff
@@ -0,0 +1,107 @@
...
+	flagSecret            = "secret"
+	flagSecretShort       = "s"
+	updateFlagSecretUsage = "ID or name of the secret to update"
+
+	flagSecrets            = "secret"
+	flagSecretsShort       = "s"
+	deleteFlagSecretsUsage = "set the list of secrets to delete by ID or Name"
+)
+
+type promptFn func([]string) survey.Prompt
```
makesitgo, line 33: [question] were there any major differences with `secrets` compared to `users` that we couldn't follow a similar `inputs` pattern that was established there?  I don't think we need something like this `promptFn`...

lets work offline to see if im missing something here

internal/commands/secrets/update.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return err
+	}
+
+	toUpdate, err := cmd.inputs.resolveSecret(ui, secretList)
+	if err != nil {
+		return err
+	}
+
+	err = cmd.realmClient.UpdateSecret(app.GroupID, app.ID, toUpdate.ID, cmd.inputs.name, cmd.inputs.value)
```
makesitgo, line 48: since we're only updating a single secret here, we can just `return cmd.realmClient.UpdateSecret(app.GroupID, app.ID, toUpdate.ID, cmd.inputs.name, cmd.inputs.value)`

and in `Feedback`, you already have the details to print the success message (no need to worry about the error message, that is taken care of by returning the `err` here)

internal/commands/secrets/update.go
```diff
@@ -0,0 +1,58 @@
...
+	cmd.output = secretOutput{secret: toUpdate, err: err}
+	return nil
+}
+
+func (cmd *CommandUpdate) Feedback(profile *cli.Profile, ui terminal.UI) error {
+	if cmd.output.err != nil {
+		return ui.Print(terminal.NewTextLog("There was an issue updating the secret: %s", cmd.output.err.Error()))
+	}
+	return ui.Print(terminal.NewTextLog("Successfully updated the secret with the name %s", cmd.output.secret.Name))
```
makesitgo, line 57: [nit] `"Successfully updated secret" (doesn't need much more in the context of running that whole command)

makesitgo (commented):
just a couple of first thoughts, id like to revisit the abstractions around finding/selecting/prompting for secrets first

internal/commands/secrets/update.go
jo-tse-mongo: how do you get the `err` in the `Feedback` function without assigning it to the `cmd.output`? is this something i'll have to do after your refactor?

internal/commands/secrets/inputs.go
jo-tse-mongo: the main difference is that we can only update _one_ secret but we can delete _many_ secrets 
that changes the `survey.Prompt` that we need to use (MultiSelect, Select) as well as the message 
`promptFn` lets those individual components stay with the corresponding command files (`delete_inputs`, `update_inputs`) while letting `inputs` be a little more generic for any other application (returning one secret, returning many secrets)

internal/commands/secrets/update.go
makesitgo: errors returned from the handler are already handled in `command_factory` when we implement the Cobra command's `RunE` hook (see [here](https://github.com/10gen/realm-cli/blob/next/internal/cli/command_factory.go#L114))

the only reason we used `output` structs in the past was when we tried to do a batch of operations, so that even if an error occurred we would still attempt the rest of the specified requests

internal/commands/secrets/inputs.go
makesitgo: discussed two possible helpers here, but otherwise I'd push a lot of this logic just to the individual command input `Resolve` functions themselves

internal/commands/secrets/update_inputs_test.go
```diff
@@ -0,0 +1,83 @@
...
+		})
+	}
+
+	t.Run("should return an error if we cannot find the secret to update", func(t *testing.T) {
+		inputs := updateInputs{
+			secret: "cannotFind",
+		}
+
+		result, err := inputs.resolveSecret(nil, secrets)
+		assert.Equal(t, realm.Secret{}, result)
```
makesitgo, line 55: when asserting `error` return values, we don't need to assert anything on the first return value

and just to avoid any potential `nil` panics with `err.Error()`, try to use the `assert.Equal` with an actual error var (`errors.New("unable to find secret: cannotFind"` in this case)

internal/commands/secrets/update_inputs_test.go
```diff
@@ -0,0 +1,83 @@
...
+		_, console, _, ui, consoleErr := mock.NewVT10XConsole()
+		assert.Nil(t, consoleErr)
+		defer console.Close()
+
+		doneCh := make(chan struct{})
+		go func() {
+			defer close(doneCh)
+			console.ExpectString("Which secret would you like to update?")
+			console.SendLine("secretID2")
+			console.SendLine(" ")
```
makesitgo, line 71: I'd be surprised both `SendLine` calls are needed here, can you test around with this?  i think with `Select`, you just need to hit "enter" to confirm your selection (as opposed to the `MultiSelect`), so this 2nd line here may not be doing anything

internal/commands/secrets/update_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+	testLen := 7
+	secrets := make([]realm.Secret, testLen)
+	for i := 0; i < testLen; i++ {
+		secrets[i] = realm.Secret{
+			ID:   fmt.Sprintf("secretID%d", i),
+			Name: fmt.Sprintf("secretName%d", i),
+		}
+	}
+	successOutput := func(t *testing.T, output string) {
```
makesitgo, line 32: we should just use this `assert.Equal` where we make the assert in the test

i think with my comment about removing the `ui.Print` when we fail to update the secret, that last test case here goes away.  if it doesn't for some reason, let's at least break it out into its own test func, we should try to avoid mixing success/failure test cases in a single test table like this (a benefit of this is it helps simplify the anonymous `struct{}`s we use for test cases)

internal/commands/secrets/update_test.go
```diff
@@ -0,0 +1,173 @@
...
+	for _, tc := range []struct {
+		description string
+		testSecret  string
+		testName    string
+		testValue   string
+		expectedErr error
+		testOutput  func(*testing.T, string)
+	}{
+		{
+			description: "should return a successful message for a successful update for name and value",
```
makesitgo, line 44: these test cases are not really testing all that much stuff different from themselves... this whole test table can be just one "happy path" test func

the one thing to add would be to make sure we're capturing and asserting on the args that are passed into `realmClient.UpdateSecretFn`

internal/commands/secrets/update_test.go
```diff
@@ -0,0 +1,173 @@
...
+						FindAppsFn: func(filter realm.AppFilter) ([]realm.App, error) {
+							return []realm.App{app}, nil
+						},
+						SecretsFn: func(groupID, appID string) ([]realm.Secret, error) {
+							return secrets, nil
+						},
+					}
+				},
+				expectedErr: errors.New("unable to find the secret with the identifier illegal"),
+			},
```
makesitgo, line 161: the extra test case to add here would be for when `UpdateSecret` returns an error (effectively moved from above)

makesitgo (commented):
first batch of comments here, nothing major just a little cleanup.  nice work!

internal/commands/secrets/update_test.go
makesitgo: edit: after looking closer, i see what you were getting at with these tests, and i agree with you they deserve to exist as such (nice stuff!).  going to leave these in here, but hopefully in the future our commands actually do something with this knowledge (either require --name or --value, or print back a message indicating specifically what was performed ... out of scope for this ticket though)

makesitgo (approved):
🆒  beans

PR #70, opened 2021-02-18
 REALMC-8182: Remove lifecycle interfaces from command implementations

makesitgo:
Definitely a noisy one, hopefully the commits help explain a bit what's going on, but in short:

* removed the `Setup` interface (all it was doing was creating and setting clients on the command structs, these are now simply passed in to the command handler as args)
* removed the `Feedback` interface
  * this allows us to be more declarative/explicit about command execution
* `UI.Print` now does not return an error (but the error handling remains effectively the same as before this change)

### DIFF
--- a/internal/commands/push/command.go
+++ b/internal/commands/push/command.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/10gen/realm-cli/internal/cli"
-	"github.com/10gen/realm-cli/internal/cloud/atlas"
 	"github.com/10gen/realm-cli/internal/cloud/realm"
 	"github.com/10gen/realm-cli/internal/local"
 	"github.com/10gen/realm-cli/internal/terminal"
@@ -17,17 +16,7 @@ import (
 
 // Command is the `push` command
 type Command struct {
-	inputs      inputs
-	outputs     outputs
-	atlasClient atlas.Client
-	realmClient realm.Client
-}
-
-type outputs struct {
-	appCreated   bool
-	diffRejected bool
-	draftKept    bool
-	noDiffs      bool
+	inputs inputs
 }
 
 // Flags is the command flags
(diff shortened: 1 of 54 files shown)


kpatel71716 (approved):
LGTM [pending the failing test]! The changes make sense to me, I think the commands are more straightforward this way 🚀

internal/commands/push/command_test.go
```diff
@@ -417,79 +383,48 @@ diff2
...
+// 				cmd := &Command{inputs: tc.inputs}
+//
+// 				err := cmd.Handler(nil, ui, cli.Clients{})
+// 				assert.Nil(t, err)
+//
+// 				assert.Equal(t, tc.expectedContents, out.String())
+// 			})
+// 		}
+// 	})
+// }
```
nathantfrank, line 427: [Q] Is this meant to still be here?

nathantfrank (approved):
Pretty much LGTM, nothing stood out that I could tell!

internal/commands/push/command_test.go
makesitgo: Nope, thanks for catching this.  Tests for this are also now more flushed out in the the future PR for spinner support

internal/commands/push/command.go
```diff
@@ -164,47 +152,24 @@ func (cmd *Command) Handler(profile *cli.Profile, ui terminal.UI) error {
...
-
-	return ui.Print(terminal.NewTextLog("Successfully pushed app changes"))
-}
-
 type namer interface{ Name() string }
 type locationer interface{ Location() realm.Location }
 type deploymentModeler interface{ DeploymentModel() realm.DeploymentModel }
 
-func (cmd *Command) createNewApp(ui terminal.UI, groupID string, appData interface{}) (realm.App, error) {
+func createNewApp(ui terminal.UI, realmClient realm.Client, appDirectory, groupID string, appData interface{}) (realm.App, bool, error) {
```
tsedgwick, line 168: I would rather throw a named error if the ui doesn't wish to proceed than add another return value of `bool`. How are we planning on handling the case where the user doesn't want the confirm prompt (ie: from a ci pipeline)?

tsedgwick (approved):
one nit from me, otherwise looks good

internal/commands/push/command.go
makesitgo: to bypass confirm prompts, the CLI (and this is existing behavior) exposes a `--yes, -y` flag (which is checked during `ui.Confirm(...)`, but is also exposed through `ui.AutoConfirm()`).

there's a smell here, that we need to pay attention to that detail two different ways, so i'm all ears if you have any thoughts.  but this behavior is something i more or less grabbed from the existing cli

PR #74, opened 2021-02-22
REALMC-8135: Implement data-source flag for app create command

nathantfrank:


internal/cloud/realm/app.go
```diff
@@ -109,7 +109,7 @@ func (c *client) FindApps(filter AppFilter) ([]App, error) {
 
 	var filtered = make([]App, 0, len(apps))
 	for _, app := range apps {
-		if strings.HasPrefix(app.ClientAppID, filter.App) {
+		if strings.HasPrefix(app.ClientAppID, strings.ToLower(filter.App)) {
```
nathantfrank, line 112: Noticed the ClientAppID would be all lowercased which would cause name not to be matched

internal/cloud/realm/command_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	t.Run("with an active session", func(t *testing.T) {
+		client := newAuthClient(t)
+		groupID := u.CloudGroupID()
+
+		t.Run("should list clusters", func(t *testing.T) {
+			app, err := client.CreateApp(groupID, "commands-test", realm.AppMeta{})
+			assert.Nil(t, err)
+
+			_, err = client.ListClusters(groupID, app.ID)
```
nathantfrank, line 31: [Q] Is this okay? There could be any number of available clusters or none.

internal/commands/app/create_test.go
```diff
@@ -12,7 +12,6 @@ import (
 	"github.com/10gen/realm-cli/internal/local"
 	"github.com/10gen/realm-cli/internal/utils/test/assert"
 	"github.com/10gen/realm-cli/internal/utils/test/mock"
-
```
nathantfrank, line 15: Will be adding this line back in next commit

nathantfrank (commented):
First Pass

nathantfrank:
@makesitgo Do we want to have an additional line of output about the datasource if all went well?

internal/cloud/realm/command_test.go
makesitgo: if im understanding this correctly, we can probably tighten this up via env vars and recommending devs to setup their own test projects to mirror the setup we have for the test user here (similar to how we leverage `u.CloudGroupName()`)

makesitgo:
Yes on the additional line of output...

question: have you tested this out locally?  is the api call at all lengthy?  I ask b/c we (will soon) support spinners, and Im wondering if this process warrants one.

either way, you could have a `01:23:45 UTC INFO  Linked data source` or some type of message similar.

if a spinner's invovlved, the way it works so far is we have a temporary message like:

`:spinner: Linking data source...`

which would then disappear and be replaced with a log message instead:

`01:23:45 UTC INFO  Linked data source`

Happy to elaborate more or demo it locally (you can check it out as well [here](https://github.com/makesitgo/realm-cli/tree/REALMC-7867) -- just no promises that branch is stable and won't see any force pushes)

internal/commands/app/create.go
```diff
@@ -109,56 +112,60 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI) error {
...
+		}
+		serviceName := newApp.Name + "_cluster"
+		serviceConfig := map[string]interface{}{
+			"name": serviceName,
+			"type": "mongodb-atlas",
+			"config": map[string]interface{}{
+				"clusterName":         dsCluster,
+				"readPreference":      "primary",
+				"wireProtocolEnabled": false,
+			},
```
kpatel71716, line 138: [major q] Should we consider data lakes here as well? Data Lakes, support for which is relatively new (mid-late 2020), and Clusters are the two big types of data sources we have. I believe the service configuration is different as well for data lakes so that is something to keep in mind. Though I haven't looked through the master branch yet to see if we added support for data lakes in the cli last year

nathantfrank:
> Yes on the additional line of output...
> 
> question: have you tested this out locally? is the api call at all lengthy? I ask b/c we (will soon) support spinners, and Im wondering if this process warrants one.
> 
> either way, you could have a `01:23:45 UTC INFO Linked data source` or some type of message similar.
> 
> if a spinner's invovlved, the way it works so far is we have a temporary message like:
> 
> `:spinner: Linking data source...`
> 
> which would then disappear and be replaced with a log message instead:
> 
> `01:23:45 UTC INFO Linked data source`
> 
> Happy to elaborate more or demo it locally (you can check it out as well [here](https://github.com/makesitgo/realm-cli/tree/REALMC-7867) -- just no promises that branch is stable and won't see any force pushes)

From my testing against dev there is not a significant time difference (> ~1 sec) when linking a data source vs just app creation.

internal/commands/commands.go
```diff
@@ -74,7 +74,20 @@ Realm app will be exported as well.`,
...
-			{
+			cli.CommandDefinition{
+				Use:         "create",
+				Display:     "app create",
+				Description: "Create a remote Realm app from your current working directory",
+				Help: `Creates a new Realm app in the cloud, and saves your configuration files on 
+a new disk. This command will create a new directory for your project. 
+To initialize a Realm app in your current working directory, use ‘app init’.
+
+You can specify a ‘From’ input to create a Realm app from either a template,
```
makesitgo, line 85: i know this is from the doc, but we should tweak this a little to avoid mentioning `templates` (as thats not supported/implemented yet).  I also think using `'--from'` instead of `'From'` makes more sense in this context, thoughts?

internal/cloud/realm/command_test.go
```diff
@@ -0,0 +1,36 @@
...
+		_, err := client.ListClusters(primitive.NewObjectID().Hex(), primitive.NewObjectID().Hex())
+		assert.Equal(t, realm.ErrInvalidSession{}, err)
+	})
+
+	t.Run("with an active session", func(t *testing.T) {
+		client := newAuthClient(t)
+		groupID := u.CloudGroupID()
+
+		t.Run("should list clusters", func(t *testing.T) {
+			app, err := client.CreateApp(groupID, "commands-test", realm.AppMeta{})
```
makesitgo, line 28: we should use the `setupApp` helper here and `defer teardown()` with it

internal/commands/app/create.go
```diff
@@ -101,70 +104,66 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
-		},
-		{
-			"Info":    "Realm UI",
-			"Details": fmt.Sprintf("%s/groups/%s/apps/%s/dashboard", profile.RealmBaseURL(), newApp.GroupID, newApp.ID),
-		},
-		{
-			"Info":    "Check out your app",
-			"Details": fmt.Sprintf("cd ./%s && realm-cli app describe", newApp.Name),
-		},
+	rows := make([]map[string]interface{}, 0, rowCount)
```
makesitgo, line 158: you can just let this be `make([]map[string]interface{}, 0, 4)`

internal/commands/app/create.go
```diff
@@ -101,70 +104,66 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
+			return err
+		}
+		rowCount++
+	}
 
 	if err := clients.Realm.Import(
 		newApp.GroupID,
 		newApp.ID,
-		loadedApp,
+		loadedApp.AppData,
```
makesitgo, line 152: was this change needed?

internal/commands/app/create_inputs.go
```diff
@@ -18,10 +18,9 @@ var (
 	flagDirectoryShort = "c"
 	flagDirectoryUsage = "the directory to create your new Realm app, defaults to Realm app name"
 
-	// TODO(REALMC-8135): Implement data-source flag for app create command
-	// flagDataSource      = "data-source"
-	// flagDataSourceShort = "s"
-	// flagDataSourceUsage = "atlas cluster to back your Realm app, defaults to first available"
+	flagDataSource      = "data-source"
+	flagDataSourceShort = "s"
+	flagDataSourceUsage = "atlas cluster to back your Realm app, defaults to first available"
```
makesitgo, line 23: `include to link an Atlas cluster to your Realm app, defaults to first available`

internal/commands/app/create_inputs.go
```diff
@@ -88,29 +85,28 @@ func (i *createInputs) resolveDirectory(wd string) (string, error) {
 		return "", appErr
 	}
 	if appOK {
-		return "", fmt.Errorf("%s is inside or is a Realm app directory", dir)
+		return "", fmt.Errorf("%s is inside or is a Realm app directory", fullPath)
```
makesitgo, line 88: ideally we use some form of `errProjectExists` here, so that its consistent b/w `app init` and `app create`

internal/commands/app/create.go
```diff
@@ -101,70 +104,66 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
-	// if idx := strings.LastIndex(zipPkgName, "_"); idx != -1 {
-	// 	zipPkgName = zipPkgName[:idx]
-	// }
-	// zipPkgPath := filepath.Join(dir, zipPkgName) // Is this the correct path
-	// if writeErr := local.WriteZip(zipPkgPath, zipPkg); writeErr != nil {
-	// 	return writeErr
-	// }
+	rowCount := 3
+	if cmd.inputs.DataSource != "" {
+		dsCluster, err := cmd.inputs.resolveDataSource(clients.Realm, groupID, newApp.ID)
```
makesitgo, line 109: if we keep this function name, id rename this `var` to `dataSource`.  otherwise I'd consider renaming this to `resolveClusterName` to be more specific about what its doing (for if and when data lakes become a thing)

internal/commands/app/create_inputs.go
```diff
@@ -88,29 +85,28 @@ func (i *createInputs) resolveDirectory(wd string) (string, error) {
...
-// 		return "", fmt.Errorf("Unable to find any available cluster for Group ID %s", groupID)
-// 	}
-// 	return dsCluster, nil
-// }
+func (i *createInputs) resolveDataSource(client realm.Client, groupID, appID string) (string, error) {
+	clusters, err := client.ListClusters(groupID, appID)
+	if err != nil {
+		return "", err
+	}
+	var dsCluster string
```
makesitgo, line 98: [nit] rename this to `clusterName`

internal/commands/app/create_inputs.go
```diff
@@ -88,29 +85,28 @@ func (i *createInputs) resolveDirectory(wd string) (string, error) {
...
+		if (i.DataSource == "" && cluster.State == "IDLE") || i.DataSource == cluster.Name {
+			dsCluster = cluster.Name
+			break
+		}
+	}
+	if dsCluster == "" {
+		if i.DataSource != "" {
+			return "", fmt.Errorf("unable to find the %s cluster", i.DataSource)
+		}
+		return "", fmt.Errorf("unable to find any available cluster for Project %s", groupID)
```
makesitgo, line 109: thoughts on simplifying this and just returning `errors.New("failed to find Atlas cluster")` here?  in the context of running this command, the user probably already know the name they specified

internal/commands/app/create.go
```diff
@@ -101,70 +104,66 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
+			"config": map[string]interface{}{
+				"clusterName":         dsCluster,
+				"readPreference":      "primary",
+				"wireProtocolEnabled": false,
+			},
+		}
+		var dsPath string
+		switch loadedApp.ConfigVersion() {
+		case realm.AppConfigVersion20210101:
+			dsPath = fmt.Sprintf("data_sources/%s/config.json", serviceName)
```
makesitgo, line 126: use `filepath.Join` for these types of things:

`filepath.Join(local.NameDataSources, serviceName, local.FileConfig.String()`

(similarly for the below case as well)

internal/commands/app/create.go
```diff
@@ -101,70 +104,66 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
+		serviceConfig := map[string]interface{}{
+			"name": serviceName,
+			"type": "mongodb-atlas",
+			"config": map[string]interface{}{
+				"clusterName":         dsCluster,
+				"readPreference":      "primary",
+				"wireProtocolEnabled": false,
+			},
+		}
+		var dsPath string
```
makesitgo, line 123: [meganit] `path` or `serviceConfigPath` here

internal/commands/app/create.go
```diff
@@ -101,70 +104,66 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
+		data, err := json.MarshalIndent(serviceConfig, local.ExportedJSONPrefix, local.ExportedJSONIndent)
+		if err != nil {
+			return err
+		}
+		err = local.WriteFile(filepath.Join(dir, dsPath), 0666, bytes.NewReader(data))
+		if err != nil {
+			return err
+		}
+		err = loadedApp.Load()
+		if err != nil {
```
makesitgo, line 143: at least in-line this `err` (i'd do it above too, even if it means splitting the args onto new lines, but your call there)

internal/commands/app/create.go
```diff
@@ -101,70 +104,66 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
+	rows := make([]map[string]interface{}, 0, rowCount)
+	rows = append(rows, map[string]interface{}{"Info": "Client App ID", "Details": newApp.ClientAppID})
+	rows = append(rows, map[string]interface{}{"Info": "Realm Directory", "Details": dir})
+	rows = append(rows, map[string]interface{}{"Info": "Realm UI", "Details": fmt.Sprintf("%s/groups/%s/apps/%s/dashboard", profile.RealmBaseURL(), newApp.GroupID, newApp.ID)})
+	if cmd.inputs.DataSource != "" {
+		rows = append(rows, map[string]interface{}{"Info": "Data Source", "Details": cmd.inputs.DataSource})
 	}
 
 	ui.Print(terminal.NewTableLog("Successfully created app", headers, rows...))
+	ui.Print(terminal.NewDebugLog("Check out your app: cd ./%s && realm-cli app describe", cmd.inputs.Directory))
```
makesitgo, line 167: lets create `terminal.NewFollowupLog` which takes in `message string` and `items ...interface{}`, and will otherwise replace what we have for `NewReferenceLinksLog`/`NewSuggestedCommandsLog`.  You could then also use that here

internal/commands/app/create_inputs.go
```diff
@@ -88,29 +85,28 @@ func (i *createInputs) resolveDirectory(wd string) (string, error) {
...
-// 	}
-// 	if dsCluster == "" {
-// 		if i.DataSource != "" {
-// 			return "", fmt.Errorf("Unable to find the %s cluster", i.DataSource)
-// 		}
-// 		return "", fmt.Errorf("Unable to find any available cluster for Group ID %s", groupID)
-// 	}
-// 	return dsCluster, nil
-// }
+func (i *createInputs) resolveDataSource(client realm.Client, groupID, appID string) (string, error) {
```
makesitgo, line 93: do we have tests for this here?  probably somewhat implicitly through the `Handler` tests, but can we write a few unit tests for this to make sure we're returning the errors we expect, and also matching on the `"IDLE"`/`cluster.Name` logic?

makesitgo (commented):
lookin real good so far... nothing major just a bunch of questions/nits

internal/commands/commands.go
nathantfrank: agree

internal/commands/app/create_inputs.go
nathantfrank: I agree, though I do like the difference in error between failing to find one they specified vs any available cluster

internal/commands/app/create.go
nathantfrank: Not required, but it does read a little more cleanly in the context of the `Import` parameter `appData`

makesitgo: talked offline, but to close the loop here, this was need (and the source of our Diff/Import bug I was seeing earlier today)

internal/commands/app/create_inputs.go
makesitgo: to close the loop here as well, this scenario does not exist with the current implementation.

should we ever support these bool/string hybrid-type flags, we can revisit this

internal/commands/app/create.go
makesitgo: this flag is new functionality, so there was no prior support for this.  this is a great point though... lets tackle that in a separate jira though

internal/commands/app/create_inputs.go
```diff
@@ -83,34 +81,39 @@ func (i *createInputs) resolveDirectory(wd string) (string, error) {
...
-// 	}
-// 	if dsCluster == "" {
-// 		if i.DataSource != "" {
-// 			return "", fmt.Errorf("Unable to find the %s cluster", i.DataSource)
-// 		}
-// 		return "", fmt.Errorf("Unable to find any available cluster for Group ID %s", groupID)
-// 	}
-// 	return dsCluster, nil
-// }
+func (i *createInputs) resolveDataSource(client realm.Client, groupID, appID string) (map[string]interface{}, error) {
```
makesitgo, line 94: instead of returning `map[string]interface{}` here, let's just a struct for this (so we can avoid the `["name"]` accessor and use `.Name` instead)

just keep it privately declared here for now (`type dataSource struct { ... }`).  its tempting to say put this in `local` and we can use it as part of the `ServiceStructure`, but I still want to try and keep that `local` structure stuff as generic as possible

internal/commands/app/create.go
```diff
@@ -101,70 +103,55 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
-	// 	return writeErr
-	// }
-
-	if err := clients.Realm.Import(
+	if cmd.inputs.DataSource != "" {
+		dataSource, err := cmd.inputs.resolveDataSource(clients.Realm, groupID, newApp.ID)
+		if err != nil {
+			return err
+		}
+		dataSourceName := fmt.Sprintf("%v", dataSource["name"])
```
makesitgo, line 111: can get rid of this once we convert to a `struct` here

internal/commands/app/create.go
```diff
@@ -101,70 +103,55 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
+		var path string
+		switch loadedApp.ConfigVersion() {
+		case realm.AppConfigVersion20210101:
```
makesitgo, line 114: let's make this the default case/behavior and the potential error here

internal/commands/app/errors.go
```diff
@@ -1,7 +1,14 @@
 package app
 
-type errProjectExists struct{}
+type errProjectExists struct {
+	details string
```
makesitgo, line 4: any objections to making this `path string`, and then using
```go
var suffix string
if err.path != "" {
  suffix = " at "+err.path
}
return "a project already exists" + suffix
```
for the error string?

internal/commands/app/create_inputs.go
```diff
@@ -83,34 +81,39 @@ func (i *createInputs) resolveDirectory(wd string) (string, error) {
...
-	_, appOK, appErr := local.FindApp(fullPath)
-	if appErr != nil {
-		return "", appErr
+	_, appOK, err := local.FindApp(fullPath)
+	if err != nil {
+		return "", err
 	}
 	if appOK {
-		return "", fmt.Errorf("%s is inside or is a Realm app directory", dir)
+		return "", errProjectExists{details: fmt.Sprintf("%s is inside or is a Realm app directory", fullPath)}
```
makesitgo, line 89: if we make the switch to `path string` for this error struct, this can be just `errProjectExists{"./"+i.Directory}`

internal/commands/app/create_inputs_test.go
```diff
@@ -193,6 +193,69 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
 		dir, err := inputs.resolveDirectory(profile.WorkingDirectory)
 
 		assert.Equal(t, "", dir)
-		assert.Equal(t, fmt.Errorf("%s is inside or is a Realm app directory", specifiedDir), err)
+		assert.Equal(t, errProjectExists{details: fmt.Sprintf("%s is inside or is a Realm app directory", fullDir)}, err)
+	})
+}
+
+func TestAppCreateInputsResolveDataSource(t *testing.T) {
+	t.Run("should return data source config of an provided cluster", func(t *testing.T) {
```
makesitgo, line 201:
```suggestion
	t.Run("should return data source config of a provided cluster", func(t *testing.T) {
```

makesitgo (approved):
minor things left, but this is good to go

🆒  beans

PR #75, opened 2021-02-23
REALMC-7171: support app diff

tsedgwick:
I wanted to get the barebones up here so we can discuss the remaining items. I'll work on adding tests in parallel to figuring out if this is all that is needed for diff.


internal/commands/app/diff.go
```diff
@@ -0,0 +1,138 @@
...
+	if len(diffs) == 0 {
+		// there are no diffs
+		cmd.outputs.noDiffs = true
+		return nil
+	}
+
+	err = ui.Print(terminal.NewTextLog(
+		"The following reflects the proposed changes to your Realm app\n%s",
+		strings.Join(diffs, "\n"),
+	))
```
tsedgwick, line 123: from the spec, it only appears we output to stdout. This command previously gave options to write to file, etc. Do we want to introduce more input flags and support that?

makesitgo: support for that is built-in via `factory.SetGlobalFlags(cmd.PersistentFlags())` (found in `cmd/root.go`)

you have it right here with using `ui.Print` and `NewTextLog`

internal/commands/app/diff.go
```diff
@@ -0,0 +1,138 @@
...
+import (
+	"errors"
+	"strings"
+
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/cloud/atlas"
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	"github.com/10gen/realm-cli/internal/local"
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/AlecAivazis/survey/v2"
```
makesitgo, line 12: [nit] move this down

internal/commands/app/diff.go
```diff
@@ -0,0 +1,138 @@
...
+
+	"github.com/spf13/pflag"
+)
+
+type diffOutputs struct {
+	noDiffs bool
```
makesitgo, line 18: if/when i get the refactor in, this should be able to go away and we can print `"Deployed app is identical to proposed version"` where you otherwise return early

internal/commands/app/diff.go
```diff
@@ -0,0 +1,138 @@
...
+	IncludeDependencies bool
+	IncludeHosting      bool
+	cli.ProjectInputs
+}
+
+// CommandDiff is the `app diff` command
+type CommandDiff struct {
+	inputs      diffInputs
+	outputs     diffOutputs
+	atlasClient atlas.Client
```
makesitgo, line 32: this can go away

internal/commands/app/diff.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// Inputs is the command inputs
+func (cmd *CommandDiff) Inputs() cli.InputResolver {
+	return &cmd.inputs
+}
+
+// Setup is the command setup
+func (cmd *CommandDiff) Setup(profile *cli.Profile, ui terminal.UI) error {
+	cmd.realmClient = profile.RealmAuthClient()
```
makesitgo, line 52: another fallout of the rebase off my refactor: this would go away and you can just use `clients.Realm` which will be passed into the handler func

internal/commands/app/diff.go
```diff
@@ -0,0 +1,138 @@
...
+	return nil
+}
+
+// Handler is the command handler
+func (cmd *CommandDiff) Handler(profile *cli.Profile, ui terminal.UI) error {
```
makesitgo, line 57: just making a note for post-rebase: this signature will want to be updated to add the `clients cli.Clients` arg

internal/commands/app/diff.go
```diff
@@ -0,0 +1,138 @@
...
+	app, err := local.LoadApp(cmd.inputs.AppDirectory)
+	if err != nil {
+		return err
+	}
+
+	apps, err := cmd.realmClient.FindApps(realm.AppFilter{
```
makesitgo, line 63: this chunk of logic for selecting an app _should_ be all covered in `cli.ResolveApp(ui, clients.Realm, cmd.inputs.Filter()` (it handles fetching the apps, errors if none are found, returns the single result if present, or otherwise presents the user with a prompt to pick the app they are targeting)

(and `cli.ProjectInputs` has a function defined on it to provide the `realm.AppFilter`)

internal/commands/app/diff.go
```diff
@@ -0,0 +1,138 @@
...
+	return nil
+}
+
+// Feedback is the command feedback
+func (cmd *CommandDiff) Feedback(profile *cli.Profile, ui terminal.UI) error {
+	if cmd.outputs.noDiffs {
+		return ui.Print(terminal.NewTextLog("Deployed app is identical to proposed version"))
+	}
+
+	return ui.Print(terminal.NewTextLog("diff app successful"))
```
makesitgo, line 137: was this part of the old command as well?  I'm thinking we might be ok without a print here, and just leave the command output to be the diff or "no changes" (or an error)

internal/commands/commands.go
```diff
@@ -66,22 +66,30 @@ Realm app will be exported as well.`,
...
 				Description: "List the MongoDB Realm applications associated with the current user",
 				Help:        "list help", // TODO(REALMC-7429): add help text description
 				Command:     &app.CommandList{},
 			},
+			{
+				Use:         "diff",
+				Aliases:     []string{},
+				Display:     "app diff",
+				Description: "Show changes between a remote Realm application and the local directory associated with the current user",
+				Help:        "diff help", // TODO(REALMC-7429): add help text description
```
makesitgo, line 90: Check out [this doc](https://docs.google.com/document/d/14fBZaN47HRe3-2Pk1uSfYcCxX3Dd95wK0tprS3WE1vk/edit#heading=h.clf8mfqw7ws8) that Kevin from design put together:

**Description:** `Show differences between your Realm app and your local directory`

**Help:** `Displays file-by-file differences between the latest version of your Realm app and your local directory. If you have more than one Realm app, you will be prompted to select a Realm app that you would like to display from a list of all Realm apps associated with your user profile.`

fwiw, i still need to go in to these help texts and clean it up/format them.  if you have any thoughts on how to do it well i'm all ears.  but the main crux of the issue here is we need to handle the line-width formatting ourselves for this long-form help text

visually speaking, it'd be pretty convenient to use go's backticks with multi-lines for this, not sure if there's a clever programmatic way about this as well

internal/commands/app/diff_inputs.go
```diff
@@ -0,0 +1,19 @@
...
+	flagIncludeDependencies      = "include-dependencies"
+	flagIncludeDependenciesShort = "d"
+	flagIncludeDependenciesUsage = "include to diff Realm app dependencies changes as well"
+	flagIncludeHosting           = "include-hosting"
+	flagIncludeHostingShort      = "s"
+	flagIncludeHostingUsage      = "include to diff Realm app hosting changes as well"
+)
+
+func (i *diffInputs) Resolve(profile *cli.Profile, ui terminal.UI) error {
+	return nil
```
makesitgo, line 18: probably want to have an `if i.AppDirectory == ""` check here, and if so, set it to `profile.WorkingDirectory`

makesitgo (commented):
thanks for turning this around, just a couple of first thoughts here while you work on tests

It doesn't seem like we used to support the `--include-dependencies` flag for this diff command, and that was probably due to the fact our "diff" for it was mostly just a single line saying "Import dependencies").

ideally, id like to have a future ticket where we actually did that diff.  we have the server endpoints that support it, so in theory its just another api call.  and then our diffs of dependencies would match the granularity of our diffs for hosting

internal/commands/app/diff.go
```diff
@@ -0,0 +1,104 @@
...
+		return err
+	}
+
+	if cmd.inputs.IncludeHosting {
+		// TODO(REALMC-7177): diff hosting changes
+		diffs = append(diffs, "Diff hosting")
+	}
+
+	if cmd.inputs.IncludeDependencies {
+		// TODO(REALMC-8242): diff hosting changes
```
makesitgo, line 83: can you also add this comment to the relevant place in `commands/push/command.go`?

internal/commands/app/diff_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/10gen/realm-cli/internal/utils/test/mock"
+	"go.mongodb.org/mongo-driver/bson/primitive"
```
makesitgo, line 11: [nit] newline here

internal/commands/app/diff_test.go
```diff
@@ -0,0 +1,87 @@
...
+		description        string
+		inputs             diffInputs
+		expectedAppFilter  realm.AppFilter
+		expectedDiff       []string
+		expectedDiffOutput string
+		expectedErr        error
+	}{
+		{
+			description:        "with no project nor app flag set should diff based on input",
+			expectedAppFilter:  realm.AppFilter{},
```
makesitgo, line 35: [stylenit] could remove this (and `expectedDiff: nil` below)

internal/commands/app/diff_test.go
```diff
@@ -0,0 +1,87 @@
...
+	} {
+		t.Run(tc.description, func(t *testing.T) {
+			out, ui := mock.NewUI()
+
+			realmClient := mock.RealmClient{}
+
+			var appFilter realm.AppFilter
+			realmClient.FindAppsFn = func(filter realm.AppFilter) ([]realm.App, error) {
+				appFilter = filter
+				return apps, nil
```
makesitgo, line 75: we've been writing tests for the commands where this call fails as well (and proving the error is returned).  typically we've had client failure tests bunched together and then handler success tests bunched together.

organize as you see fit, but i think it'd be good to get the `FindApps` error as a test case in here somewhere, just so this particular command has that coverage too

makesitgo (approved):
looks good to me, if you want to merge this in before my hosting changes then i'll be glad to pick up that `TODO(REALMC-7177)` from here

🆒  beans

PR #76, opened 2021-02-23
REALMC-7170 build out app delete command

kpatel71716:


internal/cli/inputs.go
```diff
@@ -116,6 +116,41 @@ func ResolveApp(ui terminal.UI, client realm.Client, filter realm.AppFilter) (re
 	return appsByOption[selection], nil
 }
 
+// ResolveApps will use the provided Realm client to resolve multiple apps specified by the filter
+func ResolveApps(ui terminal.UI, client realm.Client, filter realm.AppFilter) ([]realm.App, error) {
```
kpatel71716, line 120: Thoughts on a separate function here? I felt that the different return types/slightly different survey made this fine to just duplicate and repurpose

internal/commands/app/delete.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	"github.com/10gen/realm-cli/internal/terminal"
+
+	"github.com/spf13/pflag"
+)
+
+// CommandDelete is the `app delete` command
+type CommandDelete struct {
+	inputs      cli.ProjectInputs
```
kpatel71716, line 13: We can delete multiple apps with the survey multi-select, but only accept one value for the --app flag. I _think_ this is what the spec implied, but I am open to different interpretation here

internal/commands/app/delete.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Feedback is the command feedback
+func (cmd *CommandDelete) Feedback(profile *cli.Profile, ui terminal.UI) error {
+	if len(cmd.outputs) == 0 {
+		return ui.Print(terminal.NewTextLog("No apps to delete"))
+	}
+
+	var logs []terminal.Log
+	logs = append(logs, terminal.NewTableLog(
+		"Deleted app(s)",
```
kpatel71716, line 51: I originally wanted to have text indicating the number of apps we deleted, but without going through the output and  counting how many errors we got, we can't put that number here. I am fine leaving as is or storing a count in a variable (which is simple once this is all done in Handler)

internal/commands/app/delete_test.go
```diff
@@ -0,0 +1,140 @@
...
+			},
+			expectedOutput: strings.Join(
+				[]string{
+					"01:23:45 UTC INFO  Deleted app(s)",
+					"  ID                        Name  Deleted  Details     ",
+					"  ------------------------  ----  -------  ------------",
+					"  60344735b37e3733de2adf40  app1  true                 ",
+					"  60344735b37e3733de2adf41  app2  false    client error",
+					"  60344735b37e3733de2adf42  app3  true                 ",
+					"",
```
kpatel71716, line 124: If this looks good, I was going to make a super quick drive-by fix to change app list to have a table with ID and Name

internal/commands/app/delete.go
nathantfrank: I'm of the opinion if we support a procedure interactively, we should probably support it in the single command if possible. That being said I think the pattern should be `deleteInputs` with a `StringSlice`.

nathantfrank (changes requested):
Besides the rebase that's needed, I think you are almost there.

internal/cli/inputs.go
makesitgo: do we really use `ResolveApps` anywhere else besides `app delete` though?  this should probably just live in the `deleteInputs` as regular old code

and on your other question about `--app` being a string here, that's probably indication enough that we don't want to use `cli.ProjectInputs` either

internal/commands/app/delete.go
makesitgo: see my comment about `ResolveApps` ... but it just doesn't seem like trying to use any of the `cli` helpers here is appropriate

internal/commands/app/delete.go
makesitgo: first off, you should rebase and repurpose this `Feedback` function into the `Handler` itself

then, while you're looping through and deleting apps, you could keep an incrementing count of non-nil errors there

---

fwiw, i am starting to feel like this table output too verbose/heavy-handed, but i also don't want to go in and make cli-wide changes to that yet.  i say this here just so you dont go crazy over-architecting a feedback message here when we may be simplifying it all in the first place

in the context of someone running `realm-cli apps delete --app a1 --app a2 ...`, i think the following output would be sufficiently clear:

```
Successfully deleted 2 apps
```
or
```
Successfully deleted 1 / 2 apps
Failed to delete app: a2
```
or
```
Successfully deleted 0 / 2 apps
Failed to delete apps:
  - a1
  - a2
```

to be clear, we should not be implementing this in this ticket, just leave this consistent with our other delete methods for now.  but avoid making changes to any other commands (like `app list` here for now)

internal/commands/app/delete_test.go
makesitgo: this is fine here, but lets keep `app list` as is for now (see my comment above)

internal/commands/app/delete.go
kpatel71716: Done, refactored this part significantly

internal/cli/inputs.go
kpatel71716: Refactored and added this onto deleteInputs

internal/commands/app/delete_inputs_test.go
```diff
@@ -0,0 +1,112 @@
...
+				expectedApps: []realm.App{app2},
+			},
+		} {
+			t.Run(tc.description, func(t *testing.T) {
+				realmClient := mock.RealmClient{}
+				realmClient.FindAppsFn = func(filter realm.AppFilter) ([]realm.App, error) {
+					return tc.apps, nil
+				}
+
+				_, console, _, ui, consoleErr := mock.NewVT10XConsole()
```
makesitgo, line 87: can we separate out the tests that are interactive and that aren't?  im curious as to how these tests are passing if some of them don't prompt the user (how do we get past `console.ExpectString("Select App(s)")`?)

internal/commands/app/delete_test.go
```diff
@@ -0,0 +1,112 @@
...
+		description    string
+		inputs         deleteInputs
+		apps           []realm.App
+		deleteErr      error
+		expectedApps   []string
+		expectedOutput string
+	}{
+		{
+			description:    "should delete no apps if none are found",
+			inputs:         deleteInputs{},
```
makesitgo, line 36: [nit] can omit this

internal/commands/app/delete_test.go
```diff
@@ -0,0 +1,112 @@
...
+			expectedOutput: "01:23:45 UTC INFO  No apps to delete\n",
+		},
+		{
+			description:  "with no project flag set and an apps flag set should delete all apps that match the apps flag",
+			inputs:       deleteInputs{Apps: []string{"app1"}},
+			apps:         []realm.App{app1},
```
makesitgo, line 42: if im understanding the test correctly, might be worth making making an `app2` and including it here.

looking at the next test, maybe even a 3rd test app where:
* app1: groupID1, Name: app1
* app2: groupID1, Name: app2
* app3: groupID2, Name: app1

That way you can prove apps of the wrong name are filtered out here, and apps of the wrong Project are filtered out in the test below

makesitgo (commented):
first pass here, but it's looking mighty nice and clean.  great work here

internal/commands/app/delete.go
```diff
@@ -0,0 +1,78 @@
...
+		appDeleteTableHeaders,
+		appDeleteRows(outputs)...,
+	))
+
+	ui.Print(logs...)
+	return nil
+}
+
+var (
+	appDeleteTableHeaders = []string{headerID, headerName, headerDeleted, headerDetails}
```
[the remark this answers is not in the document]
kpatel71716: changed to tableHeadersDelete (I like removing App too since we have the context of the command/structure).

I can make these changes across all the commands if we want to be consistent here

internal/commands/app/delete_inputs.go
```diff
@@ -0,0 +1,79 @@
...
+	apps, err := client.FindApps(realm.AppFilter{GroupID: inputs.Project})
+	if err != nil {
+		return nil, err
+	}
+
+	if len(inputs.Apps) > 0 && len(apps) == 0 {
+		return nil, cli.ErrAppNotFound{}
+	}
+
+	if len(apps) == 0 {
```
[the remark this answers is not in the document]
kpatel71716: This was intentionally 0, since this would be hit if we found no apps and the user did not provide any inputs with the apps flag. It just serves as an early return. Here we still want to make prompt the user even is only 1 app is found so we don't automatically delete their app (which is why I modified this logic from ResolveApp a bit here)

internal/commands/app/delete_inputs_test.go
kpatel71716: Done! I was diving into this for another issue (related to comment I had) and I think the test pass because `console.ExpectString("Select App(s)")` will return an error that we do not check (so the code just keeps executing)

internal/commands/app/delete_test.go
kpatel71716: Really good idea here. Done!

internal/commands/app/delete_inputs_test.go
```diff
@@ -0,0 +1,137 @@
...
+			_, console, _, ui, consoleErr := mock.NewVT10XConsole()
+			assert.Nil(t, consoleErr)
+			defer console.Close()
+
+			doneCh := make(chan (struct{}))
+			go func() {
+				defer close(doneCh)
+				console.ExpectString("unable to delete certain apps because they were not found: nonexistent")
+				console.ExpectEOF()
+			}()
```
kpatel71716, line 125: I noticed some weird behavior here that is also in some other related tests. `console.ExpectString` returns an EOF error (that we do not check) if the expected string is not found by the time we read the buffer. The weird thing is that when I check the returned string from this call, I only get the first 16(?) characters instead of the entire line. I was diving into this and couldn't quite figure out why the entire log isn't being returned by this call? Similar tests also have the same behavior where the expectString call doesn't check the string parameter properly

makesitgo: Nathan ran into some output funkiness when trying to use this stuff too... I think for now until we can get to the bottom of what's going on here:
* we should use `out, ui := mock.NewUI()` for tests that require no interactivity and where we want to verify the actual command output
* we should use `_, console, _, ui, err := mock.NewV10XConsole()` where we need interactivity, don't care too much about the output just yet (would love to fix this though), and can verify our tests "pass" when they complete with their various "expects"

The unchecked errors worry me...I'm surprised the patterns out there encourage this type of usage, but I think we should eventually (future ticket) look into making sure any errors here are in fact checked and reported on

internal/commands/app/delete.go
```diff
@@ -0,0 +1,76 @@
...
+	deletedCount := 0
+	for _, app := range apps {
+		err := clients.Realm.DeleteApp(app.GroupID, app.ID)
+		if err == nil {
+			deletedCount++
+		}
+		outputs = append(outputs, appOutput{app, err})
+	}
+
+	if len(outputs) == 0 {
```
makesitgo, line 40: minor optimization: move this to before `outputs := ...` and check the `len(apps)` instead

internal/commands/app/delete_inputs.go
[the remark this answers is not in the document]
makesitgo: ah i see it now.... can we rearrange this bit then:

```go
if len(apps) == 0 {
  if len(i.Apps) == 0 {
    return nil, nil
  }
  return nil, cli.ErrAppNowFound{}
}
```

internal/commands/app/delete_inputs.go
```diff
@@ -0,0 +1,85 @@
...
+
+		if len(missingApps) > 0 {
+			ui.Print(terminal.NewWarningLog(
+				"unable to delete certain apps because they were not found: %s",
+				strings.Join(missingApps, ", "),
+			))
+		}
+
+		if len(appsFiltered) > 0 {
+			return appsFiltered, nil
```
makesitgo, line 62: question: would we want to just `return` here always, regardless if the length is 0 or not?  that way they get the warning that none of their apps were found, but we don't then also prompt them to pick from a list of apps

internal/commands/app/delete_inputs_test.go
```diff
@@ -0,0 +1,137 @@
...
+			apps, err := inputs.resolveApps(ui, realmClient)
+			assert.Nil(t, err)
+
+			console.Tty().Close() // flush the writers
+			<-doneCh              // wait for procedure to complete
+
+			assert.Equal(t, []realm.App{app2}, apps)
+		})
+
+		t.Run("should print a warning if input apps are not found", func(t *testing.T) {
```
makesitgo, line 110: does this test need to be "interactive"?  doesn't seem like we actually prompt for anything in this flow, we could probably just use `mock.NewUI` for this

makesitgo (approved):
just a few questions/cleanup from my end.... should be good to go after that

🆒  beans

internal/commands/app/delete_inputs.go
kpatel71716: Hmm that is a good point. Thinking about this, I agree with your point of not prompting in this scenario. I will change this to `if len(inputs.Apps) > 0` so we can return appsFiltered if the user specified app inputs

makesitgo (approved):
🆒  beans

nathantfrank (approved):
LGTM :+1:

PR #78, opened 2021-02-24
REALMC-8281: Support not having environments folder

andrewzlchen:


utils/utils.go
```diff
@@ -261,7 +263,7 @@ func unmarshalJSONFiles(path string, ignoreDirErr bool) ([]interface{}, error) {
 func unmarshalJSONFilesWithFilenames(path string) (map[string]interface{}, error) {
 	fileInfos, err := ioutil.ReadDir(path)
 	if err != nil {
-		return nil, err
+		return nil, errEnvironmentsDirNotFound
```
makesitgo, line 266: i think its fair to say we want to avoid any "environments-specific" logic to be held inside a helper function like `unmarshalJSONFilesWithFilenames` here

we can keep the `return nil, err` here and push this type of check up to where you call it around L#230.  i _think_ that the `ioutil.ReadDir` method should return an error that could be checked with `os.IsNotExist` (a quick test should prove that out), otherwise you can always just call `_, err := os.Stat(environmentsDir)` and inspect that one instead to verify the directory's existence or not

makesitgo (commented):
just one thought on logic placement here.  and one question i should probably already know the answer to: we auto-export this directory with those files for all config versions, right?  so in theory, eventually all (active) apps end up with this directory?

utils/utils.go
andrewzlchen: Oh yeah good point, you're right that I shouldn't have put that into a util function.

Sadly, doing `os.IsNotExists` doesn't work, so I'll just first check for existence on environments

utils/utils.go
```diff
@@ -225,7 +225,14 @@ func UnmarshalFromDir(path string) (map[string]interface{}, error) {
 	}
 	app[servicesName] = services
 
-	environments, err := unmarshalJSONFilesWithFilenames(filepath.Join(path, environmentsName))
+	environmentsPath := filepath.Join(path, environmentsName)
+	_, err = os.Stat(environmentsPath)
+	if err != nil {
```
makesitgo, line 230: i'm thinking we probably to invert this:

```go
environmentsPath := filepath.Join(path, environmentsName)
_, err = os.Stat(environmentsPath)
if err == nil {
  environments, err := unmarshalJSONFilesWithFilenames(environmentsPath)
  // ...
}
```

that way future devs won't have to know to make any additions here above the `environments` processing (hopefully we're not adding more code ever, but I think just with the way this whole function is constructed, it makes sense to go this route)

makesitgo (commented):
one more thought here...also just confirming that the deletion of these `environments` files in `simple_app` is effectively writing tests for this, where the tests are still able to complete successfully?

utils/utils.go
andrewzlchen: Sure, that makes sense I'll do that.

1.

andrewzlchen:
The original bug was that not having environments would cause the import to fail because it tries to unmarshal environments directory. The original removal of the other environment directory in `testdata/new_app` confirmed that.

So effectively not having that directory seems to work well as a test.

makesitgo (approved):
:cool: beans

PR #79, opened 2021-02-24
REALMC-8134: Implement dry-run for app create command

nathantfrank:


internal/cloud/atlas/clusters.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Cluster contains non sensitive data about an Atlas cluster
+type Cluster struct {
+	ID    string `json:"id"`
+	Name  string `json:"name"`
+	State string `json:"stateName"`
+}
+
+const (
+	clustersByGroupIDPattern = "/api/atlas/v1.0/groups/%s/clusters"
```
nathantfrank, line 19: I noticed for the other endpoints the route is something like "/api/public/v1.0", which wasn't working getting a 404. So switched to what the docs point to.

internal/commands/app/create.go
```diff
@@ -65,7 +65,35 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
+
+		headers := []string{"Info", "Details"}
+		rows := make([]map[string]interface{}, 0, 4)
+		rows = append(rows, map[string]interface{}{"Info": "Client App ID", "Details": "N/A"})
+		rows = append(rows, map[string]interface{}{"Info": "Realm Directory", "Details": dir})
+		rows = append(rows, map[string]interface{}{"Info": "Realm UI", "Details": "N/A"})
+		if cmd.inputs.DataSource != "" {
+			rows = append(rows, map[string]interface{}{"Info": "Data Source", "Details": cmd.inputs.DataSource})
+		}
+		ui.Print(terminal.NewTableLog("Successful dry run of app create", headers, rows...))
```
nathantfrank, line 94: Open to suggestions on the N/A's and the overall msg

nathantfrank (commented):
First Pass

internal/cloud/atlas/clusters.go
makesitgo: yeah i noticed this `public` vs `atlas` discrepancy when first implementing the atlas client, and stuck with what we previously used.

i think this is fine here, but maybe add `atlasAPI = "/api/atlas/v1.0"` in the `client` file?

internal/utils/test/mock/atlas_client.go
```diff
@@ -17,3 +18,13 @@ func (ac AtlasClient) Groups() ([]atlas.Group, error) {
 	}
 	return ac.Client.Groups()
 }
+
+// ClustersByGroupID calls the mocked Groups implementation if provided,
```
makesitgo, line 22: copy/paste here: ~`Groups`~

internal/cloud/atlas/clusters.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+func (c *client) ClustersByGroupID(groupID string) ([]Cluster, error) {
```
makesitgo, line 22: [nit] I'd name this just `Clusters` (we don't really use "ByGroupID" or "ByAppID" in any of the other routes, even though most require it)

along those same lines, the path pattern can have `ByGroupID` removed as well, and the unexpected status code below can be just "get clusters"

internal/cloud/atlas/clusters.go
```diff
@@ -0,0 +1,44 @@
...
+	)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode != http.StatusOK {
+		return nil, api.ErrUnexpectedStatusCode{"get group clusters", res.StatusCode}
+	}
+	defer res.Body.Close()
+
+	var clusters struct {
```
makesitgo, line 36: [nit] move this type outside the function, can be just `clustersResponse` or something

internal/commands/app/create.go
makesitgo: lets keep the dry run output simple/concise as best we can here, we can basically report on two things:
* we _would_ create an app (from scratch/from app <id>) at <dir>
* we _would_ link a data source (cluster name/data lake)

i think the one wrinkle is if/when we cant resolve the data source specified, we could just update that last message to indicate as much: "we would not link a data source because the Atlas cluster/data lake, 'my-data-source', cannot be found"

we can also have a follow up command here to try running the same command without --dry-run included

makesitgo: also I wanted to ask: why are we "resolving data source" differently here than with the `list_commands` below?  is there a way to consolidate those two?

makesitgo (commented):
quick first pass here with some thoughts/questions

internal/commands/app/create.go
nathantfrank: discussed offline, and consolidated

internal/commands/app/create.go
```diff
@@ -22,12 +23,47 @@ type CommandCreate struct {
...
 	fs.VarP(&cmd.inputs.Location, flagLocation, flagLocationShort, flagLocationUsage)
 	fs.StringVarP(&cmd.inputs.DataSource, flagDataSource, flagDataSourceShort, "", flagDataSourceUsage)
-	// TODO(REALMC-8134): Implement dry-run for app create command
-	// fs.BoolVarP(&cmd.inputs.DryRun, flagDryRun, flagDryRunShort, false, flagDryRunUsage)
+	fs.BoolVarP(&cmd.inputs.DryRun, flagDryRun, flagDryRunShort, false, flagDryRunUsage)
+}
+
+func (cmd *CommandCreate) commandString(omitDryRun bool) string {
+	sb := strings.Builder{}
+	sb.WriteString("app create")
```
makesitgo, line 36: should probably prefix this with `realm-cli` ... and thinking on it, we probably want that to be shared with the `cliName` const we currently have in `cmd/root.go`.

let's move that `const` to the `internal/cli` package and make it a `var` instead.  when we ultimately build our CLI, we can use build-time substitution to set this `var` to our actual CLI name (will still be `realm-cli` at this point), and then our commands here (this isn't the first "command string" we've tried to build) can reference `cli.Name` to get that same name value.

that way, we're consistent with our names no matter what it ends up being

---

also, meganit, but can we move this function towards the bottom of the file, at least after the `Handler` code?

internal/commands/app/create.go
```diff
@@ -22,12 +23,47 @@ type CommandCreate struct {
...
+	}
+	if cmd.inputs.From != "" {
+		sb.WriteString(" --from ")
+		sb.WriteString(cmd.inputs.From)
+	}
+	if cmd.inputs.Directory != "" {
+		sb.WriteString(" --app-dir ")
+		sb.WriteString(cmd.inputs.Directory)
+	}
+	sb.WriteString(" --deployment-model ")
```
makesitgo, line 54: any reason we didn't have conditionals around location/deployment model?

internal/commands/app/create.go
```diff
@@ -65,7 +101,27 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
-	// TODO(REALMC-8134): Implement dry-run for app create command
+	var ds dataSource
+	if cmd.inputs.DryRun {
+		if cmd.inputs.DataSource != "" {
+			ds, err = cmd.inputs.resolveDataSource(clients.Atlas, groupID)
+			if err != nil {
+				return err
+			}
+		}
+		logs := make([]terminal.Log, 0, 3)
```
makesitgo, line 112: i think instead of the logs array, you could probably just call `ui.Print` (especially now that we don't have errors returned from that function) at each spot

internal/commands/app/create.go
```diff
@@ -65,7 +101,27 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
 		return err
 	}
 
-	// TODO(REALMC-8134): Implement dry-run for app create command
+	var ds dataSource
+	if cmd.inputs.DryRun {
+		if cmd.inputs.DataSource != "" {
+			ds, err = cmd.inputs.resolveDataSource(clients.Atlas, groupID)
```
makesitgo, line 107: so it seems like `resolveDataSource` is invoked regardless of whether we're in a dry run or not.  do you think it'd be worth doing that before this `DryRun` check?

```go
var ds dataSource
if cmd.inputs.DataSource != "" {
  ds, err = cmd.inputs.resolveDataSource(clients.Atlas, groupID)
  if err != nil {
    return err
  }
}

if cmd.inputs.DryRun {
  // printing logic here
  return nil
}

// further down in the handler
if cmd.inputs.DataSource != "" {
  var dataSourceDir string
  switch loadedApp.ConfigVersion() { ... }
}

// run import...
```

makesitgo (commented):
quick pass here with a few more questions.  should be good to go afterwards, but i'll hold off for now

internal/commands/app/create.go
```diff
@@ -22,12 +23,47 @@ type CommandCreate struct {
 func (cmd *CommandCreate) Flags(fs *pflag.FlagSet) {
 	fs.StringVar(&cmd.inputs.Project, flagProject, "", flagProjectUsage)
 	fs.StringVarP(&cmd.inputs.Name, flagName, flagNameShort, "", flagNameUsage)
+	fs.StringVarP(&cmd.inputs.From, flagFrom, flagFromShort, "", flagFromUsage)
```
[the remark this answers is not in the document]
nathantfrank: Yeah, since the testing is really around the inputs struct (and is set directly) just missed needing this till I realized it was missing from the command.

internal/commands/app/create.go
nathantfrank: Those variables are guaranteed as of now to always have a value where this function is called (handler). Can add them if you like.

makesitgo: if they are empty/default values, lets leave them out of the printed string, on the condition that leaving them out doesn't then force the user into answering some prompt about which location/deployment model they want

basically, let's print the user the minimally viable command string here, all while not forcing them to hit another prompt

internal/commands/app/create.go
makesitgo: sounds good!  feel free to leave as is here then

internal/commands/app/create_test.go
```diff
@@ -269,7 +272,7 @@ func TestAppCreateHandler(t *testing.T) {
 				DataSource: "test-cluster"},
 		}
 
-		assert.Nil(t, cmd.Handler(profile, ui, cli.Clients{Realm: client}))
+		assert.Nil(t, cmd.Handler(profile, ui, cli.Clients{rc, ac}))
```
makesitgo, line 275: I'm adding a 3rd client with hosting changes, do you mind using the Field names here?

internal/commands/app/create_test.go
```diff
@@ -297,12 +300,121 @@ func TestAppCreateHandler(t *testing.T) {
...
+
+		expectedDir := filepath.Join(profile.WorkingDirectory, "test-app")
+		assert.Equal(t, strings.Join([]string{
+			"01:23:45 UTC INFO  A Realm app based on the Realm app from-app would be created at " + expectedDir,
+			"01:23:45 UTC DEBUG To create this app run: " + cmd.commandString(true),
+			"",
+		}, "\n"), out.String())
+	})
+
+	t.Run("should create a minimal project dry run with data source set", func(t *testing.T) {
```
makesitgo, line 382: [suggestion] seems like we could table these dry-run tests to help read them all a bit easier

---

it's also probably worth adding a quick couple of unit tests (for empty & full inputs) on the `commandString` func

makesitgo (approved):
good by me, just a couple of test suggestions before we merge this in

🆒  beans

kpatel71716 (approved):
LGTM

internal/commands/app/create_test.go
nathantfrank: added the two suggested tests

PR #80, opened 2021-02-25
REALMC-8185 move refreshAuth option check to execute earlier

kpatel71716:


internal/cloud/realm/auth.go
```diff
@@ -76,18 +76,18 @@ func (c *client) AuthProfile() (AuthProfile, error) {
...
 
-	if options.RefreshAuth {
-		if session.RefreshToken == "" {
+	if !options.NoAuth {
+		if session.AccessToken == "" {
 			return "", ErrInvalidSession{}
 		}
-		return session.RefreshToken, nil
+		return session.AccessToken, nil
 	}
```
kpatel71716, line 91: This diff is a bit noisy, but I actually just moved one small block in front of another

internal/cloud/realm/auth_test.go
```diff
@@ -78,16 +78,46 @@ func TestRealmAuthRefresh(t *testing.T) {
 		assert.Equal(t, nil, err)
 
 		// invalidate the session's tokens
-		session.AccessToken = ""
 		session.RefreshToken = session.AccessToken
+		session.AccessToken = ""
```
kpatel71716, line 82: Not super important, but I flipped these so both weren't set to "" (doesnt affect the result since we return invalid session regardless of the actual refresh error here)

internal/cloud/realm/auth_test.go
```diff
@@ -78,16 +78,46 @@ func TestRealmAuthRefresh(t *testing.T) {
...
 
 		profile := mock.NewProfileWithSession(t, session)
 
 		client = realm.NewAuthClient(profile.RealmBaseURL(), profile)
 		_, err = client.AuthProfile()
 		assert.Equal(t, realm.ErrInvalidSession{}, err)
 	})
-	// TODO: REALMC-7719 add test for expired credentials and test for ensuring profile cleared on invalid session
```
kpatel71716, line 90: I am comfortable with the scope of these tests now. We can use a non-mocked profile if we actually want to check that the tokens get set/reset appropriately. I am fine doing that in this ticket as well

internal/cloud/realm/auth_test.go
```diff
@@ -78,16 +79,58 @@ func TestRealmAuthRefresh(t *testing.T) {
...
+
+	t.Run("with an expired access token", func(t *testing.T) {
+		u.SkipUnlessExpiredAccessTokenPresent(t)
+		t.Run("should use the refresh token to generate a new access token", func(t *testing.T) {
```
makesitgo, line 94: [meganit] add a newline b/w and the `u.Skip` above

internal/cloud/realm/auth_test.go
```diff
@@ -78,16 +79,58 @@ func TestRealmAuthRefresh(t *testing.T) {
...
+			client := realm.NewClient(u.RealmServerURL())
+			u.SkipUnlessRealmServerRunning(t)
+			session, err := client.Authenticate(u.CloudUsername(), u.CloudAPIKey())
+			assert.Equal(t, nil, err)
```
makesitgo, line 98: [nit] use `assert.Nil(t, err)` here

internal/cloud/realm/auth_test.go
```diff
@@ -78,16 +79,58 @@ func TestRealmAuthRefresh(t *testing.T) {
...
 		_, err = client.AuthProfile()
 		assert.Equal(t, realm.ErrInvalidSession{}, err)
 	})
-	// TODO: REALMC-7719 add test for expired credentials and test for ensuring profile cleared on invalid session
+
+	t.Run("with an expired access token", func(t *testing.T) {
+		u.SkipUnlessExpiredAccessTokenPresent(t)
+		t.Run("should use the refresh token to generate a new access token", func(t *testing.T) {
+			client := realm.NewClient(u.RealmServerURL())
+			u.SkipUnlessRealmServerRunning(t)
```
makesitgo, line 96: should be able to remove this, it's called at the top of `TestRealmAuthRefresh` (which I _think_ we're under here? hard to tell on github...)

internal/cloud/realm/auth_test.go
```diff
@@ -78,16 +79,58 @@ func TestRealmAuthRefresh(t *testing.T) {
...
+			session, err := client.Authenticate(u.CloudUsername(), u.CloudAPIKey())
+			assert.Equal(t, nil, err)
+
+			session.AccessToken = u.ExpiredAccessToken() // Set the access token to an expired token
```
makesitgo, line 100: also fine to just remove this line, and use `u.ExpiredAccessToken()` directly when calling `profile.SetSession`

internal/cloud/realm/auth_test.go
```diff
@@ -78,16 +79,58 @@ func TestRealmAuthRefresh(t *testing.T) {
...
+
+			profile, teardown := mock.NewProfileFromTmpDir(t, "auth_refresh_test")
+			defer teardown()
+			profile.SetRealmBaseURL(u.RealmServerURL())
+			profile.SetSession(auth.Session{session.AccessToken, session.RefreshToken})
+
+			client = realm.NewAuthClient(profile.RealmBaseURL(), profile)
+			_, err = client.AuthProfile()
+			assert.Nil(t, err)
```
makesitgo, line 109: does it work here to `assert` on the `profile` having new/updated `Session` details?

internal/cloud/realm/auth_test.go
```diff
@@ -78,16 +79,58 @@ func TestRealmAuthRefresh(t *testing.T) {
...
+		})
+		t.Run("should error if both the access and refresh tokens are expired", func(t *testing.T) {
```
makesitgo, line 111: [nit] newline between `t.Run` blocks

internal/cloud/realm/auth_test.go
```diff
@@ -78,16 +79,58 @@ func TestRealmAuthRefresh(t *testing.T) {
...
+			client := realm.NewClient(u.RealmServerURL())
+			u.SkipUnlessRealmServerRunning(t)
+			session, err := client.Authenticate(u.CloudUsername(), u.CloudAPIKey())
+			assert.Equal(t, nil, err)
+
+			session.AccessToken = u.ExpiredAccessToken()  // Set the access token to an expired token
+			session.RefreshToken = u.ExpiredAccessToken() // Set the refresh token to an random expired token
```
makesitgo, line 118: i think again, just pass the `u.ExpiredAccessToken()` values directly to the `SetSession` call.

you can add a comment there saying:

```go
// set the access and refresh token values to expired tokens
profile.SetSession(auth.Session{u.ExpiredAccessToken(), u.ExpiredAccessToken()})
```

internal/cloud/realm/auth_test.go
```diff
@@ -78,16 +79,58 @@ func TestRealmAuthRefresh(t *testing.T) {
...
+			profile.SetSession(auth.Session{session.AccessToken, session.RefreshToken})
+
+			client = realm.NewAuthClient(profile.RealmBaseURL(), profile)
+			_, err = client.AuthProfile()
+			assert.Nil(t, err)
+		})
+		t.Run("should error if both the access and refresh tokens are expired", func(t *testing.T) {
+			client := realm.NewClient(u.RealmServerURL())
+			u.SkipUnlessRealmServerRunning(t)
+			session, err := client.Authenticate(u.CloudUsername(), u.CloudAPIKey())
```
makesitgo, line 114: do you even need to `Authenticate` ahead of time here?

internal/cloud/realm/auth_test.go
```diff
@@ -78,16 +79,58 @@ func TestRealmAuthRefresh(t *testing.T) {
...
+			defer teardown()
+			profile.SetRealmBaseURL(u.RealmServerURL())
+			profile.SetSession(auth.Session{session.AccessToken, session.RefreshToken})
+
+			client = realm.NewAuthClient(profile.RealmBaseURL(), profile)
+			_, err = client.AuthProfile()
+			assert.Equal(t, realm.ErrInvalidSession{}, err)
+
+			t.Log("and clear the session tokens")
+			assert.Equalf(t, "", profile.Session().AccessToken, "access token was not cleared")
```
makesitgo, line 130:
```suggestion
			assert.Equalf(t, "", profile.Session().AccessToken, "access token was not cleared: %q", profile.Session().AccessToken)
```

internal/cloud/realm/auth_test.go
```diff
@@ -78,16 +79,58 @@ func TestRealmAuthRefresh(t *testing.T) {
...
+			assert.Equalf(t, "", profile.Session().RefreshToken, "refresh token was not cleared")
```
makesitgo, line 131:
```suggestion
			assert.Equalf(t, "", profile.Session().RefreshToken, "refresh token was not cleared: %q", profile.Session().RefreshToken)
```

makesitgo: also might be worth just pulling out `profile.Session()` into its own `var` here for convenience

internal/utils/test/mock/cli_profile.go
```diff
@@ -33,7 +33,13 @@ func NewProfileFromTmpDir(t *testing.T, name string) (*cli.Profile, func()) {
 	profile := NewProfile(t)
 	profile.WorkingDirectory = tmpDir
 
-	return profile, teardown
+	_, resetHomeDir := u.SetupHomeDir(tmpDir)
+
+	return profile,
+		func() {
+			teardown()
+			resetHomeDir()
```
makesitgo, line 41: [ubernit] let's reset home dir first

internal/utils/test/test.go
```diff
@@ -103,6 +103,11 @@ func RealmServerURL() string {
 	return defaultRealmServerURL
 }
 
+// ExpiredAccessToken returns an expired access token to use for testing
+func ExpiredAccessToken() string {
+	return os.Getenv("BAAS_MONGODB_EXPIRED_ACCESS_TOKEN")
```
makesitgo, line 108: can we update this to follow the pattern used with `RealmServerURL()`?

basically, it forces to write tests that are aware of the environment restrictions they require, its fairly explicit about what's needed (from an environment standpoint) to get successful test execution.

the main difference with this implementation is we won't have a `defaultExpiredAccessToken`, and instead we are fine to just return `""` (you are going to keep the check for `""` in `SkipUnlessExpiredAccessTokenPresent`)

makesitgo (commented):
so happy you got to the bottom of this, just a bunch of random testing questions/suggestions here

internal/cloud/realm/auth_test.go
kpatel71716: I don't mind this change but I went with the above message just in case we don't want to print the auth tokens in the logs (not sure how big of a deal it is in this case). Thoughts?

internal/cloud/realm/auth_test.go
kpatel71716: Yep! added a couple of asserts

internal/cloud/realm/auth_test.go
kpatel71716: Nope I removed it. My original thinking was that we could "check" that the call succeeds before we make it fail.. but that is probably unnecessary now that I think about it

internal/cloud/realm/auth_test.go
kpatel71716: Looks like we need to call profile.Session explicitly here to get the correct values, so I left as is

internal/cloud/realm/auth_test.go
makesitgo: yeah thats fine then, lets leave it as is and out of the message here.  better safe than sorry

internal/cloud/realm/auth_test.go
makesitgo: I'm not sure how you're coming up with that, this code/test ran fine for me:

```go
		t.Run("should error if both the access and refresh tokens are expired", func(t *testing.T) {
			profile, teardown := mock.NewProfileFromTmpDir(t, "auth_refresh_test")
			defer teardown()
			profile.SetRealmBaseURL(u.RealmServerURL())
			// set the access and refresh token values to expired tokens
			profile.SetSession(auth.Session{u.ExpiredAccessToken(), u.ExpiredAccessToken()})

			client := realm.NewAuthClient(profile.RealmBaseURL(), profile)
			_, err := client.AuthProfile()
			assert.Equal(t, realm.ErrInvalidSession{}, err)

			session := profile.Session()

			t.Log("and clear the session tokens")
			assert.Equalf(t, "", session.AccessToken, "access token was not cleared")
			assert.Equalf(t, "", session.RefreshToken, "refresh token was not cleared")
		})
```

fwiw, especially with not printing the details, this comment no longer really matters.  i just want to make sure we're still on the same page and not missing something nefarious here

kpatel71716: Oh fair point, I created the `session` var above the preceding code block since I misinterpreted. I can make that change since I like it too

makesitgo (approved):
🆒  beans

PR #82, opened 2021-02-25
REALMC-8240 Add segment tracking to realm-cli

jo-tse-mongo:


go.mod
```diff
@@ -5,20 +5,23 @@ go 1.14
 require (
 	github.com/AlecAivazis/survey/v2 v2.2.2
 	github.com/Netflix/go-expect v0.0.0-20180615182759-c93bf25de8e8
+	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869 // indirect
```
makesitgo, line 8: i think you should run `go mod tidy` in your project and see what updates here... just curious if `segmentio` needs to bring all these in or if something was included accidentally

internal/cli/command_factory.go
```diff
@@ -151,7 +155,7 @@ func (factory *CommandFactory) SetGlobalFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&factory.profile.atlasBaseURL, flagAtlasBaseURL, "", flagAtlasBaseURLUsage)
 	fs.StringVar(&factory.profile.realmBaseURL, flagRealmBaseURL, "", flagRealmBaseURLUsage)
 	fs.VarP(&factory.profile.telemetryMode, telemetry.FlagMode, telemetry.FlagModeShort, telemetry.FlagModeUsage)
-
+	fs.StringVarP(&factory.profile.telemetryWriteKey, telemetry.FlagWriteKey, telemetry.FlagWriteKeyShort, "", telemetry.FlagWriteKeyUsage)
```
makesitgo, line 158: this should not be a global flag, we can remove this

internal/cli/command_factory.go
```diff
@@ -76,11 +76,15 @@ func (factory *CommandFactory) Build(command CommandDefinition) *cobra.Command {
 				os.Exit(1)
 			}
 
-			factory.telemetryService = telemetry.NewService(
+			telemetryService := telemetry.NewService(
 				factory.profile.telemetryMode,
 				factory.profile.User().PublicAPIKey,
+				factory.profile.telemetryWriteKey,
```
makesitgo, line 82: can remove this, we are going to store the write key as a `var` in the `telemetry` package

internal/cli/command_factory.go
```diff
@@ -76,11 +76,15 @@ func (factory *CommandFactory) Build(command CommandDefinition) *cobra.Command {
 				os.Exit(1)
 			}
 
-			factory.telemetryService = telemetry.NewService(
+			telemetryService := telemetry.NewService(
```
makesitgo, line 79: I'm guessing you originally changed this b/c you introduced an error return value to `NewService`, but since thats not needed can we revert this back to how it was?

internal/cli/profile.go
```diff
@@ -133,6 +135,15 @@ func (p *Profile) resolveFlags() error {
 	}
 	p.SetString(keyTelemetryMode, string(p.telemetryMode))
 
+	if p.telemetryWriteKey == "" {
```
makesitgo, line 138: all of this write key stuff should not be part of the profile, it should not be configurable by the user (they should only be able to flip the telemetry mode here)

internal/telemetry/event.go
```diff
@@ -37,3 +37,11 @@ type EventDataKey string
 const (
 	EventDataKeyErr EventDataKey = "err"
 )
+
+func (e *event) createPropertyMap() map[string]interface{} {
```
makesitgo, line 41: lets make this a bit more go-like:

```go
func eventDataProperties(data []EventData) map[string]interface{} {
  // func impl here
}
```

internal/telemetry/flags.go
```diff
@@ -12,6 +12,10 @@ const (
 	FlagMode      = "telemetry"
 	FlagModeShort = "m"
 	FlagModeUsage = `enable or disable telemetry (this setting is remembered), available options: ["off", "on"]`
+
+	FlagWriteKey      = "write_key"
```
makesitgo, line 16: ultimately these should be able to be removed as well

internal/telemetry/service.go
```diff
@@ -26,7 +27,7 @@ func NewService(mode Mode, userID string, command string) *Service {
 	case ModeOff:
 		service.tracker = &noopTracker{}
 	case ModeEmpty, ModeOn:
-		service.tracker = &segmentTracker{}
+		service.tracker = newSegmentTracker(writeKey, logger)
```
makesitgo, line 30: the constructor here should only take in the `logger`

internal/telemetry/tracker.go
```diff
@@ -26,15 +30,24 @@ func (tracker *stdoutTracker) Track(event event) {
 	)
 }
 
-type segmentTracker struct{}
+type segmentTracker struct {
+	client analytics.Client
+	logger *log.Logger
+}
+
+func newSegmentTracker(writeKey string, logger *log.Logger) *segmentTracker {
```
makesitgo, line 38: ok, so here is where you're going to want:

```go
var (
  segmentWriteKey = "" // will substitute value at build time with proper key
)
```

and then you can just refer to that when you call `analytics.New`

makesitgo: still trying to think about what we want to do to protect ourselves if this build substitution doesn't happen properly ... 

i think we maybe re-introduce that `error` return value in the telemetry service constructor, but lets discuss this more offline since you have a better idea of the nuances here.  i just want to make sure failed telemetry doesn't prevent the CLI from running at all

makesitgo (commented):
just a quick first pass here, lets sync up tomorrow about that error handling

go.mod
jo-tse-mongo: I did `go mod tidy`, but I didn't see any changes to commit

internal/telemetry/tracker.go
jo-tse-mongo: I didn't reintroduce the error for the constructor for now and made it return another tracker instead, since it's not really an error that we should fail the CLI for. I'm unsure how the `stdout` tracker would be used (other than maybe dev), but maybe we could return that instead of the `segment` tracker

makesitgo: the `stdout` tracker may go away all together... its noisy and and not all that helpful for anyone involved haha (it was cool during early developemnt).

i like the idea of just transitioning over to use the `noopTracker` if we find that `segmentWriteKey` is empty, let's roll with that

internal/cli/command_factory.go
```diff
@@ -151,7 +152,6 @@ func (factory *CommandFactory) SetGlobalFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&factory.profile.atlasBaseURL, flagAtlasBaseURL, "", flagAtlasBaseURLUsage)
 	fs.StringVar(&factory.profile.realmBaseURL, flagRealmBaseURL, "", flagRealmBaseURLUsage)
 	fs.VarP(&factory.profile.telemetryMode, telemetry.FlagMode, telemetry.FlagModeShort, telemetry.FlagModeUsage)
-
```
makesitgo, line 154: unnecessary diff here

internal/cli/profile.go
```diff
@@ -34,9 +34,10 @@ const (
 
 // Profile is the CLI profile
 type Profile struct {
-	Name          string
-	atlasBaseURL  string
-	realmBaseURL  string
+	Name         string
+	atlasBaseURL string
+	realmBaseURL string
+
```
makesitgo, line 40: unnecessary diff here as well

internal/telemetry/service_test.go
```diff
@@ -9,8 +9,7 @@ import (
 
 func TestNewService(t *testing.T) {
 	t.Run("Should create the expected Service", func(t *testing.T) {
-		service := NewService(ModeOn, "userID", "command")
-
```
makesitgo, line 13: keep the newline here, in general I'd suggest trying to keep the amount of diffs only to what's necessary.  the newline stuff (adding/removing) just adds noise to the PRs

internal/telemetry/tracker.go
```diff
@@ -2,7 +2,15 @@ package telemetry
...
 	"fmt"
+	"log"
 	"time"
+
+	"go.mongodb.org/mongo-driver/bson/primitive"
+	"gopkg.in/segmentio/analytics-go.v3"
+)
+
+var (
+	segmentWriteKey = ""
```
makesitgo, line 13:
```suggestion
	segmentWriteKey = "" // value will be injected at build-time
```

internal/telemetry/tracker.go
```diff
@@ -26,15 +34,28 @@ func (tracker *stdoutTracker) Track(event event) {
...
 
-type segmentTracker struct{}
+type segmentTracker struct {
+	client analytics.Client
+	logger *log.Logger
+}
+
+func newSegmentTracker(logger *log.Logger) Tracker {
+	if len(segmentWriteKey) == 0 {
+		log.Print("unable to make a Segment tracker with an empty write key; replacing with a noop")
```
makesitgo, line 44:
```suggestion
		log.Print("unable to connect to Segment due to missing key, CLI telemetry will be disabled")
```

internal/telemetry/tracker.go
```diff
@@ -26,15 +34,28 @@ func (tracker *stdoutTracker) Track(event event) {
...
 func (tracker *segmentTracker) Track(event event) {
-	fmt.Printf(
-		"%s UTC TELEM %s: %s%v\n",
-		event.time.In(time.UTC).Format("15:04:05"),
-		event.command,
-		event.eventType,
-		event.data,
-	)
+	if err := tracker.client.Enqueue(analytics.Track{
+		MessageId:  primitive.NewObjectID().Hex(),
```
makesitgo, line 53: use `event.id` here

internal/telemetry/tracker.go
```diff
@@ -26,15 +34,28 @@ func (tracker *stdoutTracker) Track(event event) {
...
+		Timestamp:  time.Now(),
```
makesitgo, line 54: use `event.time` here

internal/telemetry/tracker.go
```diff
@@ -26,15 +34,28 @@ func (tracker *stdoutTracker) Track(event event) {
...
+		Event:      string(event.eventType),
+		UserId:     event.userID,
+		Properties: eventDataProperties(event.data),
```
makesitgo, line 57: I might have sent you down a bad path here, but we probably don't even need `eventDataProperties` as a function (and could/should handle that logic just here in this function).

the thing that's missing right now are two additional `EventData`s we want to care about:

```go
[]EventData{
  {"cmd", event.command}, // "cmd" => eventDataKeyCommand
  {"xid", event.executionID}, // "xid" => eventDataKeyExecutionID
}
```

since this is all just segment-specific stuff, i think using `make(map[string]interface{}, len(event.data)+2)` inside this function before the `client.Enqueue` call works fine here

internal/telemetry/tracker_test.go
```diff
@@ -0,0 +1,101 @@
...
+func (client *testClient) Close() error {
+	return nil
+}
+
+var testTime = time.Date(2021, 1, 2, 3, 4, 5, 6, time.UTC)
+
+func TestNoopTracker(t *testing.T) {
+	t.Run("should create a noop tracker", func(t *testing.T) {
+		tracker := noopTracker{}
+		assert.NotNil(t, tracker)
```
makesitgo, line 31: this assert is not doing much here, we can probably get rid of this outer test-scope here

internal/telemetry/tracker_test.go
```diff
@@ -0,0 +1,101 @@
...
+			testEvent := createEvent(EventTypeCommandStart, nil, "someCommand")
+			testTrackStdoutOutput(t, &tracker, testEvent, "")
+		})
+	})
+}
+
+func TestStdoutTracker(t *testing.T) {
+	t.Run("should create an stdout tracker", func(t *testing.T) {
+		tracker := stdoutTracker{}
+		assert.NotNil(t, tracker)
```
makesitgo, line 42: same here for getting rid of the assert

internal/telemetry/tracker_test.go
```diff
@@ -0,0 +1,101 @@
...
+		})
+	})
+}
+
+func TestSegmentTracker(t *testing.T) {
+	client := &testClient{}
+	t.Run("should create the segment tracker", func(t *testing.T) {
+		tracker := segmentTracker{}
+		tracker.client = client
+		assert.NotNil(t, tracker)
```
makesitgo, line 59: same thing here

internal/telemetry/tracker_test.go
```diff
@@ -0,0 +1,101 @@
...
+		t.Run("and should print the tracking information to the logger", func(t *testing.T) {
+			tracker.Track(createEvent(EventTypeCommandError, []EventData{{Key: EventDataKeyErr, Value: "Something"}}, "someCommand"))
+			testClientResults := tracker.client.(*testClient)
+			assert.Equal(t, 1, len(testClientResults.calls))
+
+			actualTrack := testClientResults.calls[0].(analytics.Track)
+			assert.NotNil(t, actualTrack)
+			assert.Equal(t, string(EventTypeCommandError), actualTrack.Event)
```
makesitgo, line 67: would we be able to have all of this in a single `assert.Equal` by building an `analytics.Track` ourselves?

something like: `assert.Equal(t, []interface{}{analytics.Track{ ... }}, testClientResults.calls)`

makesitgo (commented):
lookin real close here, i want to dive a little deeper into that `testTrackStdoutOutput` func (looks clever!) before giving the full 👍 ... here's my last bit of cleanup comments

internal/telemetry/tracker.go
```diff
@@ -26,15 +33,32 @@ func (tracker *stdoutTracker) Track(event event) {
...
-// TODO(REALMC-7243): use Segment sdk to send events through client
 func (tracker *segmentTracker) Track(event event) {
-	fmt.Printf(
-		"%s UTC TELEM %s: %s%v\n",
-		event.time.In(time.UTC).Format("15:04:05"),
-		event.command,
-		event.eventType,
-		event.data,
-	)
+	properties := make(map[string]interface{}, len(event.data))
```
makesitgo, line 51: btw we still want to be adding `event.command` and `event.executionID` as properties here (see my previous comment re: `"cmd"` and `"xid"` event data keys

internal/telemetry/tracker_test.go
```diff
@@ -0,0 +1,103 @@
...
+		expectedTrack := analytics.Track{
+			MessageId: testID,
+			UserId:    testUserID,
+			Timestamp: testTime,
+			Event:     string(EventTypeCommandError),
+			Properties: map[string]interface{}{
+				string(EventDataKeyErr): "Something",
+			},
+		}
+		actualTrack := testClientResults.calls[0].(analytics.Track)
```
makesitgo, line 72: out of curiosity, was something failing when you tried to run:

```go
assert.Equal(t, []interface{}{expectedTrack}, testClientResults.calls)
```

If something like that works let's use it, I'd like to keep the asserts as simple as possible.

If I'm missing something there and that doesn't work, let's at least avoid any potential panics within this test.  You should always try to use the two-return-value `, ok` syntax in tests and `assert.True` on the `ok` var (this gives us a nicer test failure message than panics will).  Also, the last assert here ("equals" on the two tracks) should cover the two asserts before that ("non nil" and "length of properties is 1")

internal/telemetry/tracker_test.go
```diff
@@ -0,0 +1,103 @@
...
+		assert.NotNil(t, actualTrack)
+		assert.Equal(t, 1, len(actualTrack.Properties))
+		assert.Equal(t, expectedTrack, actualTrack)
+	})
+}
+
+func testTrackStdoutOutput(t *testing.T, tracker Tracker, event event, expected string) {
```
makesitgo, line 79: left you a comment about moving this code from a helper to the test case itself, but here are a few other comments to clean up:
* always check the error (you're ignoring errors from `os.Pipe`, `w.Close()`, & `ioutil.ReadAll` here), you have `t *testing.T` here so you can `assert.Nil` easily
* in general, when using helper functions that pass around `t *testing.T` (presumably so you can make `assert` calls), you'll want to make sure to call `t.Helper()` at the top/first thing of the helper function (it ensures proper test output if/when a test fails)
  * fwiw, if/when you move this code to the test func itself, `t.Helper()` will become unnecessary
* b/c you have the defer which replaces `os.Stdout` with the original reference of it, you won't need to do it again on L#89 here

makesitgo (commented):
last bit of thoughts here, once you clean it all up this should be good to go. nice work!

internal/telemetry/tracker.go
jo-tse-mongo: Is there a reason why `event.Data` is an array? Can it just be `"err"`? It seems strange that `cmd` and `xid` are taken from the event at a different level than `err` is (on the `event` object itself vs within the `data` property)

makesitgo: if you look at the signature for `(*telemetry.Service).TrackEvent`, you'll see the minimal piece of information needed by the calling code is the `EventType` to track on, and it optionally supports any number of additional `EventData` tuples/structs/objects

for the "start" and "complete" events, we don't really care about any extra details, but on the "error" event we leverage that ability to include an error message

otherwise, the properties that we want to be able to see in every event tracked are provided at the time when we create the `telemetry.Service` (user id, command name).  those are known ahead of time, so we avoid passing them in every time to the `start`/`error`/`complete` stages

jo-tse-mongo: Should I add `cmd` and `xid` as `EventDataKey`s ?

makesitgo: we can add them as `EventDataKey`, but no need to export them:

```go
// set of event data keys
const (
	eventDataKeyCommand     EventDataKey = "cmd"
	eventDataKeyExecutionID EventDataKey = "xid"

	EventDataKeyErr EventDataKey = "err"
)
```

internal/telemetry/tracker_test.go
```diff
@@ -0,0 +1,108 @@
...
+			t,
+			&tracker,
+			createEvent(EventTypeCommandComplete, nil, "someCommand"),
+			"03:04:05 UTC TELEM someCommand: COMMAND_COMPLETE[]\n",
+		)
+	})
+}
+
+func TestSegmentTracker(t *testing.T) {
+	client := &testClient{}
```
makesitgo, line 55: [nit] move this down into the test scope its used in

internal/telemetry/tracker_test.go
```diff
@@ -0,0 +1,108 @@
...
+	t.Run("should create the segment tracker and should print the tracking information to the logger", func(t *testing.T) {
+		tracker := segmentTracker{}
+		tracker.client = client
+		tracker.Track(createEvent(EventTypeCommandError, []EventData{{Key: EventDataKeyErr, Value: "Something"}}, "someCommand"))
+
+		testClientResults := tracker.client.(*testClient)
```
makesitgo, line 61: you don't need this type cast, you can just use the ref to `client` above

internal/telemetry/tracker_test.go
```diff
@@ -0,0 +1,108 @@
...
+			Event:     string(EventTypeCommandError),
+			Properties: map[string]interface{}{
+				string(EventDataKeyErr):         "Something",
+				string(eventDataKeyCmd):         "someCommand",
+				string(eventDataKeyExecutionID): testExecutionID,
+			},
+		}
+
+		actualTrack := testClientResults.calls[0].(analytics.Track)
+		assert.NotNil(t, actualTrack)
```
makesitgo, line 75: the stuff with `actualTrack` should go away now, we don't need it

makesitgo (commented):
_this_ is the last bit now, i promise 😬

internal/telemetry/event.go
```diff
@@ -30,10 +30,10 @@ const (
 	EventTypeCommandError    EventType = "COMMAND_ERROR"
 )
 
-// EventDataKey is an event data key
-type EventDataKey string
-
 // set of event data keys
 const (
-	EventDataKeyErr EventDataKey = "err"
+	eventDataKeyCommand     string = "cmd"
```
makesitgo, line 35: [nit] shouldn't need the `string` types here (the `const` will pick that up automatically)

internal/telemetry/event.go
```diff
@@ -30,10 +30,10 @@ const (
...
+	eventDataKeyExecutionID string = "xid"
+
+	EventDataKeyErr string = "err"
```
makesitgo, line 38: [driveby fix] rename to `EventDataKeyError`

internal/telemetry/service_test.go
```diff
@@ -47,3 +82,20 @@ type testTracker struct {
 func (tracker *testTracker) Track(event event) {
 	tracker.lastTrackedEvent = event
 }
+func testServiceStdoutput(t *testing.T, createFn func() *Service, expected string) {
```
makesitgo, line 85: `Stdoutput`, while an excellent _portmanteau_, isn't exactly meaningful in this context... let's refactor this to be:

```go
func testServiceOutput(t *testing.T, service *Service, expectedOutput string) {
  t.Helper()

  // rest of code here
}
```

internal/telemetry/service_test.go
```diff
@@ -2,14 +2,49 @@ package telemetry
...
-		service := NewService(ModeOn, "userID", "command")
+		service := NewService(ModeStdout, "userID", nil, "command")
+
+		assert.Equal(t, "command", service.command)
+		assert.True(t, service.executionID != "", "service execution id must not be blank")
+		assert.Equal(t, "userID", service.userID)
+		assert.NotNil(t, service.tracker)
+	})
+
+	createSegmentFn := func() *Service {
```
makesitgo, line 23: to tack on with the `testServiceOutput` refactor I suggested below, I'll channel my inner-Tim here and resist the one-liner helper func.  instead, we can have:

```go
const (
  testUser = "userID"
  testCommand = "command"
)

var (
  testLogger = log.New(os.Stdout, "LogPrefix ", log.Lmsgprefix)
)

// and instead of passing `createSegmentFn` around here, just instantiate a `*Service` in the tests directly and use those

// and if you *really* want a helper for this still, then go with:
func newService(mode Mode) *Service {
  return NewService(mode, testUser, testLogger, testCommand) // then make sure to be using this for all `NewService` calls
}
```

internal/telemetry/service_test.go
```diff
@@ -2,14 +2,49 @@ package telemetry
...
+	t.Run("Should create a segment tracking service if the segmentWriteKey is there", func(t *testing.T) {
+		swk := segmentWriteKey
+		defer func() { segmentWriteKey = swk }()
+
+		segmentWriteKey = "testing"
+		testServiceStdoutput(t, createSegmentFn, "")
+
+		service := createSegmentFn()
+
+		assert.Equal(t, "command", service.command)
```
makesitgo, line 35: these asserts are already verified with `ModeStdout` ... we don't need them here as well (since the logic that handles setting these details is independent of whichever `mode` is passed in)

internal/telemetry/tracker_test.go
```diff
@@ -0,0 +1,104 @@
...
+				EventDataKeyErr:         "Something",
+				eventDataKeyCommand:     "someCommand",
+				eventDataKeyExecutionID: testExecutionID,
+			},
+		}
+		assert.Equal(t, []interface{}{expectedTrack}, client.calls)
+	})
+}
+
+func testTrackStdoutOutput(t *testing.T, tracker Tracker, event event, expected string) {
```
makesitgo, line 76: lets align this with the other similar helper here:

```go
func testTrackerOutput(t *testing.T, tracker Tracker, expectedOutput string) {
  t.Helper()

  // no need to pass `event` in here, just use a constant, test event and the expectedOutput should still match accordingly
}
```

internal/telemetry/tracker_test.go
```diff
@@ -0,0 +1,104 @@
...
+			Timestamp: testTime,
+			Event:     string(EventTypeCommandError),
+			Properties: map[string]interface{}{
+				EventDataKeyErr:         "Something",
+				eventDataKeyCommand:     "someCommand",
+				eventDataKeyExecutionID: testExecutionID,
+			},
+		}
+		assert.Equal(t, []interface{}{expectedTrack}, client.calls)
+	})
```
makesitgo, line 73: not sure why i missed this until now, but we want a test here to verify the error case works as expected as well

basically, the same approach you took with `testClient`, but instead of capturing `client.calls` you'll just return an error.

while you add this, lets also rework the names a bit to avoid any ambiguity/confusion.  the `testClient` you currently have should be come `mockSegmentClient`, and this new one can be `mockFailureSegmentClient`

makesitgo (commented):
just a bit more test clean up... and after the suggestions are up here we'll be good to go

makesitgo (approved):
🆒  beans

PR #83, opened 2021-02-25
REALMC-7177: Support static hosting

makesitgo:
I imagine a lot of the questions that come up during review here will be answered with: "b/c the old cli did it"

I'm all ears on whether this approach makes sense or needs to be tweaked/simplified


internal/cloud/realm/hosting.go
```diff
@@ -0,0 +1,213 @@
...
+
+	go func() {
+		defer func() {
+			bodyWriter.Close()
+			pipeWriter.Close()
+		}()
+		// Create the first part and write the metadata into it
+		mw, err := bodyWriter.CreateFormField(paramMetadata)
+		if err != nil {
+			errChan <- fmt.Errorf("failed to create metadata multipart field: %w", err)
```
tsedgwick, line 105: should we break on these errors as well?

internal/commands/push/command.go
```diff
@@ -80,26 +80,63 @@ func (cmd *Command) Handler(profile *cli.Profile, ui terminal.UI, clients cli.Cl
...
-	if len(diffs) == 0 {
+	if len(appDiffs) == 0 && !cmd.inputs.IncludeDependencies && hostingDiffs.Size() == 0 {
 		ui.Print(terminal.NewTextLog("Deployed app is identical to proposed version, nothing to do"))
 		return nil
 	}
 
 	if !ui.AutoConfirm() && !isNewApp {
+		diffs := make([]string, 0, len(appDiffs)+1+hostingDiffs.Size()+3)
+
+		diffs = append(diffs, appDiffs...)
```
tsedgwick, line 114: do we need to create a new slice here or can we just use appDiffs?

internal/commands/push/command.go
```diff
@@ -179,6 +214,46 @@ func (cmd *Command) Handler(profile *cli.Profile, ui terminal.UI, clients cli.Cl
...
+		}
+
+		if err := importHosting(); err != nil {
+			return err
+		}
+		ui.Print(terminal.NewTextLog("Import hosting assets"))
+
+		if cmd.inputs.ResetCDNCache {
+			s := spinner.New(terminal.SpinnerCircles, 250*time.Millisecond)
+			s.Suffix = " Resetting CDN cache..."
```
tsedgwick, line 241: thoughts on putting some context for the user on why this is necessary?

tsedgwick (approved):
LGTM overall

internal/commands/push/command.go
makesitgo: i didn't want to mutate `appDiffs` because we will check it further downstream when deciding if we need to do the `create draft` => `import` => `deploy app` flow (independent of the dependencies or hosting flows)

makesitgo: i didn't want to mutate `appDiffs` because we will check it further downstream when deciding if we need to do the `create draft` => `import` => `deploy app` flow (independent of the dependencies or hosting flows)

internal/commands/push/command.go
makesitgo: i could probably use that same context 😬 

I kept it short/naive here b/c the user has to explicitly include the flag to perform this action anyways... so theoretically they should have an idea/know what theyre doing

open to suggestions to add on here if you have them though

internal/cloud/realm/hosting.go
makesitgo: i was thinking that as well, but based on the current implementation it appears we don't do that, and i was worried about changing any existing behavior too much there.  it seemed almost intentional that we try to process all the files first, and just alert/warn on any errors

kpatel71716 (approved):
LGTM. Cool stuff!

PR #84, opened 2021-03-01
REALMC-7989: Include functions in new import request

makesitgo:
to be merged after [`REALMC-8296`](https://github.com/10gen/baas/pull/3865)

### DIFF
--- a/internal/cloud/realm/import_export_test.go
+++ b/internal/cloud/realm/import_export_test.go
@@ -216,11 +216,11 @@ func appDataV1(configVersion realm.AppConfigVersion, app realm.App) local.AppDat
 		Sync:                 map[string]interface{}{"development_mode_enabled": false},
 		CustomUserDataConfig: map[string]interface{}{"enabled": false},
 		Environments: map[string]map[string]interface{}{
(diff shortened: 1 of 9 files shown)


internal/cloud/realm/import_export_test.go
```diff
@@ -319,46 +319,40 @@ func appDataV2(app realm.App) local.AppDataV2 {
...
+		},
+		Triggers: []map[string]interface{}{
+			{
+				"name":          "yell",
+				"type":          "SCHEDULED",
+				"config":        map[string]interface{}{"schedule": "0 0 * * 1"},
+				"function_name": "test",
+				"disabled":      false,
+			},
+		},
```
kpatel71716, line 340: [minor q] Do database triggers have a significantly different structure as scheduled triggers (I imagine the type and config would be the differences here but not sure)? I wonder if it's worth testing one here as a sanity test

kpatel71716 (approved):
LGTM! Just have a super minor question

UnicodeSnowman (approved):

internal/cloud/realm/import_export_test.go
makesitgo: it was easy enough to add

PR #85, opened 2021-03-02
REALMC-8310: Include previous versions when releasing a new version

UnicodeSnowman:


contrib/release.js
```diff
@@ -0,0 +1,122 @@
...
+          url: `https://s3.amazonaws.com/realm-clis/${dir}/linux-amd64/realm-cli`,
+        },
+        'macos-amd64': {
+          url: `https://s3.amazonaws.com/realm-clis/${dir}/macos-amd64/realm-cli`,
+        },
+        'windows-amd64': {
+          url: `https://s3.amazonaws.com/realm-clis/${dir}/windows-amd64/realm-cli.exe`,
+        },
+      },
+      past_releases: [
```
makesitgo, line 63: does `past_releases` here want to match `pastReleases` below?

contrib/release.js
```diff
@@ -0,0 +1,122 @@
...
+    .toString()
+    .trim();
+
+  if (!dir) {
+    throw new Error(`error: failed to find release for git hash ${githash}`);
+  }
+
+  const currentData = JSON.parse(
+    child_process
+      .execSync(`aws s3 cp 's3://realm-clis/versions/cloud-prod/CURRENT' -`)
```
makesitgo, line 44: might be worth moving the `s3://...` address to its own const

contrib/release.js
```diff
@@ -0,0 +1,122 @@
...
+    updatedData
+  );
+
+  const response = await prompt('proceed? [y/n]');
+  if (response !== 'y') {
+    console.info('update canceled');
+    return;
+  }
+
+  const uploadResult = child_process
```
makesitgo, line 86: [nit] `uploadResult` => `uploadErr`?

contrib/release.js
```diff
@@ -0,0 +1,122 @@
...
+        baseDirectory: dir,
+      },
+      null,
+      '  '
+    )
+  );
+
+  console.info('updating npm version and creating tag...');
+  child_process.execSync(`npm version --no-git-tag-version ${VERSION}`);
+  child_process.execSync(`git add ./version.json ./package*`);
```
makesitgo, line 112: what is `./package*` here?  just the `package.json` & `package-lock.json` files?

makesitgo (approved):
just questions and small things, looks right by me though otherwise

🆒  beans

contrib/release.js
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const child_process = require('child_process');
```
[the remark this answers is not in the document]
UnicodeSnowman: 🤷 could be camel, I just mindlessly used the same name here, but happen to change for consistency

contrib/release.js
UnicodeSnowman: in general, all of our public facing API (and DB-level) naming for JSON fields should be snake_cased (we're bad/inconsistent about this... but that was the conclusion we came to a long time ago)

contrib/release.js
UnicodeSnowman: this isn't necessarily an err, this is just the actual bash output. if there is an actual error (i.e. a non zero exit code), the `execSync` call will throw, but there are also cases where the call is "successful" but something happens preventing the file data from being uploaded. in that case, we'll get some text output here.

contrib/release.js
UnicodeSnowman: yep, exactly. I didn't pay much attention to this, just preserving legacy behavior.

contrib/release.js
makesitgo: i should have been clearer with this comment, I'm fine with either naming style, but I'm guessing both `currentData` and `updatedData` want the same structure, right?

if so, either this field wants to change to `pastReleases` or the line below should update to use `...(currentData.past_releases || []),`

contrib/release.js
makesitgo: roger that, makes sense to me!

contrib/release.js
UnicodeSnowman: ohh yep! good catch!

PR #86, opened 2021-03-02
REALMC-8274: Support linking data lakes with --data-source flag in app create

nathantfrank:
Upon talking offline going the path of converting --data-source to --cluster and --data-lake.


internal/commands/app/create_test.go
```diff
@@ -230,222 +230,250 @@ func TestAppCreateHandler(t *testing.T) {
...
-				DeploymentModel: realm.DeploymentModelGlobal,
+			display := make([]string, 0, 10)
+			display = append(display, "01:23:45 UTC INFO  Successfully created app",
+				fmt.Sprintf("  Info                   "+spaceBuffer+fmtStr, "Details"),
+				"  ---------------------"+dashBuffer+"  "+strings.Repeat("-", dirLength),
+				fmt.Sprintf("  Client App ID          "+spaceBuffer+fmtStr, "test-app-abcde"),
+				"  Realm Directory        "+spaceBuffer+localApp.RootDir,
+				fmt.Sprintf("  Realm UI               "+spaceBuffer+fmtStr, "http://localhost:8080/groups/123/apps/456/dashboard"),
+			)
+			if tc.cluster != "" {
```
makesitgo, line 345: [sidenote] in general i'm not that big a fan of conditionals around asserts/tests, but with this filepath `display` logic, I think it's fine here for now

internal/commands/app/create_test.go
```diff
@@ -230,222 +230,250 @@ func TestAppCreateHandler(t *testing.T) {
...
-		assert.Nil(t, cmd.Handler(profile, ui, cli.Clients{Realm: rc, Atlas: ac}))
+			// TODO(REALMC-8262): Investigate file path display options
+			dirLength := len(localApp.RootDir)
+			fmtStr := fmt.Sprintf("%%-%ds", dirLength)
 
-		localApp, err := local.LoadApp(filepath.Join(profile.WorkingDirectory, cmd.inputs.Name))
-		assert.Nil(t, err)
+			var spaceBuffer, dashBuffer string
+			if tc.dataLake != "" {
+				spaceBuffer = "  "
```
makesitgo, line 333: if we ever can figure out how to make this output deterministic (and subsequently make asserts on this output a bit easier/more straightforward), I'd like to avoid these types of `var`s in tests, mostly b/c they tend to make understanding what's being asserted a bit more difficult to parse/understand

again, I think with that filepath `display` logic below this is fine for now, just thinking out loud here...

internal/commands/app/create_test.go
```diff
@@ -230,222 +230,250 @@ func TestAppCreateHandler(t *testing.T) {
...
-		profile, teardown := mock.NewProfileFromTmpDir(t, "app_create_test")
-		defer teardown()
-		profile.SetRealmBaseURL("http://localhost:8080")
+	for _, tc := range []struct {
+		description string
+		cluster     string
+		dataLake    string
+		atlasClient atlas.Client
+	}{
+		{
```
makesitgo, line 239: [question] do we (care to) test the case where neither `cluster` nor `dataLake` are provided here?

internal/commands/app/create_test.go
```diff
@@ -230,222 +230,250 @@ func TestAppCreateHandler(t *testing.T) {
...
-			inputs: createInputs{
-				newAppInputs: newAppInputs{
-					Name:            "test-app",
-					Project:         "123",
-					Location:        realm.LocationVirginia,
-					DeploymentModel: realm.DeploymentModelGlobal,
+			createDisplay: "Realm app based on the Realm app from-app",
+		},
+		{
+			description: "should create a minimal project dry run with cluster set",
```
makesitgo, line 388: [question] do we (care to) test the case where `dataLake` is set here?

internal/commands/app/create_test.go
```diff
@@ -230,222 +230,250 @@ func TestAppCreateHandler(t *testing.T) {
...
-				newAppInputs: newAppInputs{
-					Name:            "test-app",
-					Project:         "123",
-					Location:        realm.LocationVirginia,
-					DeploymentModel: realm.DeploymentModelGlobal,
-					From:            "from-app",
+			expectedDir := filepath.Join(profile.WorkingDirectory, "test-app")
+			display := make([]string, 0, 4)
+			display = append(display, fmt.Sprintf("01:23:45 UTC INFO  A %s would be created at "+expectedDir, tc.createDisplay))
+			if tc.cluster != "" {
```
makesitgo, line 426: considering we don't have to deal with any fixed width spacing issues here, i'm tempted to suggest using `expectedOutput func(rootDir string) string` in the `tc` struct (or just a separate test func all together) instead of using the `if` check here ... but "tomato tomato" as they say

internal/commands/app/create_test.go
```diff
@@ -230,222 +230,250 @@ func TestAppCreateHandler(t *testing.T) {
...
-		}
-
-		cmd := &CommandCreate{createInputs{newAppInputs: newAppInputs{
-			From:            "test-app",
-			Location:        realm.LocationVirginia,
-			DeploymentModel: realm.DeploymentModelGlobal,
-		}}}
+			expectedErr: errors.New("realm client error"),
+		},
+	} {
```
makesitgo, line 463: do we want test cases for when the `Clusters` and `DataLakes` calls fail?

makesitgo (commented):
just a few question and other random musings... nice work here!

internal/commands/app/create.go
```diff
@@ -66,23 +67,34 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
+	if cmd.inputs.Cluster != "" {
+		cs, err = cmd.inputs.resolveCluster(clients.Atlas, groupID)
+		if err != nil {
+			return err
+		}
+	}
+
+	var ds dataLakeService
+	if cmd.inputs.DataLake != "" {
+		ds, err = cmd.inputs.resolveDataLake(clients.Atlas, groupID)
```
kpatel71716, line 80: [megaminor q] I assume we are good linking both a cluster and data lake an app if someone provides both? I really like that tbh just wanted to make sure

internal/commands/app/create_inputs.go
```diff
@@ -16,12 +16,14 @@ import (
...
 	flagDirectoryUsage = "the directory to create your new Realm app, defaults to Realm app name"
 
-	flagDataSource      = "data-source"
-	flagDataSourceShort = "s"
-	flagDataSourceUsage = "include to link an Atlas cluster to your Realm app, defaults to first available"
+	flagCluster      = "cluster"
+	flagClusterUsage = "include to link an Atlas cluster to your Realm app"
+
+	flagDataLake      = "data-lake"
+	flagDataLakeUsage = "include to link an Atlas data lake to your Realm app"
```
kpatel71716, line 26: [thoght for future] I wonder if it would be nice to have datalake as an alias to the flag. I like this as is rn and it's not important, just a thought

internal/commands/app/create_inputs.go
```diff
@@ -101,29 +114,54 @@ func (i *createInputs) resolveDirectory(wd string) (string, error) {
...
+func (i *createInputs) resolveCluster(client atlas.Client, groupID string) (clusterService, error) {
 	clusters, err := client.Clusters(groupID)
 	if err != nil {
-		return dataSource{}, err
+		return clusterService{}, err
 	}
 	var clusterName string
 	for _, cluster := range clusters {
-		if i.DataSource == cluster.Name {
+		if i.Cluster == cluster.Name {
```
kpatel71716, line 124: [q] Do we care about casing here? It looks like in Atlas, the UI doesn't stop you from creating a cluster with the same name if the casing is different, but the cluster doesnt actually get created (so I assume this is a bug on the former). If we don't want to make any underlying assumptions, we can enforce casing but I wonder if we should note that somewhere since imagine devs could run into this often

internal/commands/app/create_inputs.go
```diff
@@ -101,29 +114,54 @@ func (i *createInputs) resolveDirectory(wd string) (string, error) {
...
+}
+
+func (i *createInputs) resolveDataLake(client atlas.Client, groupID string) (dataLakeService, error) {
+	dataLakes, err := client.DataLakes(groupID)
+	if err != nil {
+		return dataLakeService{}, err
+	}
+	var dataLakeName string
+	for _, dataLake := range dataLakes {
+		if i.DataLake == dataLake.Name {
```
kpatel71716, line 151: Same as other note about casing

kpatel71716 (commented):
Looks really good to me! Just have a few questions/comments

internal/commands/app/create.go
makesitgo: intentional behavior here, no reason we'd have to limit the user to either/or (come to think of it, we probably could support these as an array, but we'll leave that for a future release)

internal/commands/app/create_inputs.go
makesitgo: i don't love `--data-lake` either, so much so I went to google to see just how we actually refer to them in the wild.  the fact we seem to use "Data Lake" everywhere sadly seems to want us to keep using the hyphenated `--data-lake`

(we can use this as motivation to figure out auto-complete, so the flag just writes itself...)

internal/commands/app/create_inputs.go
makesitgo: I think for simplicity's sake (and for parity with the UI, broken or not) we should also just leave this check as-is.  good callout though, and something we may want to keep in mind as we get more usage with these flags

internal/commands/app/create_inputs.go
```diff
@@ -30,23 +32,34 @@ var (
...
+	Cluster   string
+	DataLake  string
+	DryRun    bool
 }
 
-type dataSource struct {
-	Name   string           `json:"name"`
-	Type   string           `json:"type"`
-	Config dataSourceConfig `json:"config"`
+type clusterService struct {
```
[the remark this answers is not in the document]
nathantfrank: Yeah, I went with the postfix `Service` (since that is what the internal team knows this as) and to add as a secondary differentiator this isn't a current pkg version of a `Cluster` or `DataLake`.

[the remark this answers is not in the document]
makesitgo: so obviously "service" is a bit of an overloaded term in our ecosystem ... I still tend to think of anything with that name as something that _does something_, rather than _is something_ (if that makes sense...)

again, feel free to leave as-is.  i see in the not-so-distant future this potentially moving into `local` package alongside the other app resources, but I don't want to go crazy with that stuff just yet

internal/commands/app/create_inputs_test.go
```diff
@@ -197,7 +197,7 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
 	})
 }
 
-func TestAppCreateInputsResolveDataSource(t *testing.T) {
+func TestAppCreateInputsResolveCluster(t *testing.T) {
```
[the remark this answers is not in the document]
nathantfrank: Yeah, for consistency and incase changes are made

internal/commands/app/create_test.go
nathantfrank: Agreed it's less than ideal. An Idea I just had is in the test we strip all repeating whitespace (except newlines) and dashes, then do the comparison? Since the terminal log component is handling the those two aspects which is being tested separately.

internal/commands/app/create_inputs.go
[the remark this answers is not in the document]
nathantfrank: I am going to go with the offline suggestions of dataSource* for clarity even though it is on the verbose side.

internal/commands/app/create_test.go
nathantfrank: Those cases are above and different enough not to be in the table

internal/commands/app/create_test.go
nathantfrank: I don't think it's needed in this particular case, since it is so similar to the dry run cluster case and creates the need for the conditional spacing (if using a test table), which has been muddying the tests.

internal/commands/app/create_inputs.go
```diff
@@ -91,6 +104,16 @@ func (i *createInputs) resolveDirectory(wd string) (string, error) {
 	if !fi.Mode().IsDir() {
 		return fullPath, nil
 	}
+	if i.From != "" && i.From == i.Directory {
```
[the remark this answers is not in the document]
makesitgo: also, I'd like to get this flow/logic in a place where a user could provide `--yes` and be able to still by-pass any and all prompts, what do you think about the following flow:

```cmd
# inform the user and prompt them with a confirm
$ 01:23:45 UTC WARN  The directory './new-app' already exists, writing app contents to that destination may result in file conflicts.
$ Would you still like to write app contents to './new-app'? ('No' will prompt you to provide another destination) [y/N]

# if 'yes', then proceed as usual
# if 'no', follow-up with a prompt to enter a new destination/directory
```

pros:
  * keeps us`-y`-able and gives CI environments a route to still write contents
  * lets users still create an app into an existing directory (maybe it's got non-realm stuff, a pre-initialized git repo or something along those lines..)

cons:
  * forces a two-step input from a user should they want to correct this problem (rather than just one-step in your current impl)

what do you think?

makesitgo (commented):
couple more thoughts here

internal/commands/app/create.go
```diff
@@ -61,28 +62,39 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
+	if cmd.inputs.DataLake != "" {
+		dsDataLake, err = cmd.inputs.resolveDataLake(clients.Atlas, groupID)
 		if err != nil {
 			return err
 		}
 	}
 
 	if cmd.inputs.DryRun {
-		logs := make([]terminal.Log, 0, 3)
+		logs := make([]terminal.Log, 0, 5)
```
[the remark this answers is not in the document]
nathantfrank: Yes, must have counted the if/else as 2

internal/commands/app/create_inputs.go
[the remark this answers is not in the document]
nathantfrank: > i thought we settled on prompting the user for a new `Directory` name to write to only if the directory to write to already existed? this check looks like it does otherwise, though I may be missing something here?

The single `if` statement is for the last part of the check, you are describing. 
On lines 97-99 we determine if the file points to a file or directory or is non-existent.
On line 104 we determine if the existent path points to a file.
Which leaves us at the state that we are looking at an existing folder.
On line 107 we are checking that the existing folder is the same name as the app we are basing the new app off of.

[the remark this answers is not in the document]
nathantfrank: > also, I'd like to get this flow/logic in a place where a user could provide `--yes` and be able to still by-pass any and all prompts, what do you think about the following flow:
> 
> ```batchfile
> # inform the user and prompt them with a confirm
> $ 01:23:45 UTC WARN  The directory './new-app' already exists, writing app contents to that destination may result in file conflicts.
> $ Would you still like to write app contents to './new-app'? ('No' will prompt you to provide another destination) [y/N]
> 
> # if 'yes', then proceed as usual
> # if 'no', follow-up with a prompt to enter a new destination/directory
> ```
> 
> pros:
> 
> * keeps us`-y`-able and gives CI environments a route to still write contents
> * lets users still create an app into an existing directory (maybe it's got non-realm stuff, a pre-initialized git repo or something along those lines..)
> 
> cons:
> 
> * forces a two-step input from a user should they want to correct this problem (rather than just one-step in your current impl)
> 
> what do you think?

I do like the supporting of non-interactive commands over being slightly more verbose in interactive mode.

internal/commands/app/create.go
```diff
@@ -134,14 +146,27 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
-		data, err := local.MarshalJSON(ds)
-		if err != nil {
-			return err
+		if dsCluster.Name != "" {
+			data, err := local.MarshalJSON(dsCluster)
+			if err != nil {
+				return err
+			}
+			path := filepath.Join(dataSourceDir, dsCluster.Name, local.FileConfig.String())
+			err = local.WriteFile(filepath.Join(dir, path), 0666, bytes.NewReader(data))
```
makesitgo, line 155: [no action needed here] if/when we ever need to formalize these data source definitions (read: move them to `local`), lets consider making a `DataSource` interface that implements a `Name() string` getter.  with that, we'd be able to have a single function here `func writeDataSourceConfig(dataSource local.DataSource)` (or in the `local` package itself)

again, this is good as is here, don't want to go type-crazy and this logic isn't changing much anytime soon.  just food for thought

internal/commands/app/create.go
```diff
@@ -194,9 +222,13 @@ func (cmd *CommandCreate) commandString(omitDryRun bool) string {
 		sb.WriteString(" --location ")
 		sb.WriteString(cmd.inputs.Location.String())
 	}
-	if cmd.inputs.DataSource != "" {
-		sb.WriteString(" --data-source ")
-		sb.WriteString(cmd.inputs.DataSource)
+	if cmd.inputs.Cluster != "" {
+		sb.WriteString(" --cluster ")
```
makesitgo, line 226: calling this out for the rebase, feel free to resolve afterwards

internal/commands/app/create_inputs.go
```diff
@@ -98,32 +111,71 @@ func (i *createInputs) resolveDirectory(wd string) (string, error) {
...
 		return "", errProjectExists{fullPath}
 	}
+	ui.Print(terminal.NewWarningLog("Directory './%s' already exists, writing app contents to that destination may result in file conflicts.", i.Directory))
+	proceed, err := ui.Confirm("Would you still like to write app contents to './%s'? ('No' will prompt you to provide another destination)", i.Directory)
+	if err != nil {
+		return "", err
+	}
+	if !proceed {
+		var newAppName string
+		if err := ui.AskOne(&newAppName, &survey.Input{Message: "App Name"}); err != nil {
```
makesitgo, line 121: whats the reasoning behind updating `Name` here too?  I'm not fully thinking this one through yet, but can we allow the user to create an app with one name but in a directory named something else?  It _feels_ like that should be ok, but we're firmly in the land of nuances and caveats here...

internal/commands/app/create_inputs_test.go
```diff
@@ -167,13 +167,43 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
+		assert.Nil(t, err)
+		defer console.Close()
+
+		doneCh := make(chan (struct{}))
+		go func() {
+			defer close(doneCh)
+
+			console.ExpectString("Directory './test-app' already exists, writing app contents to that destination may result in file conflicts.")
+			console.ExpectString("Would you still like to write app contents to './test-app'? ('No' will prompt you to provide another destination)")
+			console.SendLine("no")
```
makesitgo, line 191: [curious] does typing full "yes" actually trigger an affirmative confirm here?  i thought it was single characters before (y/n)...

internal/commands/app/create_inputs_test.go
```diff
@@ -167,13 +167,43 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
+		}()
+
+		inputs := createInputs{newAppInputs: newAppInputs{Name: "test-app"}}
+
+		err = os.Mkdir(path.Join(profile.WorkingDirectory, "test-app"), os.ModePerm)
+		assert.Nil(t, err)
+
+		dir, err := inputs.resolveDirectory(ui, profile.WorkingDirectory)
+		assert.Nil(t, err)
+		assert.Equal(t, path.Join(profile.WorkingDirectory, "new-app"), dir)
```
makesitgo, line 204: should probably also assert here on `inputs.Directory` (and `inputs.Name` if we keep that too)

internal/commands/app/create_test.go
makesitgo: i hear ya, but id prefer we have muddied waters here with more tests, than not.

can we break them out the differently-spaced tests into their own test funcs here (and live with copy/pasting that fixed-width logic for now)?  if/when we ever tackle that output to get it more deterministic, then we will see a lot of these test cases converge and we can re-table them at that point

internal/commands/app/create_test.go
makesitgo: considering we still don't even have product/design's sign off on this output style, lets avoid going to crazy trying to accommodate these tests.  lets live with the quick-and-dirty here, and see where the dust settles once people start seeing this in action

makesitgo (approved):
last bits here, but nothing crazy.  nice work!

🆒  beans

internal/commands/app/create.go
```diff
@@ -61,28 +62,39 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
+		logs := make([]terminal.Log, 0, 4)
 		if from.IsZero() {
 			logs = append(logs, terminal.NewTextLog("A minimal Realm app would be created at %s", dir))
 		} else {
 			logs = append(logs, terminal.NewTextLog("A Realm app based on the Realm app %s would be created at %s", cmd.inputs.From, dir))
 		}
-		if ds.Name != "" {
-			logs = append(logs, terminal.NewTextLog("The cluster %s would be linked as data source %s", cmd.inputs.DataSource, ds.Name))
+		if dsCluster.Name != "" {
+			logs = append(logs, terminal.NewTextLog("The cluster '%s' would be linked as data source '%s'", cmd.inputs.Cluster, dsCluster.Name))
```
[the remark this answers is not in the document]
nathantfrank: They are different, `cmd.inputs.Cluster` would be something like `Cluster0` while `dsCluster.Name` would be `mongodb-atlas`.

internal/commands/app/create_inputs.go
nathantfrank: Yeah I agree, was just trying to remove confusion, since a majority of the time directory == appName. If you think we should leave it alone, I'm good with that.

internal/commands/app/create_inputs_test.go
nathantfrank: It supports all capitalization variations of 'y', 'yes', 'n', and 'no'.

internal/commands/app/create_test.go
nathantfrank: I was wrong here, the dry run tests are much easier to table

internal/commands/app/create_inputs.go
nathantfrank: Going with only updating directory

PR #88, opened 2021-03-03
REALMC-7866: Implement command string

makesitgo:


internal/commands/commands.go
```diff
@@ -76,14 +76,14 @@ Realm app will be exported as well.`,
...
-a new disk. This command will create a new directory for your project. 
+				Help: `Creates a new Realm app in the cloud, and saves your configuration files on
+a new disk. This command will create a new directory for your project.
 To initialize a Realm app in your current working directory, use ‘app init’.
 
-You can specify a '--from' input to create a Realm app from 
-an existing app. If you do not specify a '--from' input, 
+You can specify a '--from' input to create a Realm app from
+an existing app. If you do not specify a '--from' input,
 the CLI will create a non-personalized Realm app.`,
```
nathantfrank, line 87: [Thought + Q] Should we not allow the users terminal determine the width/wrapping of Help <command> information at least for each block, instead of us following the golang character length?

internal/commands/app/create_inputs.go
```diff
@@ -127,3 +128,41 @@ func (i *createInputs) resolveDataSource(client atlas.Client, groupID string) (d
...
 	return dataSource, nil
 }
+
+func (i createInputs) args(omitDryRun bool) []flags.Arg {
+	args := make([]flags.Arg, 0, 8)
+
+	if i.Project != "" {
+		args = append(args, flags.Arg{flagProject, i.Project})
+	}
+
```
nathantfrank, line 138: [minor-nit] double newline between conditionals makes the func feel longer

internal/cli/command.go
```diff
@@ -80,3 +84,16 @@ type CommandDefinition struct {
 	// This value maps 1:1 to Cobra's `Aliases` property
 	Aliases []string
 }
+
+// CommandString prints the command string with the provided flags
```
nathantfrank, line 88: [minor] 
```suggestion
// CommandString returns the command string with the provided flags
```

internal/commands/app/create.go
```diff
@@ -167,40 +171,5 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
 }
 
 func (cmd *CommandCreate) commandString(omitDryRun bool) string {
```
nathantfrank, line 173: [pref]
```suggestion
func (cmd *CommandCreate) display(omitDryRun bool) string {
```

internal/cli/command.go
```diff
@@ -80,3 +84,16 @@ type CommandDefinition struct {
...
+func CommandString(cmd string, args []flags.Arg) string {
```
nathantfrank, line 89: [pref]
```suggestion
func DisplayCommand(cmd string, args []flags.Arg) string {
```

nathantfrank (commented):
Have some suggestions and a Q

internal/commands/commands.go
makesitgo: 80 is a pretty standard width number, not just go-specific.  but by doing it on our end, we ensure a more consistent output, and in the case of terminal displays that don't soft-wrap the output, it absolutely helps there.

going to leave the approach as is for now (i do think the help text is in line for an overhaul if we ever get the bandwidth to do so -- sumedha had some thoughts about tweaking it from the out-of-the-box style `cobra` gives us)

internal/cli/command.go
makesitgo: going with `CommandDisplay` here

nathantfrank (approved):
LGTM, once tests are passing

kpatel71716 (approved):
LGTM!

PR #92, opened 2021-03-04
REALMC-8324: Fix pull target path

andrewzlchen:
## pull command

When executing the pull command, I was getting a weird behavior:

I saw that the absolute path of the intended app pull was being appended to my current working directory

As a result, my file structure looked something like this:

image

## push -s command

Another issue that I ran into was that hosting doesn't seem to work. It expects to read files from the `hosting/` folder instead of the `hosting/files/` folder, so I made a quick fix here as well.

### Fixing pull

Within the doExport helper function, I saw that the `to` path was always being `filepath.Join()`-ed to the working directory, even if they shared a common ancestor.

The fix here is to first convert the `to` path, which was an absolute path, to be relative to the working directory, which was also an absolute path before appending the `to` path to the working dir

### Fixing push -s

Added `files/` to the path that is used to find the hosting assets files to be pushed.


andrewzlchen:
This is mostly just a quick PR to let you be aware of this problem. I didn't have time to test this because I am trying to put together a demo.

If you can wait, I can try to properly test this if needed, but otherwise, you can get the diff for this PR via 
[https://github.com/10gen/realm-cli/pull/92.diff](https://github.com/10gen/realm-cli/pull/92.diff)

andrewzlchen:
At the moment, hosting seems to work trivially well.

When adding a main.css that is linked to an index.html, modifying these files seems to work without any obvious problems.

I haven't tried folder structures that are nested or anything like that, but hopefully, this PR fixes any obvious issues.

PR #93, opened 2021-03-05
REALMC-8327: Fixup various bugs from demo

makesitgo:


internal/cli/inputs.go
```diff
@@ -46,21 +46,17 @@ func (i *ProjectInputs) Resolve(ui terminal.UI, wd string) error {
...
-
-	if len(questions) > 0 {
-		if err := ui.Ask(i, questions...); err != nil {
-			return err
+		var appOption string
+
+		if app.RootDir != "" {
+			appOption = app.Option()
+		} else {
+			if err := ui.AskOne(&appOption, &survey.Input{Message: "App Filter"}); err != nil {
```
makesitgo, line 55: this change will now only prompt the user for `App Filter` if the working directory does not have app information present.  if app details are present, then the app will be resolved accordingly (i.e. locking the app context in when running commands from a project directory)

internal/commands/app/create.go
```diff
@@ -105,9 +105,15 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
 		return nil
 	}
 
+	newApp, err := clients.Realm.CreateApp(groupID, cmd.inputs.Name, realm.AppMeta{cmd.inputs.Location, cmd.inputs.DeploymentModel})
```
makesitgo, line 108: driveby fix for https://jira.mongodb.org/browse/REALMC-8326 ... @nathan feel free to take this if we want saner commits

internal/commands/pull/command.go
```diff
@@ -144,6 +144,13 @@ func (cmd *Command) doExport(profile *cli.Profile, realmClient realm.Client, gro
 		to = name
 	}
 
+	if filepath.IsAbs(to) {
+		to, err = filepath.Rel(profile.WorkingDirectory, to)
```
makesitgo, line 148: cc: @andrew ... thanks for catching this!  including your fix from #92 here

internal/commands/pull/inputs.go
```diff
@@ -99,10 +99,6 @@ type from struct {
 func (i inputs) resolveFrom(ui terminal.UI, client realm.Client) (from, error) {
 	f := from{GroupID: i.Project}
 
-	if i.From == "" {
```
makesitgo, line 102: this caused us to pull from empty apps sometimes (if not run from within a project directory), by removing this we simply "resolve the app" with an empty app filter (which just presents the user with a larger list of apps to select from)

internal/commands/push/command.go
```diff
@@ -153,23 +153,25 @@ func (cmd *Command) Handler(profile *cli.Profile, ui terminal.UI, clients cli.Cl
...
 
-	ui.Print(terminal.NewTextLog("Creating draft"))
-	draft, proceed, err := createNewDraft(ui, clients.Realm, to)
-	if err != nil {
-		return err
-	}
-	if !proceed {
-		return nil
-	}
+	if len(appDiffs) > 0 {
```
makesitgo, line 156: driveby fix so we don't go through the draft/import/deploy cycle unless we need to

internal/local/hosting.go
```diff
@@ -208,10 +207,12 @@ func (h Hosting) UploadHostingAssets(realmClient realm.Client, groupID, appID st
 		}()
 	}
 
+	assetsDir := filepath.Join(h.RootDir, NameFiles)
```
makesitgo, line 210: cc: @andrew this is where can we ensure we use the right file path when uploading assets (we have a `const` for this `files/` path part in the `local` package ... can't use that in `realm` package though)

andrewzlchen: This works for me!

andrewzlchen (approved):
lgtm!

kpatel71716 (approved):
LGTM! Good stuff

PR #94, opened 2021-03-05
REALMC-7886: App Init should initialize an auth_providers directory w…

nathantfrank:
Just wanted to make sure we are on the same page before I wrap this one up.

### DIFF
--- a/internal/cloud/realm/import_export_test.go
+++ b/internal/cloud/realm/import_export_test.go
@@ -69,9 +69,9 @@ func TestRealmImportExportRoundTrip(t *testing.T) {
 			assert.Nil(t, local.WriteZip(wd, zipPkg))
 
 			t.Log("read the exported app data from disk")
-			localApp, localAppErr := local.LoadApp(wd)
-			assert.Nil(t, localAppErr)
-			assert.Equal(t, appData, localApp.AppData)
+			appLocal, appLocalErr := local.LoadApp(wd)
+			assert.Nil(t, appLocalErr)
+			assert.Equal(t, appData, appLocal.AppData)
 		})
 	}
 }
--- a/internal/commands/app/create.go
+++ b/internal/commands/app/create.go
@@ -1,9 +1,7 @@
 package app
 
 import (
-	"bytes"
 	"fmt"
-	"path/filepath"
 
 	"github.com/10gen/realm-cli/internal/cli"
 	"github.com/10gen/realm-cli/internal/cloud/realm"
@@ -34,6 +32,8 @@ func (cmd *CommandCreate) Flags(fs *pflag.FlagSet) {
 	fs.StringVar(&cmd.inputs.Cluster, flagCluster, "", flagClusterUsage)
 	fs.StringVar(&cmd.inputs.DataLake, flagDataLake, "", flagDataLakeUsage)
 	fs.BoolVarP(&cmd.inputs.DryRun, flagDryRun, flagDryRunShort, false, flagDryRunUsage)
+	fs.Var(&cmd.inputs.ConfigVersion, flagConfigVersion, flagConfigVersionUsage)
+	fs.MarkHidden(flagConfigVersion) //nolint: errcheck
 }
 
 // Inputs is the command inputs
@@ -105,23 +105,27 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients 
 		return nil
 	}
 
-	newApp, err := clients.Realm.CreateApp(groupID, cmd.inputs.Name, realm.AppMeta{cmd.inputs.Location, cmd.inputs.DeploymentModel})
+	appRealm, err := clients.Realm.CreateApp(groupID, cmd.inputs.Name, realm.AppMeta{cmd.inputs.Location, cmd.inputs.DeploymentModel})
 	if err != nil {
 		return err
 	}
 
+	var appLocal local.App
+
 	if appRemote.IsZero() {
-		appLocal := local.NewApp(
+		appLocal = local.NewApp(
 			dir,
-			newApp.ClientAppID,
+			appRealm.ClientAppID,
 			cmd.inputs.Name,
 			cmd.inputs.Location,
 			cmd.inputs.DeploymentModel,
+			cmd.inputs.ConfigVersion,
 		)
-		err := appLocal.WriteConfig()
-		if err != nil {
-			return err
-		}
+		local.AddAuthProvider(appLocal.AppData, "api-key", map[string]interface{}{
+			"name":     "api-key",
+			"type":     "api-key",
+			"disabled": true,
+		})
 	} else {
 		_, zipPkg, err := clients.Realm.Export(
 			appRemote.GroupID,
@@ -134,59 +138,50 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients 
 		if err := local.WriteZip(dir, zipPkg); err != nil {
 			return err
 		}
-	}
-
-	loadedApp, err := local.LoadApp(dir)
-	if err != nil {
-		return err
-	}
-
-	if dsCluster.Name != "" || dsDataLake.Name != "" {
-		var dataSourceDir string
-		switch loadedApp.ConfigVersion() {
-		case
-			realm.AppConfigVersion20200603,
-			realm.AppConfigVersion20180301:
-			dataSourceDir = local.NameServices
-		default:
-			dataSourceDir = local.NameDataSources
-		}
-		if dsCluster.Name != "" {
-			data, err := local.MarshalJSON(dsCluster)
-			if err != nil {
-				return err
-			}
-			path := filepath.Join(dataSourceDir, dsCluster.Name, local.FileConfig.String())
-			err = local.WriteFile(filepath.Join(dir, path), 0666, bytes.NewReader(data))
-			if err != nil {
-				return err
-			}
-		}
-		if dsDataLake.Name != "" {
-			data, err := local.MarshalJSON(dsDataLake)
-			if err != nil {
-				return err
-			}
-			path := filepath.Join(dataSourceDir, dsDataLake.Name, local.FileConfig.String())
-			err = local.WriteFile(filepath.Join(dir, path), 0666, bytes.NewReader(data))
-			if err != nil {
-				return err
-			}
-		}
-		if err = loadedApp.Load(); err != nil {
+		appLocal, err = local.LoadApp(dir)
+		if err != nil {
 			return err
 		}
 	}
 
-	if err := clients.Realm.Import(newApp.GroupID, newApp.ID, loadedApp.AppData); err != nil {
+	if dsCluster.Name != "" {
+		local.AddDataSource(appLocal.AppData, map[string]interface{}{
+			"name": dsCluster.Name,
+			"type": dsCluster.Type,
+			"config": map[string]interface{}{
+				"clusterName":         dsCluster.Config.ClusterName,
+				"readPreference":      dsCluster.Config.ReadPreference,
+				"wireProtocolEnabled": dsCluster.Config.WireProtocolEnabled,
+			},
+		})
+	}
+	if dsDataLake.Name != "" {
+		local.AddDataSource(appLocal.AppData, map[string]interface{}{
+			"name": dsDataLake.Name,
+			"type": dsDataLake.Type,
+			"config": map[string]interface{}{
+				"dataLakeName": dsDataLake.Config.DataLakeName,
+			},
+		})
+	}
+
+	if err = appLocal.Write(); err != nil {
+		return err
+	}
+
+	if err = appLocal.Load(); err != nil {
+		return err
+	}
+
+	if err := clients.Realm.Import(appRealm.GroupID, appRealm.ID, appLocal.AppData); err != nil {
 		return err
 	}
 
 	headers := []string{"Info", "Details"}
 	rows := make([]map[string]interface{}, 0, 5)
-	rows = append(rows, map[string]interface{}{"Info": "Client App ID", "Details": newApp.ClientAppID})
+	rows = append(rows, map[string]interface{}{"Info": "Client App ID", "Details": appRealm.ClientAppID})
 	rows = append(rows, map[string]interface{}{"Info": "Realm Directory", "Details": dir})
-	rows = append(rows, map[string]interface{}{"Info": "Realm UI", "Details": fmt.Sprintf("%s/groups/%s/apps/%s/dashboard", profile.RealmBaseURL(), newApp.GroupID, newApp.ID)})
+	rows = append(rows, map[string]interface{}{"Info": "Realm UI", "Details": fmt.Sprintf("%s/groups/%s/apps/%s/dashboard", profile.RealmBaseURL(), appRealm.GroupID, appRealm.ID)})
 	if dsCluster.Name != "" {
 		rows = append(rows, map[string]interface{}{"Info": "Data Source (Cluster)", "Details": dsCluster.Name})
 	}
--- a/internal/commands/app/create_test.go
+++ b/internal/commands/app/create_test.go
@@ -4,6 +4,8 @@ import (
 	"archive/zip"
 	"errors"
 	"fmt"
+	"io/ioutil"
+	"os"
 	"path/filepath"
 	"strings"
 	"testing"
@@ -45,11 +47,14 @@ func TestAppCreateHandler(t *testing.T) {
 			Project:         "123",
 			Location:        realm.LocationVirginia,
 			DeploymentModel: realm.DeploymentModelGlobal,
+			ConfigVersion:   realm.DefaultAppConfigVersion,
 		}}}
 
 		assert.Nil(t, cmd.Handler(profile, ui, cli.Clients{Realm: client}))
 
-		appLocal, err := local.LoadApp(filepath.Join(profile.WorkingDirectory, cmd.inputs.Name))
+		fullDir := filepath.Join(profile.WorkingDirectory, cmd.inputs.Name)
+
+		appLocal, err := local.LoadApp(fullDir)
 		assert.Nil(t, err)
 
 		assert.Equal(t, &local.AppRealmConfigJSON{local.AppDataV2{local.AppStructureV2{
@@ -58,6 +63,45 @@ func TestAppCreateHandler(t *testing.T) {
 			Name:            "test-app",
 			Location:        realm.LocationVirginia,
 			DeploymentModel: realm.DeploymentModelGlobal,
+			Environments: map[string]map[string]interface{}{
+				"development.json": {
+					"values": map[string]interface{}{},
+				},
+				"no-environment.json": {
+					"values": map[string]interface{}{},
+				},
+				"production.json": {
+					"values": map[string]interface{}{},
+				},
+				"qa.json": {
+					"values": map[string]interface{}{},
+				},
+				"testing.json": {
+					"values": map[string]interface{}{},
+				},
+			},
+			Auth: &local.AuthStructure{
+				CustomUserData: map[string]interface{}{"enabled": false},
+				Providers: map[string]interface{}{
+					"api-key": map[string]interface{}{
+						"name":     "api-key",
+						"type":     "api-key",
+						"disabled": true,
+					},
+				},
+			},
+			Sync: &local.SyncStructure{Config: map[string]interface{}{"development_mode_enabled": false}},
+			Functions: &local.FunctionsStructure{
+				Configs: []map[string]interface{}{},
+				Sources: map[string]string{},
+			},
+			GraphQL: &local.GraphQLStructure{
+				Config: map[string]interface{}{
+					"use_natural_pluralization": true,
+				},
+				CustomResolvers: []map[string]interface{}{},
+			},
+			Values: []map[string]interface{}{},
 		}}}, appLocal.AppData)
 
 		// TODO(REALMC-8262): Investigate file path display options
@@ -74,6 +118,77 @@ func TestAppCreateHandler(t *testing.T) {
 			"01:23:45 UTC DEBUG Check out your app: cd ./test-app && realm-cli app describe",
 			"",
 		}, "\n"), out.String())
+
+		t.Run("should have the expected contents in the auth custom user data file", func(t *testing.T) {
+			config, err := ioutil.ReadFile(filepath.Join(fullDir, local.NameAuth, local.FileCustomUserData.String()))
+			assert.Nil(t, err)
+			assert.Equal(t, `{
+    "enabled": false
+}
+`, string(config))
+		})
+
+		t.Run("should have the expected contents in the auth providers file", func(t *testing.T) {
+			config, err := ioutil.ReadFile(filepath.Join(fullDir, local.NameAuth, local.FileProviders.String()))
+			assert.Nil(t, err)
+			assert.Equal(t, `{
+    "api-key": {
+        "disabled": true,
+        "name": "api-key",
+        "type": "api-key"
+    }
+}
+`, string(config))
+		})
+
+		t.Run("should have data sources directory", func(t *testing.T) {
+			_, err := os.Stat(filepath.Join(fullDir, local.NameDataSources))
+			assert.Nil(t, err)
+		})
+
+		t.Run("should have the expected contents in the functions config file", func(t *testing.T) {
+			config, err := ioutil.ReadFile(filepath.Join(fullDir, local.NameFunctions, local.FileConfig.String()))
+			assert.Nil(t, err)
+			assert.Equal(t, "[]\n", string(config))
+		})
+
+		t.Run("should have graphql custom resolvers directory", func(t *testing.T) {
+			_, err := os.Stat(filepath.Join(fullDir, local.NameGraphQL, local.NameCustomResolvers))
+			assert.Nil(t, err)
+		})
+
+		t.Run("should have the expected contents in the graphql config file", func(t *testing.T) {
+			config, err := ioutil.ReadFile(filepath.Join(fullDir, local.NameGraphQL, local.FileConfig.String()))
+			assert.Nil(t, err)
+			assert.Equal(t, `{
+    "use_natural_pluralization": true
+}
+`, string(config))
+		})
+
+		t.Run("should have http endpoints directory", func(t *testing.T) {
+			_, err := os.Stat(filepath.Join(fullDir, local.NameHTTPEndpoints))
+			assert.Nil(t, err)
+		})
+
+		t.Run("should have services directory", func(t *testing.T) {
+			_, err := os.Stat(filepath.Join(fullDir, local.NameServices))
+			assert.Nil(t, err)
+		})
+
+		t.Run("should have the expected contents in the sync config file", func(t *testing.T) {
+			config, err := ioutil.ReadFile(filepath.Join(fullDir, local.NameSync, local.FileConfig.String()))
+			assert.Nil(t, err)
+			assert.Equal(t, `{
+    "development_mode_enabled": false
+}
+`, string(config))
+		})
+
+		t.Run("should have values directory", func(t *testing.T) {
+			_, err := os.Stat(filepath.Join(fullDir, local.NameValues))
+			assert.Nil(t, err)
+		})
 	})
 
 	t.Run("when remote and project are not set should create minimal project and prompt for project", func(t *testing.T) {
@@ -124,6 +239,7 @@ func TestAppCreateHandler(t *testing.T) {
 			Project:         "123",
 			Location:        realm.LocationVirginia,
 			DeploymentModel: realm.DeploymentModelGlobal,
+			ConfigVersion:   realm.DefaultAppConfigVersion,
 		}}}
 
 		assert.Nil(t, cmd.Handler(profile, ui, cli.Clients{Realm: rc, Atlas: ac}))
@@ -134,15 +250,53 @@ func TestAppCreateHandler(t *testing.T) {
 		appLocal, err := local.LoadApp(filepath.Join(profile.WorkingDirectory, cmd.inputs.Name))
 		assert.Nil(t, err)
 
-		expectedAppData := local.AppRealmConfigJSON{local.AppDataV2{local.AppStructureV2{
+		assert.Equal(t, &local.AppRealmConfigJSON{local.AppDataV2{local.AppStructureV2{
 			ConfigVersion:   realm.DefaultAppConfigVersion,
 			ID:              "test-app-abcde",
 			Name:            "test-app",
 			Location:        realm.LocationVirginia,
 			DeploymentModel: realm.DeploymentModelGlobal,
-		}}}
+			Environments: map[string]map[string]interface{}{
+				"development.json": {
+					"values": map[string]interface{}{},
+				},
+				"no-environment.json": {
+					"values": map[string]interface{}{},
+				},
+				"production.json": {
+					"values": map[string]interface{}{},
+				},
+				"qa.json": {
+					"values": map[string]interface{}{},
+				},
+				"testing.json": {
+					"values": map[string]interface{}{},
+				},
+			},
+			Auth: &local.AuthStructure{
+				CustomUserData: map[string]interface{}{"enabled": false},
+				Providers: map[string]interface{}{
+					"api-key": map[string]interface{}{
+						"name":     "api-key",
+						"type":     "api-key",
+						"disabled": true,
+					},
+				},
+			},
+			Sync: &local.SyncStructure{Config: map[string]interface{}{"development_mode_enabled": false}},
+			Functions: &local.FunctionsStructure{
+				Configs: []map[string]interface{}{},
+				Sources: map[string]string{},
+			},
+			GraphQL: &local.GraphQLStructure{
+				Config: map[string]interface{}{
+					"use_natural_pluralization": true,
+				},
+				CustomResolvers: []map[string]interface{}{},
+			},
+			Values: []map[string]interface{}{},
+		}}}, appLocal.AppData)
 
-		assert.Equal(t, &expectedAppData, appLocal.AppData)
 		assert.Equal(t, realm.App{
 			ID:          "456",
 			GroupID:     "123",
@@ -214,6 +368,14 @@ func TestAppCreateHandler(t *testing.T) {
 				Providers:      map[string]interface{}{},
 			},
 			Sync: &local.SyncStructure{Config: map[string]interface{}{"development_mode_enabled": false}},
+			Functions: &local.FunctionsStructure{
+				Configs: []map[string]interface{}{},
+				Sources: map[string]string{},
+			},
+			GraphQL: &local.GraphQLStructure{
+				CustomResolvers: []map[string]interface{}{},
+			},
+			Values: []map[string]interface{}{},
 		}}}, appLocal.AppData)
 
 		// TODO(REALMC-8262): Investigate file path display options
@@ -311,10 +473,10 @@ func TestAppCreateHandler(t *testing.T) {
 
 			assert.Nil(t, cmd.Handler(profile, ui, cli.Clients{Realm: rc, Atlas: tc.atlasClient}))
 
-			localApp, err := local.LoadApp(filepath.Join(profile.WorkingDirectory, cmd.inputs.Name))
+			appLocal, err := local.LoadApp(filepath.Join(profile.WorkingDirectory, cmd.inputs.Name))
 			assert.Nil(t, err)
 
-			assert.Equal(t, importAppData, localApp.AppData)
+			assert.Equal(t, importAppData, appLocal.AppData)
 			assert.Equal(t, realm.App{
 				ID:          "456",
 				GroupID:     "123",
@@ -327,7 +489,7 @@ func TestAppCreateHandler(t *testing.T) {
 			}, createdApp)
 
 			// TODO(REALMC-8262): Investigate file path display options
-			dirLength := len(localApp.RootDir)
+			dirLength := len(appLocal.RootDir)
 			fmtStr := fmt.Sprintf("%%-%ds", dirLength)
 
 			var spaceBuffer, dashBuffer string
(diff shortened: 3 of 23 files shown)


internal/commands/app/init_inputs.go
```diff
@@ -34,6 +40,9 @@ func (i *initInputs) Resolve(profile *cli.Profile, ui terminal.UI) error {
 		if i.Location == realm.LocationEmpty {
 			i.Location = flagLocationDefault
 		}
+		if i.ConfigVersion == realm.AppConfigVersionZero {
+			i.ConfigVersion = realm.AppConfigVersion20210101
```
makesitgo, line 44: could use `DefaultAppConfigVersion` here

internal/commands/app/init.go
```diff
@@ -50,16 +55,77 @@ func (cmd *CommandInit) Handler(profile *cli.Profile, ui terminal.UI, clients cl
...
+		}
+		err = local.WriteFile(filepath.Join(wd, "auth_providers", "api-key.json"), 0666, bytes.NewReader(data))
+		if err != nil {
+			return err
+		}
+		if err := os.MkdirAll(filepath.Join(wd, "functions"), os.ModePerm); err != nil {
+			return err
+		}
+	default:
+		err := local.WriteFile(filepath.Join(wd, "auth", "providers.json"), 0666, bytes.NewReader([]byte("{}")))
```
makesitgo, line 79: I'd say this looks like roughly the right idea (didn't look too closely at each file being written), but I wonder if it makes sense to try and push this "writing" logic to `(local.App).Write() error`?  I think no matter what this is going to be verbose, but it seems nice to tuck it away behind the `local.App` struct, i'm hoping it makes it a bit clearer in this spot of what the app structure "looks" like per config version

internal/commands/app/init.go
```diff
@@ -21,6 +25,7 @@ func (cmd *CommandInit) Flags(fs *pflag.FlagSet) {
 	fs.StringVarP(&cmd.inputs.Name, flagName, flagNameShort, "", flagNameUsage)
 	fs.VarP(&cmd.inputs.DeploymentModel, flagDeploymentModel, flagDeploymentModelShort, flagDeploymentModelUsage)
 	fs.VarP(&cmd.inputs.Location, flagLocation, flagLocationShort, flagLocationUsage)
+	fs.Var(&cmd.inputs.ConfigVersion, flagConfigVersion, flagConfigVersionUsage)
```
makesitgo, line 28: if we're going to add this, let's keep it consistent b/w `app init` and `app create` and also figure out how to make these "hidden" flags (cobra supports some notion of that, i believe)

makesitgo (commented):
just a few first thoughts, but looks right

internal/commands/app/create_test.go
```diff
@@ -45,19 +47,107 @@ func TestAppCreateHandler(t *testing.T) {
...
+		// 		Configs: []map[string]interface{}{},
+		// 		Sources: map[string]string{},
+		// 	},
+		// 	GraphQL: &local.GraphQLStructure{
+		// 		Config: map[string]interface{}{
+		// 			"use_natural_pluralization": true,
+		// 		},
+		// 		CustomResolvers: []map[string]interface{}{},
+		// 	},
+		// }}}, localApp.AppData)
```
nathantfrank, line 150: Need a second pair of eyes on this

internal/commands/app/new_app_inputs.go
```diff
@@ -26,6 +26,9 @@ const (
 	flagLocationShort   = "l"
 	flagLocationUsage   = `select the Realm app's location, available options: ["US-VA", "local"]`
 	flagLocationDefault = realm.LocationVirginia
+
+	flagConfigVersion      = "config-version"
+	flagConfigVersionUsage = "the config version and app structure to initialize your Realm App with; defaults to 20210101"
```
makesitgo, line 31:
```suggestion
	flagConfigVersionUsage = "the config version of the Realm app structure; defaults to latest stable config version"
```

internal/local/app.go
```diff
@@ -37,30 +38,139 @@ func (a App) Option() string {
 }
 
 // NewApp returns a new local app
-func NewApp(rootDir, clientAppID, name string, location realm.Location, deploymentModel realm.DeploymentModel) App {
+func NewApp(rootDir, clientAppID, name string, location realm.Location, deploymentModel realm.DeploymentModel, cv realm.AppConfigVersion) App {
```
makesitgo, line 41: [nit] `cv` => `configVersion`

internal/local/app.go
```diff
@@ -37,30 +38,139 @@ func (a App) Option() string {
...
 }
 
 // AsApp converts the realm.App into a local app
-func AsApp(rootDir string, app realm.App) App {
-	return App{
-		RootDir: rootDir,
-		Config:  FileRealmConfig,
-		AppData: &AppRealmConfigJSON{AppDataV2{AppStructureV2{
-			ConfigVersion:   realm.AppConfigVersion20210101,
+func AsApp(rootDir string, app realm.App, cv realm.AppConfigVersion) App {
```
makesitgo, line 53: [nit] `cv` => `configVersion`

internal/local/app.go
```diff
@@ -37,30 +38,139 @@ func (a App) Option() string {
...
+		}}}
+	}
+	return App{
+		RootDir: rootDir,
+		Config:  FileRealmConfig,
+		AppData: appData,
+	}
+}
+
+// Write writes the app skeleton including config to disk
```
makesitgo, line 88:
```suggestion
// Write writes the app data to disk
```

internal/local/app_test.go
```diff
@@ -278,6 +279,245 @@ func TestAppWriteConfig(t *testing.T) {
...
+
+		var config AppStitchJSON
+		assert.Nil(t, json.Unmarshal(data, &config))
+		assert.Equal(t, AppStitchJSON{AppDataV1{AppStructureV1{
+			ConfigVersion:   realm.AppConfigVersion20180301,
+			ID:              "test-app-abcde",
+			Name:            "test-app",
+			Location:        realm.LocationIreland,
+			DeploymentModel: realm.DeploymentModelLocal,
+			// CustomUserDataConfig: map[string]interface{}{
```
makesitgo, line 303: comments still needed?

makesitgo (commented):
first pass!

internal/local/app_test.go
nathantfrank: nope

internal/local/app_test.go
```diff
@@ -278,6 +279,245 @@ func TestAppWriteConfig(t *testing.T) {
...
+
+		var config AppConfigJSON
+		assert.Nil(t, json.Unmarshal(data, &config))
+		assert.Equal(t, AppConfigJSON{AppDataV1{AppStructureV1{
+			ConfigVersion:   realm.AppConfigVersion20180301,
+			ID:              "test-app-abcde",
+			Name:            "test-app",
+			Location:        realm.LocationIreland,
+			DeploymentModel: realm.DeploymentModelLocal,
+			// CustomUserDataConfig: map[string]interface{}{
```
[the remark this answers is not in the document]
nathantfrank: nope

nathantfrank:
@makesitgo I'm writing a few more tests

internal/cloud/realm/import_export_test.go
```diff
@@ -246,7 +246,7 @@ console.log('got heem!');
 				"disabled":      false,
 			},
 		},
-		GraphQL: local.GraphQLStructure{
+		GraphQL: &local.GraphQLStructure{
```
nathantfrank, line 249: Made `V1.GraphQL` consistent with all other `*Structure`s.

nathantfrank (commented):
@makesitgo Ready for another look!

internal/cloud/realm/import_export_test.go
makesitgo: the pointer thing isn't supposed to apply everywhere... that was more when it was needed to understand whether something was present or not, and with v1 graphql was _always present

you can revert this

internal/local/structure_v1_test.go
```diff
@@ -0,0 +1,104 @@
...
+			{
+				"config": map[string]interface{}{
+					"name":    "test",
+					"private": true,
+				},
+				"source": "exports = function(){\n  console.log('Hello World!');\n};",
+			},
+		}
+
+		writeFunctionsV1(tmpDir, data)
```
makesitgo, line 28: should check the error here

internal/local/structure_v1_test.go
```diff
@@ -0,0 +1,104 @@
...
+	t.Run("should write auth to disk", func(t *testing.T) {
+		data := []map[string]interface{}{
+			{
+				"name":     "api-key",
+				"type":     "api-key",
+				"disabled": true,
+			},
+		}
+
+		writeAuthProviders(tmpDir, data)
```
makesitgo, line 60: and check the error here

internal/local/structure_v1_test.go
```diff
@@ -0,0 +1,104 @@
...
+			{
+				"name":          "yell",
+				"type":          "SCHEDULED",
+				"config":        map[string]interface{}{"schedule": "0 0 * * 1"},
+				"function_name": "test",
+				"disabled":      false,
+			},
+		}
+
+		writeTriggers(tmpDir, data)
```
makesitgo, line 89: 😬

internal/local/structure_v2.go
```diff
@@ -331,3 +328,228 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+	return nil
+}
+
+func writeAuth(rootDir string, auth *AuthStructure) error {
+	if auth != nil && auth.Providers != nil {
+		dataProvider, err := MarshalJSON(auth.Providers)
+		if err != nil {
+			return err
+		}
+		dataProvider = append(dataProvider, "\n"...)
```
makesitgo, line 423: question, do we not need the newline for other outputs from `MarshalJSON`?  i don't see it _everywhere_, is why i ask.

if we do need it, lets just add that to the `MarshalJSON` implementation.  alternatively, i know `json.NewDecoder(...).Decode(..)` will automatically append that newline, so maybe we just switch to using that?

internal/local/structure_v2.go
```diff
@@ -331,3 +328,228 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+		dataCustomUser, err := MarshalJSON(auth.CustomUserData)
+		if err != nil {
+			return err
+		}
+		dataCustomUser = append(dataCustomUser, "\n"...)
+		err = WriteFile(filepath.Join(rootDir, NameAuth, FileCustomUserData.String()), 0666, bytes.NewReader(dataCustomUser))
+		if err != nil {
+			return err
+		}
+	} else {
```
makesitgo, line 444: just to retiterate, ideally these `else` blocks go away, and we simply just write what the app has

and now that i see this more and more, maybe it makes sense to instead of pushing that logic to `app init`/`app create`, but rather to the `NewApp`/`AsApp` constructors?

makesitgo (commented):
ok i really like where is at / where this is headed.

last major bit here is probably figuring out where this initialization code goes.  i'm liking the idea we build these app data structures out in the App constructor functions, rather than jam in logic to the `write*` functions and/or duplicate it b/w `app init`/`app create`.  thoughts?

internal/commands/app/create.go
```diff
@@ -134,59 +138,49 @@ func (cmd *CommandCreate) Handler(profile *cli.Profile, ui terminal.UI, clients
...
-	loadedApp, err := local.LoadApp(dir)
-	if err != nil {
-		return err
+		appLocal, err = local.LoadApp(dir)
+		if err != nil {
+			return err
+		}
 	}
 
 	if dsCluster.Name != "" || dsDataLake.Name != "" {
```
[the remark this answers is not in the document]
nathantfrank: totally is now!

internal/commands/app/create_test.go
```diff
@@ -311,10 +497,51 @@ func TestAppCreateHandler(t *testing.T) {
 
 			assert.Nil(t, cmd.Handler(profile, ui, cli.Clients{Realm: rc, Atlas: tc.atlasClient}))
 
-			localApp, err := local.LoadApp(filepath.Join(profile.WorkingDirectory, cmd.inputs.Name))
+			appLocal, err := local.LoadApp(filepath.Join(profile.WorkingDirectory, cmd.inputs.Name))
 			assert.Nil(t, err)
 
-			assert.Equal(t, importAppData, localApp.AppData)
+			ad := importAppData.(*local.AppRealmConfigJSON)
```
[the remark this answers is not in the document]
nathantfrank: I think I fixed this test, when I shoulda taken a breather. Looking back now I needed to `appLocal.Load()` after we write to disk.

internal/local/structure.go
```diff
@@ -230,3 +232,149 @@ func parseServices(rootDir string) ([]ServiceStructure, error) {
 
 	return out, nil
 }
+
+func writeSecrets(rootDir string, secrets *SecretsStructure) error {
+	if secrets != nil {
```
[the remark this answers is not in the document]
nathantfrank: Only works on writeSecrets all others have else statements.

internal/local/structure_v2.go
nathantfrank: Yeah, I'm gonna go in this direction. I didn't like putting defaults into the cmd (since they weren't cmd specific defaults)

internal/local/structure.go
```diff
@@ -230,3 +232,149 @@ func parseServices(rootDir string) ([]ServiceStructure, error) {
...
+			}
+			data = append(data, "\n"...)
+			nameFile := fmt.Sprintf("%s_%s%s", strings.ToLower(vals["on_type"].(string)), strings.ToLower(vals["field_name"].(string)), extJSON)
+			err = WriteFile(filepath.Join(rootDir, NameGraphQL, NameCustomResolvers, nameFile), 0666, bytes.NewReader(data))
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		err := WriteFile(filepath.Join(rootDir, NameGraphQL, FileConfig.String()), 0666, bytes.NewReader([]byte("{\n    \"use_natural_pluralization\": true\n}\n")))
```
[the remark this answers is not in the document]
nathantfrank: Went with suggestion made in `structure_v2`.

internal/local/structure.go
```diff
@@ -230,3 +232,149 @@ func parseServices(rootDir string) ([]ServiceStructure, error) {
...
+				return err
+			}
+			data = append(data, "\n"...)
+			err = WriteFile(filepath.Join(rootDir, NameEnvironments, env), 0666, bytes.NewReader(data))
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		for _, fileName := range []string{"development", "no-environment", "production", "qa", "testing"} {
```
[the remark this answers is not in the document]
nathantfrank: Went with suggestion made in structure_v2.

internal/local/app_test.go
```diff
@@ -278,6 +275,215 @@ func TestAppWriteConfig(t *testing.T) {
...
+		})
+
+		t.Run("should have functions directory", func(t *testing.T) {
+			_, err := os.Stat(filepath.Join(tmpDir, NameFunctions))
+			assert.True(t, !os.IsNotExist(err), "%s should exist", NameFunctions)
+		})
+
+		t.Run("should have graphql custom resolvers directory", func(t *testing.T) {
+			_, err := os.Stat(filepath.Join(tmpDir, NameGraphQL, NameCustomResolvers))
+			assert.True(t, !os.IsNotExist(err), "%s should exist", filepath.Join(NameGraphQL, NameCustomResolvers))
```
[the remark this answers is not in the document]
nathantfrank: Yes, idk why I thought I needed to check the error.

internal/local/structure.go
```diff
@@ -230,3 +232,118 @@ func parseServices(rootDir string) ([]ServiceStructure, error) {
...
+		if err != nil {
+			return err
+		}
+	}
+	for _, vals := range graphql.CustomResolvers {
+		data, err := MarshalJSON(vals)
+		if err != nil {
+			return err
+		}
+		nameFile := fmt.Sprintf("%s_%s%s", strings.ToLower(vals["on_type"].(string)), strings.ToLower(vals["field_name"].(string)), extJSON)
```
makesitgo, line 297: we should avoid the possibility of panics with unchecked `.(string)` casts here (even though our data model suggests that'll never happen) by wrapping the whole `fmt.Sprintf` with `strings.ToLower`

also, its hard for me not to just suggest naming this `filename`, _crazy with go-style names we dont have to be_

---

sidenote, this is another good example of the benefits to having a shared library holding these import/export app structure definitions

internal/local/structure.go
```diff
@@ -230,3 +232,118 @@ func parseServices(rootDir string) ([]ServiceStructure, error) {
...
+		err = WriteFile(filepath.Join(rootDir, NameGraphQL, NameCustomResolvers, nameFile), 0666, bytes.NewReader(data))
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func writeServices(rootDir string, services []ServiceStructure) error {
+	if err := os.MkdirAll(filepath.Join(rootDir, NameServices), os.ModePerm); err != nil {
```
makesitgo, line 307: `dir` var here

internal/local/structure.go
```diff
@@ -230,3 +232,118 @@ func parseServices(rootDir string) ([]ServiceStructure, error) {
...
+		nameService := svc.Config["name"].(string)
+		data, err := MarshalJSON(svc.Config)
+		if err != nil {
+			return err
+		}
+		err = WriteFile(filepath.Join(rootDir, NameServices, nameService, FileConfig.String()), 0666, bytes.NewReader(data))
+		if err != nil {
+			return err
+		}
+		for _, wh := range svc.IncomingWebhooks {
```
makesitgo, line 320: [nit] `webhook`

internal/local/structure.go
```diff
@@ -230,3 +232,118 @@ func parseServices(rootDir string) ([]ServiceStructure, error) {
...
+			dataSrc := wh[NameSource].(string)
+			nameWebhook := wh["name"].(string)
+			delete(wh, NameSource)
```
makesitgo, line 323: server code looks like we still write the `name` field, any reason this is deleted here?

internal/local/structure.go
```diff
@@ -230,3 +232,118 @@ func parseServices(rootDir string) ([]ServiceStructure, error) {
...
+		data, err := MarshalJSON(svc.Config)
+		if err != nil {
+			return err
+		}
+		err = WriteFile(filepath.Join(rootDir, NameServices, nameService, FileConfig.String()), 0666, bytes.NewReader(data))
+		if err != nil {
+			return err
+		}
+		for _, wh := range svc.IncomingWebhooks {
+			dataSrc := wh[NameSource].(string)
```
makesitgo, line 321: [nit] can just be `src` here

internal/local/structure.go
```diff
@@ -230,3 +232,118 @@ func parseServices(rootDir string) ([]ServiceStructure, error) {
...
+			nameWebhook := wh["name"].(string)
+			delete(wh, NameSource)
+			dataConfig, err := MarshalJSON(wh)
```
makesitgo, line 324: [nit] can just be `data` here

internal/local/structure.go
```diff
@@ -230,3 +232,118 @@ func parseServices(rootDir string) ([]ServiceStructure, error) {
...
+			if err != nil {
+				return err
+			}
+			err = WriteFile(filepath.Join(rootDir, NameServices, nameService, NameIncomingWebhooks, nameWebhook, FileConfig.String()), 0666, bytes.NewReader(dataConfig))
```
makesitgo, line 328: can we go multiline on these function invocations?

```go
dirWebhook := filepath.Join(dirSvc, NameIncomingWebhooks, nameWebhook)
if err := WriteFile(
  filepath.Join(dirWebhook, FileConfig.String()),readability
  0666,
  bytes.NewReader(dataConfig),
); err != nil {
  return err
}
```

internal/local/structure.go
```diff
@@ -230,3 +232,118 @@ func parseServices(rootDir string) ([]ServiceStructure, error) {
...
+	if err := os.MkdirAll(filepath.Join(rootDir, NameServices), os.ModePerm); err != nil {
+		return err
+	}
+	for _, svc := range services {
+		nameService := svc.Config["name"].(string)
+		data, err := MarshalJSON(svc.Config)
+		if err != nil {
+			return err
+		}
+		err = WriteFile(filepath.Join(rootDir, NameServices, nameService, FileConfig.String()), 0666, bytes.NewReader(data))
```
makesitgo, line 316: could also make `dirSvc` to re-use for webhooks and rules below

internal/local/structure.go
```diff
@@ -230,3 +232,118 @@ func parseServices(rootDir string) ([]ServiceStructure, error) {
...
+			if err != nil {
+				return err
+			}
+			err = WriteFile(filepath.Join(rootDir, NameServices, nameService, NameIncomingWebhooks, nameWebhook, FileSource.String()), 0666, bytes.NewReader([]byte(dataSrc)))
+			if err != nil {
+				return err
+			}
+		}
+		for _, rule := range svc.Rules {
+			dataRule, err := MarshalJSON(rule)
```
makesitgo, line 338: [nit] `data` here

internal/local/structure_v1.go
```diff
@@ -110,3 +113,115 @@ func (a *AppDataV1) LoadData(rootDir string) error {
...
+		}
+		err = WriteFile(filepath.Join(rootDir, NameAuthProviders, fmt.Sprintf("%s%s", ap["name"], extJSON)), 0666, bytes.NewReader(dataProvider))
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func writeTriggers(rootDir string, triggers []map[string]interface{}) error {
```
makesitgo, line 214: should this be in `structure.go` and used in `structure_v2` as well?

internal/local/structure_v1.go
```diff
@@ -110,3 +113,115 @@ func (a *AppDataV1) LoadData(rootDir string) error {
...
+	}
+	return nil
+}
+
+func writeFunctionsV1(rootDir string, functions []map[string]interface{}) error {
+	if err := os.MkdirAll(filepath.Join(rootDir, NameFunctions), os.ModePerm); err != nil {
+		return err
+	}
+	for _, function := range functions {
+		configFunction := function[NameConfig].(map[string]interface{})
```
makesitgo, line 178: if/when we _have_ to use these casts, use the `var, ok` syntax and return an error if `!ok` to avoid the potential of panics

internal/local/structure_v1.go
```diff
@@ -110,3 +113,115 @@ func (a *AppDataV1) LoadData(rootDir string) error {
...
+		nameFunction := configFunction["name"].(string)
```
makesitgo, line 179: avoid the panic here as well, and i'd say `name` is fine here within the scope of the for loop

internal/local/structure_v1.go
```diff
@@ -110,3 +113,115 @@ func (a *AppDataV1) LoadData(rootDir string) error {
...
+		dataFunction, err := MarshalJSON(configFunction)
```
makesitgo, line 180: i like ` config` here to match `src`, or otherwise `data` is just fine here too

internal/local/structure_v1.go
```diff
@@ -110,3 +113,115 @@ func (a *AppDataV1) LoadData(rootDir string) error {
...
+		if err != nil {
+			return err
+		}
+		err = WriteFile(filepath.Join(rootDir, NameFunctions, nameFunction, FileConfig.String()), 0666, bytes.NewReader(dataFunction))
+		if err != nil {
```
makesitgo, line 185: [nit] a lot of this can be cleaned/consolidated with common `dir` vars and inline the if statements

internal/local/structure_v1.go
```diff
@@ -110,3 +113,115 @@ func (a *AppDataV1) LoadData(rootDir string) error {
...
+		}
+	}
+	return nil
+}
+
+func writeAuthProviders(rootDir string, authProviders []map[string]interface{}) error {
+	if err := os.MkdirAll(filepath.Join(rootDir, NameAuthProviders), os.ModePerm); err != nil {
+		return err
+	}
+	for _, ap := range authProviders {
```
makesitgo, line 201: [nit] `authProvider`

internal/local/structure_v1.go
```diff
@@ -110,3 +113,115 @@ func (a *AppDataV1) LoadData(rootDir string) error {
...
+		dataProvider, err := MarshalJSON(ap)
```
makesitgo, line 202: `data` here

internal/local/structure_v1.go
```diff
@@ -110,3 +113,115 @@ func (a *AppDataV1) LoadData(rootDir string) error {
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func writeTriggers(rootDir string, triggers []map[string]interface{}) error {
+	for _, trigger := range triggers {
+		nameTrigger := trigger["name"].(string)
```
makesitgo, line 216: `name` is probably fine here within the scope of the for loop

internal/local/structure_v1.go
```diff
@@ -110,3 +113,115 @@ func (a *AppDataV1) LoadData(rootDir string) error {
...
+		dataTrigger, err := MarshalJSON(trigger)
```
makesitgo, line 217: `data` here

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+		return err
+	}
+	if err := writeValues(rootDir, a.Values); err != nil {
+		return err
+	}
+	graphQL := a.GraphQL
+	if a.GraphQL == nil {
+		graphQL = &GraphQLStructure{}
+	}
+	if err := writeGraphQL(rootDir, *graphQL); err != nil {
```
makesitgo, line 368: [question] does this _always_ want to be written (at least the directory)?  if so, then i was probably mistaken in making this field a pointer to `GraphQLStructure` (and we could remove that).

if not, we should do just do the rare double-nesting here:
```go
if a.GraphQL != nil {
  if err := writeGraphQL(rootDir, *a.GraphQL); err != nil { return err }
}
```

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+	if err := writeSync(rootDir, a.Sync); err != nil {
+		return err
+	}
+	if err := writeDataSources(rootDir, a.DataSources); err != nil {
+		return err
+	}
+	if err := writeHTTPEndpoints(rootDir, a.HTTPEndpoints); err != nil {
+		return err
+	}
+	return nil
```
makesitgo, line 389: triggers?

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+}
+
+func writeFunctionsV2(rootDir string, functions *FunctionsStructure) error {
+	configs := []map[string]interface{}{}
+	if functions != nil {
+		configs = functions.Configs
+	}
+	dataFuncs, err := MarshalJSON(configs)
```
makesitgo, line 397: `data` or `configs` here

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+	if err != nil {
+		return err
+	}
+	err = WriteFile(filepath.Join(rootDir, NameFunctions, FileConfig.String()), 0666, bytes.NewReader(dataFuncs))
+	if err != nil {
+		return err
+	}
+	if functions == nil {
+		return nil
```
makesitgo, line 406: id either add a comment here why we're returning early (`// if no functions, write the empty config and no more`), or use a similar pattern above:

```go
var sources map[string]string
if functions != nil {
  sources = functions.Sources // you could even do this all at the top of the function in one "if not nil" check
}
```

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+	}
+	return nil
+}
+
+func writeAuth(rootDir string, auth *AuthStructure) error {
+	if auth == nil {
+		return nil
+	}
+	if auth.Providers != nil {
+		dataProvider, err := MarshalJSON(auth.Providers)
```
makesitgo, line 422: `data` is fine here

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+		if err != nil {
+			return err
+		}
+		err = WriteFile(filepath.Join(rootDir, NameAuth, FileProviders.String()), 0666, bytes.NewReader(dataProvider))
+		if err != nil {
+			return err
+		}
+	}
+	if auth.CustomUserData != nil {
+		dataCustomUser, err := MarshalJSON(auth.CustomUserData)
```
makesitgo, line 432: `data` is fine here

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+func writeAuth(rootDir string, auth *AuthStructure) error {
+	if auth == nil {
+		return nil
+	}
+	if auth.Providers != nil {
+		dataProvider, err := MarshalJSON(auth.Providers)
+		if err != nil {
+			return err
+		}
+		err = WriteFile(filepath.Join(rootDir, NameAuth, FileProviders.String()), 0666, bytes.NewReader(dataProvider))
```
makesitgo, line 426: common `dir` var here could be re-used b/w the two if blocks

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+	}
+	return nil
+}
+
+func writeDataSources(rootDir string, dataSources []DataSourceStructure) error {
+	if err := os.MkdirAll(filepath.Join(rootDir, NameDataSources), os.ModePerm); err != nil {
+		return err
+	}
+	for _, ds := range dataSources {
+		nameDataSource := ds.Config["name"].(string)
```
makesitgo, line 464: `name` is probably fine here within this for loop scope

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+		dataDataSource, err := MarshalJSON(ds.Config)
```
makesitgo, line 465: `data` or `config` also keeps this short/simple

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+		if err != nil {
+			return err
+		}
+		for _, rule := range ds.Rules {
+			schema := rule[NameSchema]
+			dataSchema, err := MarshalJSON(schema)
+			if err != nil {
+				return err
+			}
+			delete(rule, NameSchema)
```
makesitgo, line 479: can we make sure this isn't mutating the actual `AppData`?  just want to make sure after we write data, the data source still has its schema field

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+			dataRule, err := MarshalJSON(rule)
+			if err != nil {
+				return err
+			}
+			err = WriteFile(filepath.Join(rootDir, NameDataSources, nameDataSource, fmt.Sprintf("%s", rule["database"]), fmt.Sprintf("%s", rule["collection"]), FileRules.String()), 0666, bytes.NewReader(dataRule))
```
makesitgo, line 484: should probably consolidate the logic for these filepaths here too, common `dir` vars and whatnot

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+			return err
+		}
+		err = WriteFile(filepath.Join(rootDir, NameHTTPEndpoints, nameHTTPEndpoint, FileConfig.String()), 0666, bytes.NewReader(data))
+		if err != nil {
+			return err
+		}
+		for _, wh := range httpEndpoint.IncomingWebhooks {
+			dataSrc := wh[NameSource].(string)
+			nameWebhook := wh["name"].(string)
+			delete(wh, NameSource)
```
makesitgo, line 514: likewise to make sure we're not mutating the `AppData` here

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+		data, err := MarshalJSON(httpEndpoint.Config)
+		if err != nil {
+			return err
+		}
+		err = WriteFile(filepath.Join(rootDir, NameHTTPEndpoints, nameHTTPEndpoint, FileConfig.String()), 0666, bytes.NewReader(data))
+		if err != nil {
+			return err
+		}
+		for _, wh := range httpEndpoint.IncomingWebhooks {
+			dataSrc := wh[NameSource].(string)
```
makesitgo, line 512: `src` is fine here

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+			nameWebhook := wh["name"].(string)
+			delete(wh, NameSource)
+			dataConfig, err := MarshalJSON(wh)
```
makesitgo, line 515: `config` is fine here too

internal/local/structure_v2.go
```diff
@@ -331,3 +327,204 @@ func parseSync(rootDir string) (*SyncStructure, error) {
...
+			if err != nil {
+				return err
+			}
+			err = WriteFile(filepath.Join(rootDir, NameHTTPEndpoints, nameHTTPEndpoint, nameWebhook, FileConfig.String()), 0666, bytes.NewReader(dataConfig))
+			if err != nil {
+				return err
+			}
+			err = WriteFile(filepath.Join(rootDir, NameHTTPEndpoints, nameHTTPEndpoint, nameWebhook, FileSource.String()), 0666, bytes.NewReader([]byte(dataSrc)))
```
makesitgo, line 523: DRY up the filepath logic here

makesitgo (commented):
going to look through tests now, but here's likely the last of my cleanup comments

internal/local/structure.go
nathantfrank: This is deleting the source code from the map, so that it doesn't get thrown into the config file.

makesitgo: gotcha, i read that wrong at first and thought we were deleting the `"name"` field 🙄 

as with the other cases of this, lets just make sure the `AppData` stays in tact after this `delete`

nathantfrank: Yeah, I just did a test and it doesn't... I thought the value loop in go created a separate copy. So I'm creating temp copies in my next commit.

internal/local/structure_v1.go
nathantfrank: Triggers is v1 only. It got combined into function in functions in v2.

internal/local/structure_v2.go
nathantfrank: yus

internal/local/structure_v2.go
nathantfrank: Yes, making temp versions in next commit.

internal/local/structure_v1.go
nathantfrank: I was incorrect. Fixing in next commit.

internal/local/structure_v2.go
nathantfrank: I believe so.

nathantfrank: But rather not change this before our beta release.

makesitgo: don't disagree, so let's roll with the wrapping in `!= nil` then

internal/local/structure_test.go
```diff
@@ -0,0 +1,374 @@
...
+				"values": map[string]interface{}{
+					"greeting": "hello",
+				},
+			},
+		}
+
+		err := writeEnvironments(tmpDir, data)
+		assert.Nil(t, err)
+
+		environment, err := ioutil.ReadFile(filepath.Join(tmpDir, NameEnvironments, "development"+extJSON))
```
makesitgo, line 87: [nit/optional] you could table (for loop) these test asserts

internal/local/structure_v2.go
nathantfrank: Yes, making temp versions in next commit.

internal/local/structure_v2_test.go
```diff
@@ -66,3 +68,259 @@ func TestParseDataSources(t *testing.T) {
...
+	t.Run("should write functions to disk", func(t *testing.T) {
+		data := &FunctionsStructure{
+			Configs: []map[string]interface{}{
+				{
+					"name":    "test",
+					"private": true,
+				},
+			},
+			Sources: map[string]string{
+				"test.json": "exports = function(){\n  console.log('Hello World!');\n};",
```
makesitgo, line 86: [nit] the extension here should be `.js`

and it might be worth adding an entry here where:

`"nested/test.js": "exports = function(){\n  console.log('Hello World!');\n}"`

and ensure we write that to the proper directory

nathantfrank: good catch!

internal/local/structure.go
```diff
@@ -230,3 +233,180 @@ func parseServices(rootDir string) ([]ServiceStructure, error) {
...
+	return WriteFile(filepath.Join(rootDir, FileSecrets.String()), 0666, bytes.NewReader(data))
+}
+
+func writeEnvironments(rootDir string, environments map[string]map[string]interface{}) error {
+	for env, values := range environments {
+		data, err := MarshalJSON(values)
+		if err != nil {
+			return err
+		}
+		if err = WriteFile(
```
makesitgo, line 254: [nothing to change here] but with this inline ifs, you can use `:=` and it'll keep it scoped properly (https://play.golang.org/p/WP8C0eMoyL7)

makesitgo (approved):
🔥 ... that's all i can say

🆒  beans

PR #96, opened 2021-03-09
REALMC-7429: Finalize CLI help text

makesitgo:
looking for all kinds of thoughts/suggestions here.  a bulk of these changes are centered around:
* sorting the commands/flags so they appeared logically and consistently
* trying to keep language consistent across the various command texts
* removing most shorthands for flags, some of them became pretty unintuitive (e.g. `s` for `include-hosting`??) and i think the better approach here long-term would be figuring out proper auto-complete
* various command text cleanups with regards to spacing and such

### DIFF
--- a/internal/cli/inputs.go
+++ b/internal/cli/inputs.go
@@ -8,6 +8,7 @@ import (
 	"github.com/10gen/realm-cli/internal/cloud/realm"
 	"github.com/10gen/realm-cli/internal/local"
 	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/10gen/realm-cli/internal/utils/flags"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/spf13/pflag"
@@ -15,7 +16,8 @@ import (
 
 const (
 	flagApp      = "app"
-	flagAppUsage = "the Realm app name or id to manage"
+	flagAppShort = "a"
+	flagAppUsage = "the remote Realm app name or id"
 
 	flagProject      = "project"
 	flagProjectUsage = "the MongoDB cloud project id"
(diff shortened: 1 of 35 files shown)


internal/commands/app/delete_inputs.go
```diff
@@ -11,8 +11,9 @@ import (
 )
 
 var (
-	flagApps      = "apps"
-	flagAppsUsage = "the Realm app names or ids to manage"
+	flagApp      = "app"
```
kpatel71716, line 14: 👍  I like this change + having a short

internal/cli/inputs.go
```diff
@@ -32,8 +34,10 @@ func (i ProjectInputs) Filter() realm.AppFilter { return realm.AppFilter{i.Proje
 
 // Flags registers the project app input flags to the provided flag set
 func (i *ProjectInputs) Flags(fs *pflag.FlagSet) {
+	fs.StringVarP(&i.App, flagApp, flagAppShort, "", flagAppUsage)
+
 	fs.StringVar(&i.Project, flagProject, "", flagProjectUsage)
-	fs.StringVar(&i.App, flagApp, "", flagAppUsage)
+	flags.MarkHidden(fs, flagProject)
```
kpatel71716, line 40: [q] Is there a reason the project flag is to be hidden? I imagine there would be a use case where someone would want to see apps filtered by project

kpatel71716 (commented):
Looks really good! Just have one question from this pass

internal/cli/inputs.go
makesitgo: historically that project flag was really only useful for mongo employees (who have access to _a lot_ of apps).  in a similar vein to the base urls for realm/atlas, I made them hidden.

nathantfrank (approved):
LGTM, pretty slick stuff :metal:

PR #100, opened 2021-03-11
REALMC-7181: Add release plan

makesitgo:
Plan on first running this with `./contrib/release.js 2.0.0-beta.0`, and then using `npm publish --tag beta`

There's a fair amount of copy paste here, i'll call out in downstream comments any questions I had


package.json
```diff
@@ -0,0 +1,41 @@
...
+  "keywords": [
+    "mongodb",
+    "realm"
+  ],
+  "author": "",
+  "dependencies": {
+    "request": "^2.88.0"
+  },
+  "devDependencies": {
+    "babel-eslint": "^8.2.2",
```
makesitgo, line 26: ⚠️  if we're not installing/building the transpiler anymore, are these dependencies needed?

testInstall.js
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env node
```
makesitgo, line 1: ⚠️  couldn't really determine if this is used or not (is it just a manual test for our own purposes?)... kept it, but not sure if its needed

package.json
UnicodeSnowman: there are dependencies required for the install script, so we do need _some_ dependencies, but likely not all of these. probably worth going through and cleaning this up as a separate ticket.

testInstall.js
UnicodeSnowman: yeah, I believe so cc @chickenbug. maybe you guys can sync up to utilize this script pre-release to ensure everything is working (I just don't recall exactly how it works)

version.json
```diff
@@ -0,0 +1,4 @@
+{
+  "version": "1.3.1",
```
UnicodeSnowman, line 2: this will get resolved when we merge `next => master`, but this is `1.3.2` now.

.evg.yml
```diff
@@ -261,9 +361,9 @@ buildvariants:
     realm_server_url: "http://localhost:8080"
     yarn_url: "https://s3.amazonaws.com/stitch-artifacts/yarn/latest.tar.gz"
     node_version: 12.16.2
-    transpiler_target: node8-linux
-    version: 0.0.0
+    version: 2.0.0-beta.0
```
UnicodeSnowman, line 364: can we leave this at the current version (`1.3.2`) and bump this to `2.0.0-beta.0` as part of a separate ticket for actually releasing? would be nice to keep release commits/changes separate from actual development updates.

UnicodeSnowman (approved):
👌 just a few small things

package.json
makesitgo: 👍 ... filed https://jira.mongodb.org/browse/REALMC-8405

testInstall.js
chickenbug: Whoa what a throwback. Yeah I _think_ this tests the install.js script by running npm in a temp subfolder. I don't fully remember though

makesitgo: ok ran this locally, looks like we are `npm install`ing to a temp directory our package (`mongodb-realm-cli`) and then ensuring the `realm-cli` and `transpiler` executables are present.

going to leave this in since it works as a manual/local check.  if/when we can get this release process more automated (ideally through evergreen, we won't need an team member's `npm` access ((nor good internet)) to actually push new releases), we can use this script as a post-check to make sure everything ran smoothly

alternatively, we could also include this as a check on "non-patchable" commits just to make sure we're still publicly available.  will file these ideas as a follow up ticket to pursue in the v2 release of this

makesitgo: https://jira.mongodb.org/browse/REALMC-8411

PR #106, opened 2021-03-18
REALMC-8430: chmod +x contrib/release.js

UnicodeSnowman:


tsedgwick:
is this supposed to be just an empty file?

UnicodeSnowman:
The file is there, just made it executable. The file itself doesn't show up because the contents didn't change

tsedgwick (approved):
whoops. I read this as adding the command `chmod +x` inside of `contrib/release.js`. makes a lot more sense now 🤦🏻

PR #110, opened 2021-03-18
REALMC-8451: Ensure all segment events are flushed

UnicodeSnowman:
the Segment library batches enqueued events until either the batch limit is hit, or a certain amount of time has elapsed (see https://segment.com/docs/connections/sources/catalog/libraries/server/go/), which was preventing events from being sent (since the program would exit before this happened).

you can create a new analytics client with a config and set the batch size to 1 according to the docs I linked agove, but I tried this and the program still exits before dispatching the events to segment. this seems to be the best way to force a flush.

I also think we need to revisit some of this soon for the case where users are blocking calls to segment and/or paranoid about tracking. I tried this with Little Snitch and it's really noisy (the Segment library uses it's own logger by default, which prints to stderr in the event of this sort of failure... it supports passing a custom logger, which we could use to minimize this and/or send errors directly to our servers instead).

### DIFF
--- a/internal/cli/command_factory.go
+++ b/internal/cli/command_factory.go
@@ -144,6 +144,12 @@ func (factory *CommandFactory) Build(command CommandDefinition) *cobra.Command {
 
 // Close closes the command factory
 func (factory *CommandFactory) Close() {
+	if factory.telemetryService != nil {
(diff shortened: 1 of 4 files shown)


tsedgwick (approved):
lgtm

PR #114, opened 2021-03-24
REALMC-5765: Add GCP Support

Gabri3l:


etc/transpiler/index.js
```diff
@@ -25,7 +25,15 @@ function processData(input) {
...
         continue;
       }
-      output.results.push({ code: result.code, map: result.map });
+      output.results.push({
+        // This has been introduced to replace the use of `eval` in @protobufjs/inquire
+        // which is used by any service module from the GCP family.
+        // TODO(REALMC-5102): Remove this temporary workaround to have the inquire
+        // module work (no more .replace should be needed with `eval` support). Alternatively
+        // if we decide to stay away from `eval` then remove this TODO.
+        code: result.code.replace(`eval("quire".replace(/^/,"re"))`, `require`),
```
arahmanan, line 34: I noticed that there isn't a space between the two `replace` arguments here. Should that string be the same as the one in `baas` and `baas-ui`?

arahmanan (approved):
LGTM! Just a question

etc/transpiler/index.js
Gabri3l: Not sure how that happened, but yeah there shouldn't be any space! Awesome catch!!

PR #115, opened 2021-03-26
REALMC-8432: Set correct file permissions for static files

UnicodeSnowman:
* 'rw' should be sufficient here, don't need executable

### DIFF
--- a/internal/local/hosting.go
+++ b/internal/local/hosting.go
@@ -23,12 +23,12 @@ const (
 
 var (
 	validAttrNames = map[string]struct{}{
-		api.HeaderContentType:             struct{}{},
-		api.HeaderContentDisposition:      struct{}{},
-		api.HeaderContentLanguage:         struct{}{},
-		api.HeaderContentEncoding:         struct{}{},
-		api.HeaderCacheControl:            struct{}{},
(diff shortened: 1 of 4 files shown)


internal/local/hosting.go
```diff
@@ -23,12 +23,12 @@ const (
...
-		api.HeaderContentLanguage:         struct{}{},
-		api.HeaderContentEncoding:         struct{}{},
-		api.HeaderCacheControl:            struct{}{},
-		api.HeaderWebsiteRedirectLocation: struct{}{},
+		api.HeaderContentType:             {},
+		api.HeaderContentDisposition:      {},
+		api.HeaderContentLanguage:         {},
+		api.HeaderContentEncoding:         {},
+		api.HeaderCacheControl:            {},
+		api.HeaderWebsiteRedirectLocation: {},
```
UnicodeSnowman, line 31: my editor did this 🤷

UnicodeSnowman:
FYI gonna fix this in `1.3.x` too

internal/local/hosting.go
makesitgo: i learned something today... neat!

makesitgo (approved):
🆒  beans

PR #116, opened 2021-03-29
REALMC-7243 Add segment tracking to CLI 

jo-tse-mongo:


jo-tse-mongo:
Need to add tests, but this is a preliminary PR to get some feedback

makesitgo:
going to ping @UnicodeSnowman here for thoughts as well, but I think this is on the right track for what we need out of this.  just a couple of random thoughts:

* if we constructed the telemetry service inside `main.go` (before `c.Run()` is called), you'd be able to achieve two things:
  * a consistent place to track "START"
  * a consistent to place to track "COMPLETE" (if we do a `exitStatus == 0` check)
* that leaves us just to be calling `Track` on errors encountered in the command implementations ... I have a couple of ideas how we could achieve this:
  * could just pass the telemetry service in, from `main.go`, into the `New*CommandFactory` funcs (along with `ui`).  this enables us to track errors on a case-by-case basis, like you currently seem to have
  * could create a wrapper around `*cli.BaseUi` and override the default `Error(string)` behavior to both log the error to the user and send details to Segment

if we truly do want to capture any error in Segment that we've logged back to the user (I don't see why not, at least at first), then the second option there may make the most sense.

thoughts?

utils/telemetry/tracker.go
```diff
@@ -0,0 +1,58 @@
+package telemetry
+
+import (
+	"fmt"
+	"github.com/mitchellh/cli"
+	"gopkg.in/segmentio/analytics-go.v3"
+)
+
+var (
+	segmentWriteKey string //value will be injected at build-time
```
tsedgwick, line 10: how are we injecting this?

utils/telemetry/tracker.go
```diff
@@ -0,0 +1,58 @@
...
+		properties[datum.Key] = datum.Value
+	}
+	if err := s.client.Enqueue(analytics.Track{
+		MessageId:  event.id,
+		Timestamp:  event.time,
+		Event:      string(event.eventType),
+		UserId:     event.userID,
+		Properties: properties,
+	}); err != nil {
+		s.ui.Info(fmt.Sprintf("failed to send Segment event %q: %s", event.eventType, err))
```
tsedgwick, line 52: we should change this from `Segment` to `tracking`

commands/command.go
```diff
@@ -209,6 +212,8 @@ func (c *BaseCommand) run(args []string) error {
 		c.storage = storage.New(fileStrategy)
 	}
 
+	c.service = telemetry.NewService(c.flagTelemetryMode, c.user.APIKey, c.Name, c.UI)
```
tsedgwick, line 215: if this sets up the service, won't all of the START tracks fail?

tsedgwick: I think an alternative approach would be to create the service here as well as issue the START track call here

Something like: 
```
c.service = telemetry.NewService(c.flagTelemetryMode, c.user.APIKey, c.Name, c.UI)
dc.service.TrackEvent(telemetry.EventTypeCommandStart)

**Review by tsedgwick — CHANGES_REQUESTED** [review]

**Comment by tsedgwick** [comment]
> going to ping @UnicodeSnowman here for thoughts as well, but I think this is on the right track for what we need out of this. just a couple of random thoughts:
> 
> * if we constructed the telemetry service inside `main.go` (before `c.Run()` is called), you'd be able to achieve two things:
>   
>   * a consistent place to track "START"
>   * a consistent to place to track "COMPLETE" (if we do a `exitStatus == 0` check)
> * that leaves us just to be calling `Track` on errors encountered in the command implementations ... I have a couple of ideas how we could achieve this:
>   
>   * could just pass the telemetry service in, from `main.go`, into the `New*CommandFactory` funcs (along with `ui`).  this enables us to track errors on a case-by-case basis, like you currently seem to have
>   * could create a wrapper around `*cli.BaseUi` and override the default `Error(string)` behavior to both log the error to the user and send details to Segment
> 
> if we truly do want to capture any error in Segment that we've logged back to the user (I don't see why not, at least at first), then the second option there may make the most sense.
> 
> thoughts?

I'm actually onboard with this approach of keeping everything in main.go if we can help it.

File: utils/telemetry/tracker.go (hunk 1 above)
Comment (makesitgo) [reply utils/telemetry/tracker.go:10]: in the `go build` command, using:

`-ldflags="-X 'github.com/10gen/realm-cli/utils/telemetry.segmentWriteKey=$SEGMENT_WRITE_KEY'"`

File: utils/telemetry/tracker.go (hunk 2 above)
Comment (makesitgo) [reply utils/telemetry/tracker.go:52]: this may even just want to fail silently?  Chris was talking to me about this while I was OOO in regards to the new CLI, but some of these errors/warnings can potentially be noisy and unnecessary to the end user.  they cant do anything about it (if its failing)

File: utils/telemetry/events.go [hunk 4]
```diff
@@ -0,0 +1,67 @@
...
+
+	EventDataKeyError = "err"
+)
+
+type Mode string
+
+const (
+	ModeOff   Mode = "off"
+	ModeOn    Mode = "on"
+	ModeEmpty Mode = ""
```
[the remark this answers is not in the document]
makesitgo: i imagine this was done to more closely mirror the new CLI, where empty meant "no choice" which meant "use what we know you previously set it to based on your profile" (it's a sticky flag)

for these purposes and this 1.x branch, we dont easily support sticky flags here so we don't need that complexity.  using a boolean flag will work fine here

utils/telemetry/service.go
```diff
@@ -0,0 +1,56 @@
...
+	userID      string
+	executionID string
+	command     string
+	tracker     Tracker
+}
+
+func NewService(mode Mode, userID string, command string, ui cli.Ui) *Service {
+	if !isValid(mode){
+		return nil
+	}
```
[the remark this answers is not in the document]
makesitgo: in fact, this is actually dead code, b/c the `Set` implementation on `Mode` ensures all flag values are valid (which occurs during `flag.Parse`).  we won't ever get to `NewService` calls unless a proper value was passed in (moot point here though, considering we can just switch to to `bool` flag anyways)

utils/telemetry/tracker.go
tsedgwick: should that be in the .evg.yml file then?

utils/telemetry/tracker.go
tsedgwick: I'm fine with failing silently too

utils/telemetry/tracker.go
makesitgo: indeed (i'll be making sure to do a more thorough review now that weve landed on an approach and get some tests eventually)... @jo-tse-mongo, if you want to get ahead of it here, you can check out the `go build...` command in `evg.yml` on `master` branch ... that should show you how we're trying to include these write keys (you also should be able to use that same env var)

utils/telemetry/tracker.go
UnicodeSnowman: yeah, IMO this is one of the few cases where failing silently is the right idea. we really don't want to be littering CLI output with these failure messages in the event of a non critical failure, especially one that essentially says "Hey! Normally we'd be tracking you, but something is wrong! You can't do anything about it!"

we would know that something is up on our end when we notice we aren't seeing any `1.3.x` events coming in.

commands/command.go
```diff
@@ -36,8 +38,9 @@ type BaseCommand struct {
 
 	Name string
 
-	CLI *cli.CLI
-	UI  cli.Ui
+	CLI     *cli.CLI
+	UI      cli.Ui
+	Service *telemetry.Service
```
makesitgo, line 43: [naming nit] rename `Service` to be a bit more specific: `TelemetryService`

commands/command.go
```diff
@@ -50,6 +53,7 @@ type BaseCommand struct {
 	flagBaseURL       string
 	flagAtlasBaseURL  string
 	flagYes           bool
+	flagTelemetryOn   bool
```
makesitgo, line 56: I prefer to avoid "default true" boolean flags, perhaps we switch this to `flagNoTelemetry` or `flagTelemetryOff` or something along those lines? `--no-telemetry` feels easiest to write out, but your call

---

also, let's move this field to live on the `telemetry.Service` struct definition.  you can still `set.BoolVar` the same with just with `&c.TelemetryService.NoTelemetry`, and avoid including it in the `SetFields` function signature

commands/command.go
```diff
@@ -209,6 +213,14 @@ func (c *BaseCommand) run(args []string) error {
 		c.storage = storage.New(fileStrategy)
 	}
 
+	user, err := c.User()
+	if err != nil {
+		return err
+	}
+
+	c.Service.SetFields(c.flagTelemetryOn, user.PublicAPIKey, c.Name)
```
makesitgo, line 221: rather than a catch-all `SetFields` method here, lets break it out into proper setters:
```go
c.TelemetryService.SetCommand(c.Name)
c.TelemetryService.SetUserID(user.PublicAPIKey)
```

also, make sure to call `SetCommand` prior to attempting to read `c.User()`, just so we get as much info as possible (I think moving `SetCommand` to be the very first line of code in this `run` func might be best)

commands/diff.go
```diff
@@ -4,13 +4,15 @@ import (
 	"io"
 	"os"
 
+	"github.com/10gen/realm-cli/utils/telemetry"
+
```
makesitgo, line 8: this shouldn't be in its own import group, and as a driveby fix could you move `mitchellh/cli` into its own 3rd group?

commands/diff.go
```diff
@@ -4,13 +4,15 @@ import (
...
 	"github.com/10gen/realm-cli/models"
 	"github.com/10gen/realm-cli/utils"
 	"github.com/mitchellh/cli"
 )
 
 // NewDiffCommandFactory returns a new cli.CommandFactory given a cli.Ui
-func NewDiffCommandFactory(ui cli.Ui) cli.CommandFactory {
+func NewDiffCommandFactory(ui cli.Ui, service *telemetry.Service) cli.CommandFactory {
```
makesitgo, line 15: when outside the `telemetry` package, lets make sure to refer to this as `telemetryService` as opposed to just a generic `service`

commands/diff_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"io"
 	"testing"
 
+	"github.com/10gen/realm-cli/utils/telemetry"
+
```
makesitgo, line 8: same thing here, can you fixup the import groups (not sure why/how we missed this in the first go-round)

commands/diff_test.go
```diff
@@ -13,7 +15,8 @@ import (
 
 func setUpBasicDiffCommand() (*DiffCommand, *cli.MockUi) {
 	mockUI := cli.NewMockUi()
-	cmd, err := NewDiffCommandFactory(mockUI)()
+	mockService := &telemetry.Service{}
```
makesitgo, line 18: [nit] `telemetryService ` (you could call it "mock" I suppose, but there's really nothing "mockish" about it (unlike the UI, which actually has an alternate mock implementation)

commands/export.go
```diff
@@ -7,6 +7,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/10gen/realm-cli/utils/telemetry"
+
```
makesitgo, line 11: no newline here

commands/export.go
```diff
@@ -19,7 +21,7 @@ import (
 const numWorkers = 4
 
 // NewExportCommandFactory returns a new cli.CommandFactory given a cli.Ui
-func NewExportCommandFactory(ui cli.Ui) cli.CommandFactory {
+func NewExportCommandFactory(ui cli.Ui, service *telemetry.Service) cli.CommandFactory {
```
makesitgo, line 24: same thing about `telemetryService` here

commands/export_test.go
```diff
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/10gen/realm-cli/utils/telemetry"
+
```
makesitgo, line 13: no newline here (and the existing import group could use a little driveby fixin too)

commands/export_test.go
```diff
@@ -24,7 +26,8 @@ import (
 func TestExportCommand(t *testing.T) {
 	setup := func() (*ExportCommand, *cli.MockUi) {
 		mockUI := cli.NewMockUi()
-		cmd, err := NewExportCommandFactory(mockUI)()
+		mockService := &telemetry.Service{}
```
makesitgo, line 29: same thing about `telemetryService` here (same thing about "mock" here too)

commands/import.go
```diff
@@ -50,7 +52,7 @@ func errIncludeHosting(err error) error {
 }
 
 // NewImportCommandFactory returns a new cli.CommandFactory given a cli.Ui
-func NewImportCommandFactory(ui cli.Ui) cli.CommandFactory {
+func NewImportCommandFactory(ui cli.Ui, service *telemetry.Service) cli.CommandFactory {
```
makesitgo, line 55: same thing about `telemetryService` here

commands/import_test.go
```diff
@@ -12,6 +12,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/10gen/realm-cli/utils/telemetry"
+
```
makesitgo, line 16: existing import group could also use some love here... just our standard pattern:

```go
import (
  // std lib imports

  // realm-cli imports

  // all other imports
)
```

commands/import_test.go
```diff
@@ -29,7 +31,8 @@ import (
 
 func setUpBasicCommand() (*ImportCommand, *cli.MockUi) {
 	mockUI := cli.NewMockUi()
-	cmd, err := NewImportCommandFactory(mockUI)()
+	mockService := &telemetry.Service{}
```
makesitgo, line 34: same thing about `telemetryService` here (and mock here)

commands/login.go
```diff
@@ -16,12 +17,13 @@ const (
 )
 
 // NewLoginCommandFactory returns a new cli.CommandFactory given a cli.Ui
-func NewLoginCommandFactory(ui cli.Ui) cli.CommandFactory {
+func NewLoginCommandFactory(ui cli.Ui, service *telemetry.Service) cli.CommandFactory {
```
makesitgo, line 20: same thing about `telemetryService` here

commands/login_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/10gen/realm-cli/utils/telemetry"
+
```
makesitgo, line 9: starting to wish i didn't choose to die on this hill, but a re-org here to fit our standards would be much appreciated

commands/login_test.go
```diff
@@ -17,7 +19,8 @@ func TestLoginCommand(t *testing.T) {
 	t.Run("required arguments", func(t *testing.T) {
 		setup := func() (*LoginCommand, *cli.MockUi) {
 			mockUI := cli.NewMockUi()
-			cmd, err := NewLoginCommandFactory(mockUI)()
+			mockService := &telemetry.Service{}
```
makesitgo, line 22: same thing about `telemetryService` here (and mock too)

commands/login_test.go
```diff
@@ -56,7 +59,8 @@ func TestLoginCommand(t *testing.T) {
 	t.Run("when the user is not logged in", func(t *testing.T) {
 		setup := func() (*LoginCommand, *cli.MockUi) {
 			mockUI := cli.NewMockUi()
-			cmd, err := NewLoginCommandFactory(mockUI)()
+			mockService := &telemetry.Service{}
```
makesitgo, line 62: same thing about `telemetryService` here (and mock too)

commands/login_test.go
```diff
@@ -107,7 +111,8 @@ func TestLoginCommand(t *testing.T) {
 	t.Run("when the user is logged in", func(t *testing.T) {
 		setup := func() (*LoginCommand, *cli.MockUi, *u.MockClient) {
 			mockUI := cli.NewMockUi()
-			cmd, err := NewLoginCommandFactory(mockUI)()
+			mockService := &telemetry.Service{}
```
makesitgo, line 114: same thing about `telemetryService` here (and mock too)

commands/logout.go
```diff
@@ -1,16 +1,18 @@
 package commands
 
 import (
+	"github.com/10gen/realm-cli/utils/telemetry"
```
makesitgo, line 4: add a newline below here

commands/logout.go
```diff
@@ -1,16 +1,18 @@
...
 	"github.com/mitchellh/cli"
 )
 
 // NewLogoutCommandFactory returns a new cli.CommandFactory given a cli.Ui
-func NewLogoutCommandFactory(ui cli.Ui) cli.CommandFactory {
+func NewLogoutCommandFactory(ui cli.Ui, service *telemetry.Service) cli.CommandFactory {
```
makesitgo, line 9: same thing about `telemetryService` here

commands/logout_test.go
```diff
@@ -3,6 +3,8 @@ package commands
 import (
 	"testing"
 
+	"github.com/10gen/realm-cli/utils/telemetry"
+
```
makesitgo, line 7: could use some import group 💚  here

commands/logout_test.go
```diff
@@ -14,7 +16,8 @@ import (
 func TestLogoutCommand(t *testing.T) {
 	setup := func(storage *storage.Storage) (*LogoutCommand, *cli.MockUi) {
 		mockUI := cli.NewMockUi()
-		cmd, err := NewLogoutCommandFactory(mockUI)()
+		mockService := &telemetry.Service{}
```
makesitgo, line 19: same thing about `telemetryService` here (and mock too)

commands/project_command.go
```diff
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"github.com/10gen/realm-cli/utils/telemetry"
```
makesitgo, line 4: newline below here

commands/project_command.go
```diff
@@ -9,11 +10,12 @@ const (
 )
 
 // NewProjectCommand returns a new *ProjectCommand
-func NewProjectCommand(name string, ui cli.Ui) *ProjectCommand {
+func NewProjectCommand(name string, ui cli.Ui, service *telemetry.Service) *ProjectCommand {
```
makesitgo, line 13: same thing about `telemetryService` here

commands/secrets.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/10gen/realm-cli/utils/telemetry"
+
```
makesitgo, line 8: import group 💚

commands/secrets.go
```diff
@@ -12,12 +14,13 @@ import (
 )
 
 // NewSecretsCommandFactory returns a new cli.CommandFactory given a cli.Ui
-func NewSecretsCommandFactory(ui cli.Ui) cli.CommandFactory {
+func NewSecretsCommandFactory(ui cli.Ui, service *telemetry.Service) cli.CommandFactory {
```
makesitgo, line 17: same thing about `telemetryService` here

commands/secrets_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/10gen/realm-cli/utils/telemetry"
+
```
makesitgo, line 8: no newline here

commands/secrets_test.go
```diff
@@ -50,6 +52,7 @@ func setUpBasicSecretsCommand(
 }
 
 func TestSecretsCommand(t *testing.T) {
+	mockService := &telemetry.Service{}
```
makesitgo, line 55: same thing about `telemetryService` here (and mock too)

commands/whoami.go
```diff
@@ -3,16 +3,18 @@ package commands
 import (
 	"fmt"
 
+	"github.com/10gen/realm-cli/utils/telemetry"
```
makesitgo, line 6: newline below here

commands/whoami.go
```diff
@@ -3,16 +3,18 @@ package commands
...
 	"github.com/mitchellh/cli"
 )
 
 // NewWhoamiCommandFactory returns a new cli.CommandFactory given a cli.Ui
-func NewWhoamiCommandFactory(ui cli.Ui) cli.CommandFactory {
+func NewWhoamiCommandFactory(ui cli.Ui, service *telemetry.Service) cli.CommandFactory {
```
makesitgo, line 11: same thing about `telemetryService` here

commands/whoami_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"net/http"
 	"testing"
 
+	"github.com/10gen/realm-cli/utils/telemetry"
+
```
makesitgo, line 8: no newline here

commands/whoami_test.go
```diff
@@ -15,7 +17,8 @@ import (
 func TestWhoamiCommand(t *testing.T) {
 	setup := func(inMemoryUser *user.User, storage *storage.Storage) (*WhoamiCommand, *cli.MockUi) {
 		mockUI := cli.NewMockUi()
-		cmd, err := NewWhoamiCommandFactory(mockUI)()
+		mockService := &telemetry.Service{}
```
makesitgo, line 20: same thing about `telemetryService` here (and mock too)

main.go
```diff
@@ -5,6 +5,8 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/10gen/realm-cli/utils/telemetry"
+
```
makesitgo, line 9: no newline here

main.go
```diff
@@ -15,29 +17,39 @@ func main() {
 	c := cli.NewCLI(filepath.Base(os.Args[0]), utils.CLIVersion)
 	c.Args = os.Args[1:]
 
+	service := &telemetry.Service{}
```
makesitgo, line 20: same thing about `telemetryService` here

utils/telemetry/events.go
```diff
@@ -0,0 +1,37 @@
+package telemetry
+
+import (
+	"time"
+)
+
+// REALMC-7243 same as the new cli Segment tracking
```
makesitgo, line 7: we can just remove these comments (sometimes its nice to mention when a file was copy/pasted like this), but it's fine here without that

utils/telemetry/events.go
```diff
@@ -0,0 +1,37 @@
...
+
+const (
+	EventTypeCommandStart EventType = "COMMAND_START"
+	EventTypeCommandEnd   EventType = "COMMAND_END"
+	EventTypeCommandError EventType = "COMMAND_ERROR"
+)
+
+const (
+	eventDataKeyCommand     = "command"
+	eventDataKeyExecutionID = "xid"
```
makesitgo, line 34: can we include a `version`/`v` field (check how we're doing it on `master` now, it was a recent change/update)

utils/telemetry/service.go
```diff
@@ -0,0 +1,46 @@
+package telemetry
+
+import (
+	"time"
+
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+// REALMC-7243 same as the new cli Segment tracking
```
makesitgo, line 9: same thing here the comment can go away

utils/telemetry/service.go
```diff
@@ -0,0 +1,46 @@
...
+
+type Service struct {
+	userID      string
+	executionID string
+	command     string
+	tracker     Tracker
+}
+
+func (s *Service) SetFields(telemetryOn bool, userID string, command string) {
```
makesitgo, line 18: so I know I previously mentioned having `SetCommand` and `SetUserID` methods, but I can see `SetFields` was doing even more than just that now.

I still think `SetUserID` should be its own call, so we can do that separately, down the line in the `run` implementation (and still have the `tracker` and `command` set).  Instead of `SetCommand`, let's have a `Setup(command string)` function, and it can also do the `tracker` logic you have here (based on the `s.NoTelemetry` field)

utils/telemetry/service.go
```diff
@@ -0,0 +1,46 @@
...
+	if telemetryOn {
+		s.tracker = newSegmentTracker()
+	} else {
+		s.tracker = &noopTracker{}
+	}
+}
+
+func (s *Service) TrackEvent(eventType EventType, data ...EventData) {
+	s.tracker.Track(
+		event{
```
makesitgo, line 32: [meganit] unnecessary linebreaks here, could reduce a little indentation with just `s.tracker.Track(event{` as the first line (and `})` at the end)

utils/telemetry/service.go
```diff
@@ -0,0 +1,46 @@
...
+			userID:      s.userID,
+			time:        time.Now(),
+			executionID: s.executionID,
+			command:     s.command,
+			data:        data,
+		},
+	)
+}
+
+func (s *Service) Close() error {
```
makesitgo, line 44: `Close() implementation also changed on `master` since you probably last looked here... lets update this to match that now

utils/telemetry/tracker.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"gopkg.in/segmentio/analytics-go.v3"
+)
+
+var (
+	segmentWriteKey string //value will be injected at build-time
+)
+
+// REALMC-7243 same as the new cli Segment tracking
```
makesitgo, line 11: can get rid of these comments (but also as a sidenote: i'm noticing a lot of missing docstrings on otherwise exported package contents).  can we look into why `lint` isn't failing?  i see a "Command failed: no files found to be parsed" in the evg output, not sure if that just always shows up or not, but let's investigate (and add docstrings)

utils/telemetry/tracker.go
```diff
@@ -0,0 +1,57 @@
...
+type Tracker interface {
+	Track(event event)
+	Close() error
```
makesitgo, line 14: just a reminder to update this `interface` to match what's current on `master`

utils/telemetry/tracker.go
```diff
@@ -0,0 +1,57 @@
...
+func newSegmentTracker() Tracker {
+	if len(segmentWriteKey) == 0 {
+		return &noopTracker{}
+	}
+	client := analytics.New(segmentWriteKey)
+	return &segmentTracker{client}
+}
+
+func (s *segmentTracker) Track(event event) {
+	properties := make(map[string]interface{}, len(event.data)+2)
```
makesitgo, line 35: update this to `+3` and we'll want to include `eventDataKeyVersion` below

makesitgo (commented):
buncha little nitty things from me... should be ready to go thereafter, but i'll let you work through these before approving.  nice job with all this!

commands/command.go
```diff
@@ -209,6 +211,15 @@ func (c *BaseCommand) run(args []string) error {
 		c.storage = storage.New(fileStrategy)
 	}
 
+	user, err := c.User()
+	if err != nil {
+		return err
+	}
+
+	c.TelemetryService.Setup(c.Name)
```
makesitgo, line 219: `c.TelemetryService.Setup(c.Name)` can still be moved up higher in this code chain (in case an `error` is returned from `run() error` here early, we'll at least get "command name" in some of the event properties

I'd place it b/w the version check and the storage check in this method

commands/diff_test.go
makesitgo: looks like there's still an extra newline here on L#8 (`"github.com/10gen/realm-cli/utils/telemetry"` can be grouped with the other "project imports" in the middle group here)

utils/telemetry/events.go
makesitgo: can we match these keys to what's in `master`?  looks like instead of `command` we use `cmd`, and instead of `version` we use `v`

utils/telemetry/events.go
```diff
@@ -0,0 +1,41 @@
...
+	Value interface{}
+}
+
+// EventType signifies where in the command this event occurred
+type EventType string
+
+// set of supported EventTypes
+const (
+	EventTypeCommandStart EventType = "COMMAND_START"
+	EventTypeCommandEnd   EventType = "COMMAND_END"
```
makesitgo, line 30: likewise to match `master` ... we use `COMMAND_COMPLETE`, so let's update this value (and `var` name here(

makesitgo (commented):
one last overarching nit here would be: `emptyService` still isn't a sufficient name in those tests.. "service" is still too generic when we're outside of the `telemetry` package.. any references to that `Service` struct outside the `telemetry` package should still include some indication it a _telemetry service_

some thoughts for alternative names:
- `telemetryService` (no need to add extra adjectives at all in a test, it's simple and straightforward enough to be self-explanatory)
- `mockTelemetryService` (its a "mock implementation", though i'd argue its more of a "no-op" than a "mock"
- `telemetryServiceNoop` / `noopTelemetryService` ("idiomatic go" prefers the former over the latter)
- `telemetryServiceEmpty` / `telemetryServiceZero` / `telemetryServiceNil` (empty/zero/nil are common choices in this case, but again "no-op" feels more appropriate than any of these)

since that's such a widespread change, I'll withhold the the 👍  for merging until after that, but really this is good to go after that

makesitgo (approved):
looks good to me... nice work on cleaning all this up!

we should reach out to chris to coordinate making a support release for this work (and then also work with sumedha to verify this is actually working after the release)

🆒  beans

PR #117, opened 2021-03-30
REALMC-7173: Buildout command: app describe

nathantfrank:


internal/commands/app/describe.go
```diff
@@ -0,0 +1,65 @@
+package app
+
+import (
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/spf13/pflag"
```
kpatel71716, line 6: [nit] newline above this import

internal/commands/app/describe_test.go
```diff
@@ -0,0 +1,351 @@
...
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/cloud/atlas"
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/10gen/realm-cli/internal/utils/test/mock"
+	"github.com/Netflix/go-expect"
```
kpatel71716, line 13: [nit] newline above this import

internal/cloud/realm/app_test.go
```diff
@@ -34,13 +35,37 @@ func TestRealmApps(t *testing.T) {
...
+					DataSources:    []realm.DataSourceSummary{},
+					HTTPEndpoints:  []realm.HTTPEndpointSummary{},
+					ServiceDescs:   []realm.ServiceSummary{},
+					AuthProviders:  []realm.AuthProviderSummary{{"api-key", "api-key", false}},
+					CustomUserData: realm.CustomUserDataSummary{},
+					Values:         []string{},
+					Functions:      []realm.FunctionSummary{},
+					Sync:           realm.SyncSummary{},
+					GraphQL: realm.GraphQLSummary{
+						URL:             fmt.Sprintf("%s/api/client/v2.0/app/%s/graphql", u.RealmServerURL(), app.ClientAppID),
```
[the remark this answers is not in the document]
nathantfrank: I have discovered the issue. If we want the server to run on port 9090 in the `test_cli_config.json` we need to add `"api": {"listenAddresses": [{"hostname": "0.0.0.0", "port": 9090}]` to the file. Currently, it is defaulting to port 8080, and since the `baseUrl`s appear to not affect the listening endpoints and therefore cause the difference. The reason I'm assuming we haven't crossed this issue until now is that the `.evg.yml` is targetting the defaulted port of 8080. So either way, the `test_cli_config.json` and the `.evg.yml` need to be updated to be consistent with the port we want to run against.

internal/cloud/realm/app.go
```diff
@@ -11,10 +11,98 @@ import (
...
+type AppDescription struct {
+	ClientAppID       string                     `json:"client_app_id"`
+	Name              string                     `json:"name"`
+	RealmURL          string                     `json:"realm_url"`
+	DataSources       []DataSourceSummary        `json:"data_sources"`
+	HTTPEndpoints     []HTTPEndpointSummary      `json:"http_endpoints"`
+	ServiceDescs      []ServiceSummary           `json:"services"`
+	AuthProviders     []AuthProviderSummary      `json:"auth_providers"`
+	CustomUserData    CustomUserDataSummary      `json:"custom_user_data"`
+	Values            []string                   `json:"values"`
```
kpatel71716, line 97: [q] Are we intentionally showing only the names of values here? (i.e. not secrets or environment variables)?

kpatel71716 (commented):
Looks really good to me. Just a small question and some really small nits

internal/cloud/realm/app.go
nathantfrank: That's correct, just supposed to be kinda like a summary rather than an in-depth look is my understanding.

internal/cloud/realm/app.go
```diff
@@ -11,10 +11,98 @@ import (
...
 const (
-	appsPathPattern = adminAPI + "/groups/%s/apps"
-	appPathPattern  = appsPathPattern + "/%s"
+	appsPathPattern           = adminAPI + "/groups/%s/apps"
+	appPathPattern            = appsPathPattern + "/%s"
+	appDescriptionPathPattern = appPathPattern + "/description"
 )
 
+// DataSourceSummary is a short summary for a data source model
+type DataSourceSummary struct {
```
makesitgo, line 20: [just a thought] might be nice to move this stuff to its own `app_description.go` file ... just reduce the overhead in this one file (it was already pretty hefty with the `Find` logic)

internal/commands/app/describe.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// set of supported `app describe` command strings
+const (
+	CommandDescribeUse     = "describe"
```
makesitgo, line 11: if these are simply only used in the `commands/commands.go` file, then we don't need consts for them (those consts only really became necessary when we were trying to print out suggested commands for any log messages)

internal/commands/app/describe.go
```diff
@@ -0,0 +1,65 @@
...
+	CommandDescribeDisplay = "app describe"
+)
+
+// set of supported `app describe` command strings
+var (
+	CommandDescribeAliases = []string{}
```
makesitgo, line 17: can remove this as well

internal/commands/app/describe.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type describeInputs struct {
+	cli.ProjectInputs
+}
+
+func (i *describeInputs) Resolve(profile *cli.Profile, ui terminal.UI) error {
+	return nil
```
makesitgo, line 25: do we need to be invoking `i.ProjectInputs.Resolve(ui, profile.WorkingDirectory)` in this method?  if so, there's likely a few other spots we may have missed this (`app diff` perhaps, maybe some others?)

we should audit this real quick and make sure we're resolving the project inputs where need be

internal/commands/app/describe.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Inputs is the command inputs
+func (cmd *CommandDescribe) Inputs() cli.InputResolver {
+	return &cmd.inputs
+}
+
+// Handler is the command handler
+func (cmd *CommandDescribe) Handler(profile *cli.Profile, ui terminal.UI, clients cli.Clients) error {
+	if cmd.inputs.Project == "" && cmd.inputs.App == "" {
```
makesitgo, line 45: I wonder if we have the `ProjectInputs.Resolve` in the `describeInputs.Resolve` implementation, if this is necessary anymore?

internal/commands/commands.go
```diff
@@ -130,6 +130,15 @@ Realm apps associated with your user profile.`,
...
 The list includes Realm apps from all projects associated with your user profile.`,
 			},
+			{
+				Command:     &app.CommandDescribe{},
+				Use:         app.CommandDescribeUse,
+				Display:     app.CommandDescribeDisplay,
+				Description: "View all of the configured and enabled aspects of your Realm app",
+				Help: `Displays information about your Realm app.
+If you have more than one Realm app, you will be prompted to select a Realm app that
+you would like to view from a list of all Realm apps associated with your user profile.`,
```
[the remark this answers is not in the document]
makesitgo: seems fine for now, I agree we don't need to be describing the type of output (tables or w/e) in this portion here

makesitgo (commented):
didn't really go through the tests just yet (I want to understand more about the evg.yml/server config stuff first), but thought I'd give you some early feedback here

internal/commands/app/describe.go
nathantfrank: I should probably do something in between the current implementation and `i.ProjectInputs.Resolve(ui, profile.WorkingDirectory)`. I didn't go with the latter, because it forces app input to be set and therefore never lets a user resolve the app to describe by Project

nathantfrank: I guess what should be asked is what is our default no flags behavior?

internal/cloud/realm/app_test.go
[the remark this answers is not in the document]
makesitgo: let's update the `test_cli_config.json` with the `listenAddresses` like you suggest here, which then if I'm understanding properly, after we merge that we'll need to make the corresponding changes here in the `.evg.yml`

good chance I missed something here though, lmk if that doesn't work

internal/commands/app/describe.go
makesitgo: discussed offline, but we can tweak that `ProjectInputs.Resolve` signature to allow us to skip prompting for an app filter, which is something we can leverage here now (and avoid the `GroupID` resolution below in the `Handler`)

I'll also file a followup ticket for us to do an audit of where we are (and aren't but should be) using `ProjectInputs.Resolve` and we can work with @kpham1231 to figure out the right messaging for our long-form help texts that help identify the expected behavior for each relevant command

internal/commands/app/describe.go
makesitgo: I still don't think we need this bit... `cli.ResolveApp` handles the case where the `filter.GroupID` is blank

internal/commands/app/describe.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return err
+	}
+
+	appDesc, err := clients.Realm.AppDescription(app.GroupID, app.ID)
+	if err != nil {
+		return err
+	}
+
+	ui.Print(terminal.NewJSONLog("App Description", appDesc))
```
makesitgo, line 52: [nit] none of our other messages really ever go with "Title Casing for the Log Messages", lets use something like `NewJSONLog("App description:", appDesc)` here (or `"The following describes your Realm app:"`)

internal/commands/app/describe.go
```diff
@@ -0,0 +1,55 @@
...
+
```
makesitgo, line 53: [ubernit] remove this newline

internal/commands/app/describe_test.go
```diff
@@ -0,0 +1,380 @@
...
+	"github.com/10gen/realm-cli/internal/cloud/atlas"
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	"github.com/10gen/realm-cli/internal/local"
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/10gen/realm-cli/internal/utils/test/mock"
+
+	"github.com/Netflix/go-expect"
+)
+
+func TestDescribeInputsResolve(t *testing.T) {
```
makesitgo, line 20: [nit] I probably havent done a great job enforcing this, but best to name these tests in a cascading fashion so the regex targeting can be a bit more effective: `TestAppDescribe...`

internal/commands/app/describe_test.go
```diff
@@ -0,0 +1,380 @@
...
+	t.Run("should set app if in app directory", func(t *testing.T) {
+		profile, teardown := mock.NewProfileFromTmpDir(t, "app-describe-test")
+		defer teardown()
+
+		assert.Nil(t, ioutil.WriteFile(
+			filepath.Join(profile.WorkingDirectory, local.FileRealmConfig.String()),
+			[]byte(`{"config_version":20210101,"app_id":"test-app-abcd","name":"test-app"}`),
```
makesitgo, line 27: i cant help it, but our client app ids have a 5-character suffix to them
```suggestion
			[]byte(`{"config_version":20210101,"app_id":"test-app-abcde","name":"test-app"}`),
```

internal/commands/app/describe_test.go
```diff
@@ -0,0 +1,380 @@
...
+				},
+				Hosting: realm.HostingSummary{
+					Enabled: true,
+					Status:  "setup_ok",
+					URL:     "https://hosting.domain/",
+				},
+				Functions: []realm.FunctionSummary{
+					{
+						Name: "func1",
+						Path: "func1.js",
```
makesitgo, line 194: unrelated to the test functionality, but `path` here is intended to be blank, and below it'd just be `nested`

internal/commands/app/describe_test.go
```diff
@@ -0,0 +1,380 @@
...
+      "name": "trigger1",
+      "type": "DATABASE",
+      "enabled": true
+    }
+  ]
+}
+`, out.String())
+	})
+
+	t.Run("should error if apps not found", func(t *testing.T) {
```
makesitgo, line 322: is this test case duped?

internal/commands/app/describe_test.go
```diff
@@ -0,0 +1,380 @@
...
+		}
+
+		cmd := &CommandDescribe{inputs: describeInputs{cli.ProjectInputs{App: "test-app"}}}
+		assert.Equal(t, cli.ErrAppNotFound{App: "test-app"}, cmd.Handler(nil, nil, cli.Clients{Realm: realmClient}))
+	})
+
+	for _, tc := range []struct {
+		description   string
+		inputs        describeInputs
+		groups        []atlas.Group
```
makesitgo, line 336: i think we should probably remove this test table, and just write the individual error test cases separately.  some of these test case struct fields appear to be unused, and the test cases could be elaborated a bit more to explain what each test is actually doing (i.e. `"should return an error when finding atlas groups fails"`, etc)

makesitgo (approved):
just my last few thoughts here, really only minor clean up from here.  once thats done and evergreen is happy, this is good to go!

🆒  beans

internal/commands/app/describe_test.go
nathantfrank: Yeah, it totally is.

PR #118, opened 2021-04-05
REALMC-8613: Remove timestamp and log level from text messages

makesitgo:
For context here, with trying to display logs output, timestamps were getting a little noisy/difficult to discern which were "command timestamps" and which were "server timestamps".  The log level was probably never all that important (although maybe we need some differentiator for ERROR?).

This has been run by @sumedham and @drewdipalma ... though maybe they have further thoughts regarding the need to differentiate errors at all

These fields are still exported as the JSON/structured log message.  The log level also can be useful down the line for niche formatting (error in red, warn in yellow, etc) as well as a potential verbose/quiet flag.

### DIFF
--- a/internal/cli/command_factory.go
+++ b/internal/cli/command_factory.go
@@ -25,23 +25,17 @@ type CommandFactory struct {
 	inReader         *os.File
 	outWriter        *os.File
 	errWriter        *os.File
-	errLogger        *log.Logger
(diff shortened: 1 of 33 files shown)


UnicodeSnowman (approved):

nathantfrank (approved):
Like the change!

makesitgo:
to be merged after #119

PR #121, opened 2021-04-06
REALMC-8242: Support dependencies with app diff command

nathantfrank:


internal/commands/push/command.go
```diff
@@ -126,8 +126,38 @@ func (cmd *Command) Handler(profile *cli.Profile, ui terminal.UI, clients cli.Cl
 		diffs = append(diffs, appDiffs...)
 
 		if cmd.inputs.IncludeDependencies {
-			// TODO(REALMC-8242): diff dependencies better
-			diffs = append(diffs, "+ New function dependencies")
+			dependencies, err := local.FindAppDependencies(app.RootDir)
```
makesitgo, line 129: so we need to collect diffs for the dependencies before this point (similar to how hosting diffs are handled)

that is because if we find no diffs, then we know we don't need to do anything for this import and return early

makesitgo (commented):
nice start here... i think the biggest thing is pushing the dependencies diff logic higher in the `push` command chain.  feel free to rebase and get the latest log message formats now too (and force push if need be)

internal/commands/app/diff_test.go
```diff
@@ -97,27 +98,61 @@ func TestAppDiffHandler(t *testing.T) {
 	}
 
 	t.Run("with include dependencies set should diff function dependencies", func(t *testing.T) {
-		out, ui := mock.NewUI()
+		out := new(bytes.Buffer)
+		ui := mock.NewUIWithOptions(mock.UIOptions{AutoConfirm: true}, out)
```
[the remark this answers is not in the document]
nathantfrank: For some reason, I had it in my head that the spinner required `AutoConfirm: true`. Not actually needed.

internal/commands/app/diff_test.go
```diff
@@ -97,27 +98,61 @@ func TestAppDiffHandler(t *testing.T) {
...
+Removed Dependencies
+	- debug@4.3.1
+Modified Dependencies
+	* underscore@1.9.1 -> underscore@1.9.2
 `, out.String())
 	})
 
+	t.Run("should return error with include dependencies set and diff dependencies returns an error", func(t *testing.T) {
+		out := new(bytes.Buffer)
+		ui := mock.NewUIWithOptions(mock.UIOptions{AutoConfirm: true}, out)
```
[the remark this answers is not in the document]
nathantfrank: Not Actually needed.

internal/commands/push/command.go
```diff
@@ -97,6 +97,20 @@ func (cmd *Command) Handler(profile *cli.Profile, ui terminal.UI, clients cli.Cl
...
+		uploadPathDependencies, err = local.PrepareDependencies(app, ui)
+		if err != nil {
+			return err
+		}
+		dependenciesDiff, err = clients.Realm.DiffDependencies(appRemote.GroupID, appRemote.AppID, uploadPathDependencies)
+		if err != nil {
+			return err
+		}
+	}
+	defer os.Remove(uploadPathDependencies) //nolint:errcheck
```
makesitgo, line 112: i feel like this `defer` should be moved after the first `err` check inside the `IncludeDependencies` block

internal/commands/push/command.go
```diff
@@ -115,7 +129,7 @@ func (cmd *Command) Handler(profile *cli.Profile, ui terminal.UI, clients cli.Cl
 		}
 	}
 
-	if len(appDiffs) == 0 && !cmd.inputs.IncludeDependencies && hostingDiffs.Size() == 0 {
+	if len(appDiffs) == 0 && (!cmd.inputs.IncludeDependencies || !dependenciesDiff.HasChanges()) && hostingDiffs.Size() == 0 {
```
makesitgo, line 132: couldn't something like this suffice as a check?  (the usage of `IncludeDependencies` here was a stopgap b/c we had otherwise no data relevant to how the dependences had changed)
```suggestion
	if len(appDiffs) == 0 && dependenciesDiffs.Size() == 0 && hostingDiffs.Size() == 0 {
```

[naming nit] would call it `dependenciesDiffs`, or switch the other two to be singule with `*Diff`...

internal/local/dependencies.go
```diff
@@ -8,6 +8,10 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"time"
+
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/briandowns/spinner"
```
makesitgo, line 14: [nit] newline between these

internal/terminal/table.go
```diff
@@ -14,8 +14,12 @@ const (
 
 	// gutter is the gap between table columns
 	gutter = "  "
-	// indent is the indentation used
-	indent = "  "
+)
+
+// set of supported spacing options
+const (
+	// Indent is the indentation used
```
makesitgo, line 21: [optional] to me, the `// set of ...` comment here suffices for docstrings and you could remove this one here

i would also probably move `gutter` down into this block (rather than above), just since its also a "spacing option" (supported =/= exported, i think we're fine here by the letter of the law 🙃 )

makesitgo (approved):
last few things here, but this looks good to me

🆒  beans

PR #122, opened 2021-04-06
REALMC-8498: Move relationships into its own json file under data_sources

nathantfrank:
Appeared to be pretty straightforward. Ensure relationships key exists in the V2 `DataSourceStructure.Rules` and always write the file out to disk.


internal/cloud/realm/import_export_test.go
```diff
@@ -297,35 +297,18 @@ func appDataV2(app realm.App) local.AppDataV2 {
...
-				Rules: []map[string]interface{}{{
-					"name":    "rule",
-					"actions": []interface{}{"get"},
-					"when": map[string]interface{}{`%%args.url.host`: map[string]interface{}{
-						`%in`: []interface{}{"google.com"},
-					}},
-				}},
+				Config:           map[string]interface{}{"name": "api", "type": "http", "config": map[string]interface{}{}},
+				IncomingWebhooks: []map[string]interface{}{},
+				Rules:            []map[string]interface{}{},
```
nathantfrank, line 311: Pretty dramatic changes to what is being returned now from the api. Not sure if this logic is completely intentional.

makesitgo: yeah this doesn't look right, why did you need to change it?  can you rebase the CLI and make sure you're off of fresh `master` with `baas`?

internal/local/structure_v2.go
```diff
@@ -536,6 +560,13 @@ func writeDataSources(rootDir string, dataSources []DataSourceStructure) error {
 			); err != nil {
 				return err
 			}
+			if err := WriteFile(
+				filepath.Join(dir, name, fmt.Sprintf("%s", rule["database"]), fmt.Sprintf("%s", rule["collection"]), FileRelationships.String()),
```
makesitgo, line 564: I didn't notice these before, but we can avoid using `fmt.Sprintf("%s"...` to cast to strings (especially for known fields `database` and `collection` here).

considering we need db and coll for 3 different filepaths here, lets DRY this up a bit:

```go
var database, collection string
if db, ok := rule["database"]; ok {
  if db, ok := db.(string); ok { database = db }
}
// same for collection
ruleDir := filepath.Join(dir, name, database, collection)
```

internal/local/structure_v2.go
```diff
@@ -509,33 +521,64 @@ func writeDataSources(rootDir string, dataSources []DataSourceStructure) error {
...
 			if err != nil {
 				return err
 			}
+			var database, collection string
+			if db, ok := rule["database"]; ok {
+				if db, ok := db.(string); ok {
+					database = db
+				}
+			}
+			if col, ok := rule["collection"]; ok {
```
makesitgo, line 555: [its stupid but] `coll` instead of `col` 😬  :bushes:

makesitgo (approved):
:cool: beans

PR #126, opened 2021-04-08
REALMC-8641: Revert the use of dependencies/diff admin endpoint

makesitgo:


UnicodeSnowman (approved):
you could also just revert the commit that added these changes (and then later revert that revert to get back), but however you want to handle it is cool with me 👍

makesitgo:
didn't think about that approach, but i think ive got this covered so i'll stick with this

PR #127, opened 2021-04-08
REALMC-7180: Build logs list command

makesitgo:
Going to implement "tailing" (read: polling the logs endpoint) in a separate, follow-up ticket


internal/cloud/realm/logs.go
```diff
@@ -0,0 +1,121 @@
...
+	LogTypeStreamFunction        = "STREAM_FUNCTION"
+	LogTypeSyncClientWrite       = "SYNC_CLIENT_WRITE"
+	LogTypeSyncConnectionEnd     = "SYNC_CONNECTION_END"
+	LogTypeSyncConnectionStart   = "SYNC_CONNECTION_START"
+	LogTypeSyncError             = "SYNC_ERROR"
+	LogTypeSyncOther             = "SYNC_OTHER"
+	LogTypeSyncSessionEnd        = "SYNC_SESSION_END"
+	LogTypeSyncSessionStart      = "SYNC_SESSION_START"
+	LogTypeWebhook               = "WEBHOOK"
+)
```
tsedgwick, line 49: is this necessary? it would be really nice to be able to control this from baas since logging is more fluid in nature of adding new types.

tsedgwick (commented):
one over-arching question here on if we can push more logic into baas instead of having it here. if not, we'll have to remember to add log type to the cli with every new type we introduce.

internal/cloud/realm/logs.go
makesitgo: based on the way our logs api works currently, yes.  the directive from product here has been "keep it like the UI" and for better or worse, a lot of this logic is pushed to the frontend (so we're more or less duplicating it here)

ive communicated this to drew and sumedha that should we ever make changes to log types and/or names, we have multiple places to remember to update.  hopefully this is something we can address with the logs rework

tsedgwick: I'm fine doing it this way for now, but this will add to the logs project scope coming up.

tsedgwick (approved):

internal/cloud/realm/logs_test.go
```diff
@@ -0,0 +1,100 @@
...
+			logs, err := client.Logs(groupID, app.ID, realm.LogsOptions{})
+			assert.Nil(t, err)
+			assert.Equal(t, []realm.Log{{Messages: []interface{}{"one", "two", "three"}}}, logs)
+		})
+	})
+}
+
+// insertTestLogs will insert logs to the Realm server backing db for the provided app
+// in a way such that API calls will be able to retrieve the app's associated log documents
+func insertTestLogs(t *testing.T, mongoClient *mongo.Client, app realm.App, logMessages ...[]string) func() {
```
UnicodeSnowman, line 57: I don't love the fact that we're explicitly inserting data into the (private) server db here in a client/CLI test. This is brittle and certainly has the potential to change out from under us and break here despite the behavior we're really testing here (i.e. can we fetch logs via this command?) remaining unchanged.

since this test already depends on cloud, could we just fetch some logs from an existing app (maybe the app we're using here already exists, I can't tell) and assert that we get some data back? I suppose this also has the potential to break since we delete logs after N days, but we could also fire off some actions via the public client API beforehand (to populate log data) which by definition cannot break/change.

internal/commands/commands.go
```diff
@@ -237,16 +238,38 @@ be modified.`,
...
  - A list of error logs, if present
 `,
+			},
+		},
+	}
+
+	Logs = cli.CommandDefinition{
+		Use:         "logs",
+		Aliases:     []string{"log"},
+		Description: "Interact the logs of your Realm app",
```
UnicodeSnowman, line 263: looks like we need a "with" here

internal/commands/logs/list_inputs.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/10gen/realm-cli/internal/utils/flags"
+)
+
+const (
+	flagType      = "type"
+	flagTypeUsage = "specify the type(s) of logs to list, available options: []"
```
UnicodeSnowman, line 12: no available options here? or does `[]` get populated by the CLI lib somehow?

internal/cloud/realm/logs_test.go
makesitgo: I spent a long time trying to get client calls to register logs without any success, but I agree that would be the ideal route.  I couldn't tell if I just wasn't issuing commands that created logs (I was executing functions that invoked `console.log`) or if I couldn't get the logs to actually flush to DB, but it was proving difficult for me.  Any thoughts/tips here?

Honestly, I wrote this mostly as a sanity check for myself to make sure the proper api endpoint was hit (I intentionally didn't go into testing start/end dates and the other filters).  I'd be happy to remove it entirely as well (the CLI's not really in the business of being the integration tester for our admin api anyways).

If you think we can get the client calls and log flushing work, I'm all ears there (but ideally in a follow up ticket to this)

internal/commands/commands.go
makesitgo: I missed the other help texts as well for the subcommand

internal/commands/logs/list_inputs.go
makesitgo: i wish, flags (and usages, etc) are something id love to find an elegant solution for at some point... but i just forgot to add these (needed to copy them from the UI first)

internal/cloud/realm/logs_test.go
UnicodeSnowman: hmm yeah true, flushing should happen within a few seconds, but there's no guarantee of _when_ it will happen (or at least no way to wait, other than looping and making the same request multiple times before timing out, etc.). personally, I think a smoke-test here of just making sure the endpoint exists and doesn't blow up (even if it returns no results) is sufficient, especially since we already have a ton of baas server tests for the logs endpoint (and likely some UI E2E tests, if I remember correctly).

UnicodeSnowman (approved):

internal/cloud/realm/logs_test.go
makesitgo: that sounds good to me, i'll remove the log inserts and asserts on non-empty result sets (and leave the empty/smoke test)

PR #128, opened 2021-04-09
REALMC-7179: Implement tailing for logs list

makesitgo:


internal/commands/logs/list.go
```diff
@@ -38,25 +44,79 @@ func (cmd *CommandList) Inputs() cli.InputResolver {
...
+	logsCh, errCh, closeCh := make(chan realm.Logs), make(chan error, 1), make(chan struct{}, 1)
+
+	opts.Start = cmdStart
+	go pollForLogs(clients.Realm, app.GroupID, app.ID, opts, logsCh, errCh, closeCh)
+
+	for {
+		select {
+		case logs := <-logsCh:
+			printLogs(ui, logs)
+		case err := <-errCh:
```
UnicodeSnowman, line 86: do we need to `close(closeCh)` here too?

internal/commands/logs/list.go
```diff
@@ -38,25 +44,79 @@ func (cmd *CommandList) Inputs() cli.InputResolver {
...
+		case <-cmd.inputs.sigShutdown:
+			close(closeCh)
+			return nil
+		}
+	}
+	return nil
+}
+
+func pollForLogs(realmClient realm.Client, groupID, appID string, opts realm.LogsOptions, logsCh chan<- realm.Logs, errCh chan<- error, closeCh <-chan struct{}) {
+	ticker := time.NewTicker(5 * time.Second)
```
UnicodeSnowman, line 97: this feels like something we could (eventually) make configurable

internal/commands/logs/list.go
```diff
@@ -38,25 +44,79 @@ func (cmd *CommandList) Inputs() cli.InputResolver {
...
+		case <-closeCh:
+			return
+		case <-ticker.C:
+			startNext := time.Now()
 
+			logs, err := realmClient.Logs(groupID, appID, opts)
+			if err != nil {
+				errCh <- err
+			}
+			logsCh <- logs
```
UnicodeSnowman, line 111: seems like we could have a potential race between taking from `logsCh` and `errCh` in the `select...for` loop on line 82, but maybe that's okay (since if `logsCh` happens to take first it'll just loop back to taking from `errCh` and then exit). just confirming the intended behavior here.

or potentially... if the `errCh` take happens first, we exit, and then nothing is pulling from `logsCh`, so this goroutine will block forever. you might need to make sure you drain `logsCh` in that case. or potentially just return in the error case to exit.

in reality this probably doesn't matter, since the whole program will shut down and exit, but I still think it's good to get in the habit of making sure we're always managing goroutine exit conditions properly.

UnicodeSnowman (approved):
LGTM overall, but let's make sure we're thinking about exit conditions for these goroutines.

internal/commands/logs/list.go
makesitgo: yeah, probably.  i'll add it as a defer above all of this

internal/commands/logs/list.go
makesitgo: agreed, I'll keep it 5 for now but it would be an easy flag to add (we can pull design/product in for the right messaging)

internal/commands/logs/list.go
makesitgo: discussed offline, adding a `return` after sending on the `errCh`, and double-checking the `closeCh` before sending on `logsCh` to prevent this goroutine from hanging

PR #133, opened 2021-04-21
REALMC-8731 Check for null tracker for segment tracking and add version

jo-tse-mongo:


commands/command.go
```diff
@@ -219,6 +219,7 @@ func (c *BaseCommand) run(args []string) error {
 	}
 
 	c.TelemetryService.SetUser(user.PublicAPIKey)
+	c.TelemetryService.SetVersion(utils.CLIVersion)
```
makesitgo, line 222: do we need to "set" this explicitly?  or can it just be referenced directly when setting the "version" value on `Track` calls?

utils/telemetry/service.go
```diff
@@ -31,17 +32,25 @@ func (s *Service) SetUser(userID string) {
...
-		data:        data,
-	})
+	if s.tracker != nil {
+		s.tracker.Track(event{
+			id:          primitive.NewObjectID().Hex(),
+			eventType:   eventType,
+			userID:      s.userID,
+			time:        time.Now(),
+			executionID: s.executionID,
+			version:     s.version,
```
makesitgo, line 49: so instead of `s.version`, could this just be `utils.CLIVersion` here?

utils/telemetry/service.go
```diff
@@ -31,17 +32,25 @@ func (s *Service) SetUser(userID string) {
...
-	s.tracker.Track(event{
-		id:          primitive.NewObjectID().Hex(),
-		eventType:   eventType,
-		userID:      s.userID,
-		time:        time.Now(),
-		executionID: s.executionID,
-		command:     s.command,
-		data:        data,
-	})
+	if s.tracker != nil {
```
makesitgo, line 42: instead of the `nil` check here, let's just initialize the telemetry service as `&telemetry.Service{Tracker: &noopTracker{}}` in `main.go`

then in `Setup`, you can also remove the `noopTracker` assignment since it will already be there

makesitgo (approved):
just a few thoughts, but once they're cleaned up:

🆒  beans

PR #136, opened 2021-04-23
REALMC-8543: Implement the schema models command

makesitgo:


internal/cloud/realm/schema.go
```diff
@@ -1,5 +1,60 @@
...
+	Imports   []string           `json:"import_statements"`
+	Code      string             `json:"schema"`
+	Warnings  []SchemaModelAlert `json:"warnings"`
+	Error     SchemaModelAlert   `json:"error"`
+}
+
+type SchemaModelAlert struct {
+	Message string `json:"error"`
+	Code    string `json:"error_code"`
+}
```
nathantfrank, line 39: [q] Understand this is coming from the API so nothing to be done here, but maybe something to keep in mind later. Does it make sense that a `Warning`'s alert is considered an `error` vs `message`?

internal/commands/schema/models_inputs.go
```diff
@@ -0,0 +1,155 @@
...
+			i.NoImports = noImports
+		}
+
+		if !i.Flat {
+			var flat bool
+			if err := ui.AskOne(&flat, &survey.Confirm{Message: "Would you like group all generated data models together?"}); err != nil {
+				return err
+			}
+			i.Flat = flat
+		}
```
nathantfrank, line 70: [q] Going the route of always asking if flag isn't set?

nathantfrank (commented):
Looking good so far, just some Q's and probably should put something for the help text

internal/cloud/realm/schema.go
makesitgo: I'm with you 💯  on this ... I also don't love we use "error code" terminology for a warning, but I think these warnings/errors are considered more supplemental than anything.  this is all simply reverse-engineering & mirroring what the UI does with these things

internal/commands/schema/models_inputs.go
makesitgo: haha yeah, based off the feedback from our past bug bash, it felt like users appreciated more prompts/less flags.  fwiw, this "prompt if not set" behavior still will only occur if the user omits `--language` flag ... if they were to set `--language` but not set `--flat` or `--no-imports`, the prompts would still not pop up

internal/commands/commands.go
```diff
@@ -274,4 +275,20 @@ your logs and follow any newly created logs in real-time.`,
...
+		Aliases:     []string{"schemas"},
+		Description: "Manage the schemas of your Realm app",
+		SubCommands: []cli.CommandDefinition{
+			{
+				Command:     &schema.CommandModels{},
+				Use:         "models",
+				Aliases:     []string{"model"},
+				Display:     "schema models",
+				Description: "Generate source code for the schema data models",
+				Help:        `TODO: this`,
```
UnicodeSnowman, line 290: What's the plan for this TODO?

UnicodeSnowman (approved):

internal/commands/commands.go
makesitgo: was going to reach out to @kpham1231 for this one

PR #137, opened 2021-05-04
REALMC-8399: Check for new cli versions each day per profile

makesitgo:
* Refactor cli.Profile to its own package: cli/user.Profile
  * allows it to be mocked AND tested in the `cli` package
* Remove auth package (no longer needed with cli/user package now)
  * this "other" package was always needed, now it just has the `Profile` in it
* Refactor telemetry service to be interface
  * allows it to be mocked and tested with during the `CommandFactory` version check tests


tsedgwick (approved):
LGTM.

PR #138, opened 2021-05-06
REALMC-8882: Include Atlas product apps with standard apps by default

makesitgo:


internal/cloud/realm/app.go
```diff
@@ -83,20 +84,31 @@ func (c *client) DeleteApp(groupID, appID string) error {
...
 type AppFilter struct {
-	GroupID string
-	App     string // can be client app id or name
+	GroupID  string
+	App      string // can be client app id or name
+	Products []string
 }
 
+const (
+	productEmpty    = ""
```
tsedgwick, line 93: [nit] does this really need to be a const?

internal/cloud/realm/app.go
```diff
@@ -156,3 +195,13 @@ func (c *client) getApps(groupID string) ([]App, error) {
 	}
 	return apps, nil
 }
+
+func appsPathWithProduct(groupID, product string) string {
```
tsedgwick, line 199: can this just be a simple if condition to check if `product != ""`? do we need this as an abstracted func?

tsedgwick (approved):
couple nits, overall LGTM

internal/cloud/realm/app.go
makesitgo: no, i even forgot to use it down below.  i'll remove

internal/cloud/realm/app.go
makesitgo: i put it in a func for some quick unit tests, there's no real straightforward way currently for the CLI to assert on request URLs sent from the realm client (this probably _should_ exist...)

i can remove the func and tests if you feel strongly against it, its relatively simple/straightforward logic.  long-term i think I'd rather put effort into adding an easy way to test the request urls we attempt to hit when sending client requests, maybe as a new-eng/quick win ticket?

tsedgwick: Honestly, I would just remove this - the test doesn't benefit it us much from my perspective. I would opt for a TODO on the longterm approach here instead.

makesitgo: 👍 https://jira.mongodb.org/browse/REALMC-8886

PR #139, opened 2021-05-10
REALMC-8395: Remove pointers from v2 local app structure

makesitgo:


internal/commands/push/inputs.go
```diff
@@ -70,7 +70,7 @@ func (i *inputs) Resolve(profile *user.Profile, ui terminal.UI) error {
 	}
 
 	if i.RemoteApp == "" {
-		i.RemoteApp = app.Option()
+		i.RemoteApp = app.ID()
```
makesitgo, line 73: ⚠️ drive by fix to only resolve app via ID for `realm-cli push` (so that the user can push to create a new app (otherwise it would force the user to select between all apps named similarly to this one)

kpatel71716 (approved):
LGTM!

tsedgwick (approved):

PR #141, opened 2021-05-10
REALMC-8405: Remove unnecessary dependencies from package.json and ./etc

makesitgo:
Some of the removed stuff here I think was mistakenly brought in when merging `next` => `master` (I couldn't find any usages for `./etc/tools`, `./etc/transpiler`, `./testdata`


tsedgwick (approved):
😎

PR #153, opened 2021-05-27
REALMC-8599: Resolve app info when commands are run from app directory

makesitgo:


internal/commands/app/diff.go
```diff
@@ -125,12 +144,34 @@ func (cmd *CommandDiff) Handler(profile *user.Profile, ui terminal.UI, clients c
...
 	if i.LocalPath == "" {
-		i.LocalPath = profile.WorkingDirectory
+		if app.RootDir != "" {
+			i.LocalPath = app.RootDir
+		} else {
+			if err := ui.AskOne(&i.LocalPath, &survey.Input{Message: "App filepath (local)"}); err != nil {
+				return err
+			}
+
+			app, err = local.LoadAppConfig(i.LocalPath)
```
makesitgo, line 165: ⚠️  is this "go"?  I don't love the deeply-nested reassignment to `app` here, but tried to keep it idiomatic.  lmk if anyone sees any alternative approaches here!

internal/commands/function/run_inputs_test.go
```diff
@@ -7,70 +7,90 @@ import (
 	"github.com/10gen/realm-cli/internal/cloud/realm"
 	"github.com/10gen/realm-cli/internal/utils/test/assert"
 	"github.com/10gen/realm-cli/internal/utils/test/mock"
-
-	"github.com/Netflix/go-expect"
+	"github.com/AlecAivazis/survey/v2/terminal"
```
makesitgo, line 10: ⚠️  this file is effectively entirely new, I wouldn't try to "diff" it from the past version too much.  the changes here amount to getting the "app picker/selector" to appear before prompting a user for "function name", and then instead of just an input prompt for "function name" we'll provide a list of functions instead

internal/commands/app/diff.go
```diff
@@ -125,12 +144,34 @@ func (cmd *CommandDiff) Handler(profile *user.Profile, ui terminal.UI, clients c
...
+			if err != nil {
+				return nil
```
tsedgwick, line 167: should this be returning nil?

internal/commands/app/diff.go
tsedgwick: something like this might be cleaner:
```
	searchPath := i.LocalPath
	if searchPath == "" {
		searchPath = profile.WorkingDirectory
	}

	app, err := local.LoadAppConfig(searchPath)
	if err != nil {
		return err
	}

	if i.LocalPath == "" && app.RootDir == "" {
		if err := ui.AskOne(&i.LocalPath, &survey.Input{Message: "App filepath (local)"}); err != nil {
			return err
		}

		app, err = local.LoadAppConfig(i.LocalPath)
		if err != nil {
			return err
		}
	}
	i.LocalPath = app.RootDir

	if i.RemoteApp == "" {
		i.RemoteApp = app.Option()
	}

	return nil
```

internal/commands/app/diff.go
```diff
@@ -125,12 +144,34 @@ func (cmd *CommandDiff) Handler(profile *user.Profile, ui terminal.UI, clients c
...
+	if err != nil {
 		return err
 	}
 
 	if i.LocalPath == "" {
-		i.LocalPath = profile.WorkingDirectory
+		if app.RootDir != "" {
+			i.LocalPath = app.RootDir
+		} else {
+			if err := ui.AskOne(&i.LocalPath, &survey.Input{Message: "App filepath (local)"}); err != nil {
```
tsedgwick, line 161: should we be checking if this `i.LocalPath != ""` or will the AskOne take care of that?

makesitgo: the larger if block around this is `i.LocalPath == ""`, so we're only setting `LocalPath` if it wasn't set

tsedgwick: I meant afterwards when we try to use it to get the app. Is the error user friendly when the user inputs an empty string here on the prompt?

internal/commands/app/diff.go
tsedgwick: app isn't a pointer though so it would just default to an empty string

we have logic in `.Option()`:
```
func (a App) Option() string {
	if a.AppData == nil {
		return a.RootDir
	}
	if a.AppData.ID() != "" {
		return a.AppData.ID()
	}
	return a.AppData.Name()
}
```

makesitgo: forgot I added nil checks in that function... go me

internal/commands/app/diff.go
makesitgo: yeah we return the error in the handler:

```go
	if app.RootDir == "" {
		return fmt.Errorf("no app directory found at %s", cmd.inputs.LocalPath)
	}
```

nathantfrank (approved):
n🧊 !

tsedgwick (approved):
😎

jandersongo (approved):

PR #154, opened 2021-05-28
REALMC-8944: Use default urls if not previously set in profile

makesitgo:
its a messy diff, but basically:

* `git mv internal/cli/user/profile_test.go internal/cli/user/profile_ext_test.go`
* entirely new test file: `internal/cli/user/profile_test.go`

### DIFF
--- a/internal/cli/user/profile_test.go
+++ b/internal/cli/user/profile_test.go
@@ -1,57 +1,58 @@
-package user_test
+package user
 
 import (
-	"fmt"
-	"io/ioutil"
-	"strings"
 	"testing"
 
-	"github.com/10gen/realm-cli/internal/cli/user"
-	u "github.com/10gen/realm-cli/internal/utils/test"
+	"github.com/10gen/realm-cli/internal/telemetry"
 	"github.com/10gen/realm-cli/internal/utils/test/assert"
+
+	"go.mongodb.org/mongo-driver/bson/primitive"
 )
 
-func TestProfile(t *testing.T) {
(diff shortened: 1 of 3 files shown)


internal/cli/user/profile_test.go
```diff
@@ -1,57 +1,58 @@
...
-	profile, profileErr := user.NewDefaultProfile()
-	assert.Nil(t, profileErr)
+		assert.Nil(t, profile.ResolveFlags())
 
-	t.Run("Should initialize as an empty, default profile", func(t *testing.T) {
-		assert.Equal(t, user.DefaultProfile, profile.Name)
-		assert.Equal(t, tmpDir+"/.config/realm-cli", profile.Dir())
-	})
+		assert.Equal(t, telemetry.ModeEmpty, profile.Flags.TelemetryMode)
+		assert.Equal(t, telemetry.ModeEmpty, profile.TelemetryMode())
```
jandersongo, line 25: question out of curiosity: what is this TelemetryMode?

jandersongo (approved):

internal/cli/user/profile_test.go
makesitgo: the "telemetry" for the CLI is what connects to our Segment service which then sends events as data.  these events have data like `userId`, `command`, etc. so product can better understand how people are using our tools.  we do similar tracking in the UI (for various button clicks) and the backend as well.

the "mode" for telemetry here is basically allowing people to turn that tracking on/off.  similar to those popups you occasionally see asking you to allow an app to report usage data and metrics.

nathantfrank (approved):
n🧊 !

PR #155, opened 2021-05-28
REALMC-8824: Ensure command factory closes properly before exiting cli

makesitgo:
i'll switch `*CommandFactory.Run` to return `error` if you still don't like this, I just wanted one last chance to defend the semantics of this code (why return `err` if you shouldn't do anything with it, as its already been consumed/reported?)

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -2,6 +2,8 @@ package cmd
 
 import (
 	"fmt"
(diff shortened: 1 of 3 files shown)


tsedgwick (approved):
I won't split hairs for this, as long as this is the only place we are returning ints.

PR #156, opened 2021-06-01
REALMC-8980: set up CLI commands for IP Access

asudhakal:
Set up bare bones CLI commands for IP Access and testing suite.


internal/cloud/realm/ip_access.go
```diff
@@ -0,0 +1,105 @@
...
+	defer res.Body.Close()
+
+	var allowedIPs []AllowedIP
+	if err := json.NewDecoder(res.Body).Decode(&allowedIPs); err != nil {
+		return nil, err
+	}
+	return allowedIPs, nil
+}
+
+type allowedIPsPayload struct{}
```
asudhakal, line 46: will wait until the API is finished to populate this if that's okay

cmd/root.go
```diff
@@ -47,5 +47,8 @@ func Run() {
 	cmd.AddCommand(factory.Build(commands.Function))
 	cmd.AddCommand(factory.Build(commands.Schema))
 
+	// Uncomment once backend API and data models are complete
```
Gabri3l, line 50: [q] Is this something you'll do before merging this PR ?

asudhakal: We will most likely come back and uncomment this once we've deployed the fully implemented data models/API in the back end. Technically this could be uncommented right now, but runs the risk of having functionality within the CLI (although hidden) that isn't yet available. @makesitgo feel free to chime in!

Gabri3l: I would either associate this to a ticket or create a small ticket for this, for TODOs that need to be addressed after the PR is merged we do:

`// TODO(REALMC-XXXX): do the thing`

makesitgo: I imagine there's a point where you'll want this set of commands to be included in the CLI but things aren't yet fully released to prod?

if that's the case (maybe for testing purposes), you'll want to also mark this command as `Hidden`.  I would add that to the commented code here, just so we don't forget down the line.  Extra credit for filing the _other_ JIRA ticket to cover "un-hiding" the command and adding _that_ `TODO` as well...

internal/commands/ip_access/create.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/cli/user"
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/spf13/pflag"
+)
+
+// CommandMetaCreate is the command meta for the `ip-access create` command
+var CommandMetaCreate = cli.CommandMeta{
+	Use:     "create",
+	Aliases: []string{"add"},
```
makesitgo, line 13: no other "create" command yet uses `"add"` as an alias, is this coming from product or somewhere?  if so, we should standardize that across all the "create" commands, not just here

internal/commands/ip_access/delete.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/cli/user"
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/spf13/pflag"
+)
+
+// CommandMetaDelete is the command meta for the `ip-access delete` command
+var CommandMetaDelete = cli.CommandMeta{
+	Use:     "delete",
+	Display: "allowed IP delete",
```
makesitgo, line 13: use `"ip-access delete"` here

---

also need to add `HelpText` and `Description` here to be complete, work with @kpham1231 if need be

internal/commands/ip_access/update.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/cli/user"
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/spf13/pflag"
+)
+
+// CommandMetaUpdate is the command meta for the `ip-access update` command
+var CommandMetaUpdate = cli.CommandMeta{
+	Use:     "update",
+	Display: "allowed IP update",
```
makesitgo, line 13: use `"ip-access update"` here

---

also need to add `HelpText` and `Description` here to be complete, work with @kpham1231 if need be

internal/utils/test/mock/realm_client.go
```diff
@@ -61,6 +61,11 @@ type RealmClient struct {
 
 	SchemaModelsFn func(groupID, appID, language string) ([]realm.SchemaModel, error)
 
+	AllowedIPsFn      func(groupID, appID string) ([]realm.AllowedIP, error)
+	CreateAllowedIPFn func(groupID, appID, allowedIP, comment string) (realm.AllowedIP, error)
```
makesitgo, line 65: If you go with the naming switch in the client, make sure to update these as well.  It would also seem that you still need to implement the mocked functions here

internal/commands/ip_access/create.go
```diff
@@ -0,0 +1,52 @@
...
+	Display: "allowed IP create",
+	Hidden:  true,
+}
+
+// CommandCreate is the ip access create command
```
makesitgo, line 18:
```suggestion
// CommandCreate is the 'ip-access create' command
```

internal/commands/ip_access/create_inputs.go
```diff
@@ -0,0 +1,22 @@
...
+
+type createInputs struct {
+	cli.ProjectInputs
+	IPAddress  string
+	Comment    string
+	UseCurrent bool
+	AllowAll   bool
+}
+
+func (i *createInputs) Resolve(profile *user.Profile, ui terminal.UI) error {
```
makesitgo, line 17: the separate files for these inputs seems a bit overkill, let's consolidate all things "create" related to here to the `create.go` file

internal/commands/ip_access/create_inputs.go
```diff
@@ -0,0 +1,22 @@
...
+	if err := i.ProjectInputs.Resolve(ui, profile.WorkingDirectory, false); err != nil {
+		return err
+	}
+	return nil
```
makesitgo, line 21: do you need to do anything if the `IPAddress` or `Comment` field is still left blank, prompt for any values and/or error here?

internal/commands/ip_access/create_inputs.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/cli/user"
+	"github.com/10gen/realm-cli/internal/terminal"
+)
+
+type createInputs struct {
+	cli.ProjectInputs
+	IPAddress  string
+	Comment    string
+	UseCurrent bool
```
makesitgo, line 13: what are `UseCurrent` or `AllowAll` used for?

internal/commands/ip_access/delete.go
```diff
@@ -0,0 +1,60 @@
...
+	Hidden:  true,
+}
+
+// CommandDelete for the ip access delete command
```
makesitgo, line 17:
```suggestion
// CommandDelete for the 'ip-access delete' command
```

internal/commands/ip_access/delete_inputs.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	"github.com/10gen/realm-cli/internal/terminal"
+)
+
+type deleteInputs struct {
+	cli.ProjectInputs
+	IPAddress string
+}
+
+func (i *deleteInputs) Resolve(profile *user.Profile, ui terminal.UI) error {
```
makesitgo, line 17: same thing about consolidating these inputs to the `delete.go` file

internal/commands/ip_access/delete_inputs.go
```diff
@@ -0,0 +1,35 @@
...
+	if len(i.IPAddress) > 0 {
+		for _, allowedIP := range allowedIPs {
+			if allowedIP.IPAddress == i.IPAddress {
+				return allowedIP, nil
+			}
+		}
+		return realm.AllowedIP{}, fmt.Errorf("unable to find allowed IP: %s", i.IPAddress)
+	}
+
+	return realm.AllowedIP{}, nil
```
makesitgo, line 34: if the `IPAddress` flag was never set, we'll return the empty `AllowedIP` here which is then passed to the `Delete` call (which will presumably fail).  we should either prompt the user for this information (you _do_ have a list of ip addresses they could select from), or take this moment to return an error early (rather than rely on the api call to fail)

internal/commands/ip_access/delete.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return err
+	}
+
+	allowedIPs, err := clients.Realm.AllowedIPs(app.GroupID, app.ID)
+	if err != nil {
+		return err
+	}
+
+	allowedIP, err := cmd.inputs.resolveAllowedIP(ui, allowedIPs)
```
makesitgo, line 45: the `user` and `secrets` delete commands allow the user to specify multiple items to delete. it seems like that might be worth pursuing here as well

internal/commands/ip_access/create.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return err
+	}
+
+	allowedIP, err := clients.Realm.CreateAllowedIP(app.GroupID, app.ID, cmd.inputs.IPAddress, cmd.inputs.Comment)
+	if err != nil {
+		return err
+	}
+
+	ui.Print(terminal.NewTextLog("Successfully created allowed IP, id: %s", allowedIP.ID))
```
makesitgo, line 50: ⚠️  reminder to self: follow up with drew/sumedha on format of output here (similar situation in `users create` & `secrets create` ... should probably make this consistent)

internal/commands/ip_access/delete.go
```diff
@@ -0,0 +1,60 @@
...
+
+	if err := clients.Realm.DeleteAllowedIP(
+		app.GroupID,
+		app.ID,
+		allowedIP.ID,
+	); err != nil {
+		return err
+	}
+
+	ui.Print(terminal.NewTextLog("Successfully deleted allowed IP"))
```
makesitgo, line 58: if you make the switch to "delete many" here, then printing a table (similar to `users` and `secrets`) would be good here

internal/commands/ip_access/inputs.go
```diff
@@ -0,0 +1,21 @@
...
+	flagNewIPUsageUpdate = "the new allowed IP address"
+
+	flagComment            = "comment"
+	flagCommentUsageCreate = "the comment of the allowd IP address"
+	flagCommentUsageUpdate = "the new comment of the allowed IP"
+
+	flagUseCurrent            = "use-current"
+	flagUseCurrentUsageCreate = "use current IP address"
+
+	flagAllowAll            = "allow-all"
```
makesitgo, line 19: one note on flags: the codebase still hasn't really landed on a solid pattern for these yet.  so first and foremost, what you have here is fine, don't sweat the organization (or re-use) of flags all that much.

I can see the point of some shared "names"/concepts for flags, but its rarely the case a usage string wants to be exactly the same across commands. just food for thought, in case you have any ideas here

internal/commands/ip_access/list.go
```diff
@@ -0,0 +1,82 @@
...
+
+// CommandMetaList is the command meta for the `ip-access list` command
+var CommandMetaList = cli.CommandMeta{
+	Use:     "list",
+	Aliases: []string{"ls"},
+	Display: "allowed IPs list",
+	Hidden:  true,
+}
+
+// CommandList is the ip access list command
```
makesitgo, line 21:
```suggestion
// CommandList is the 'ip-access list' command
```

internal/commands/ip_access/output.go
```diff
@@ -0,0 +1,10 @@
+package ipaccess
+
+const (
+	headerIP      = "IP Address"
+	headerComment = "Comment"
+)
+
+func tableHeaders(additionalHeaders ...string) []string {
```
makesitgo, line 8: if the additional headers aren't necessary yet, I'd just create:

```go
var (
  tableHeaders = []string{headerIP, headerComment}
)
```

internal/commands/ip_access/update.go
```diff
@@ -0,0 +1,64 @@
...
+	Hidden:  true,
+}
+
+// CommandUpdate is the ip access update command
```
makesitgo, line 17:
```suggestion
// CommandUpdate is the 'ip-access update' command
```

internal/commands/ip_access/list.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// Inputs are the command inputs
+func (cmd *CommandList) Inputs() cli.InputResolver {
+	return &cmd.inputs
+}
+
+// Handler is the command handler
+func (cmd *CommandList) Handler(profile *user.Profile, ui terminal.UI, clients cli.Clients) error {
+	app, appErr := cli.ResolveApp(ui, clients.Realm, cmd.inputs.Filter())
```
makesitgo, line 42: I will choose to not comment on the merits of using either `namedErr` or `err`, but I would say let's keep it consistent across the files worked on

internal/commands/ip_access/update_inputs.go
```diff
@@ -0,0 +1,37 @@
...
+	if len(i.IPAddress) > 0 {
+		for _, allowedIP := range allowedIPs {
+			if allowedIP.IPAddress == i.IPAddress {
+				return allowedIP, nil
+			}
+		}
+		return realm.AllowedIP{}, fmt.Errorf("unable to find allowed IP: %s", i.IPAddress)
+	}
+
+	return realm.AllowedIP{}, nil
```
makesitgo, line 36: if we make it here, we should either error or prompt the user for more info (rather than rely on a failed api call for an error)

internal/commands/ip_access/update_inputs.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+type updateInputs struct {
+	cli.ProjectInputs
+	IPAddress    string
+	NewIPAddress string
+	Comment      string
+}
+
+func (i *updateInputs) Resolve(profile *user.Profile, ui terminal.UI) error {
```
makesitgo, line 19: same thing about consolidating these inputs to the `update.go` file

internal/commands/ip_access/update_inputs.go
```diff
@@ -0,0 +1,37 @@
...
+	if err := i.ProjectInputs.Resolve(ui, profile.WorkingDirectory, false); err != nil {
+		return err
+	}
+	return nil
```
makesitgo, line 23: if both `NewIPAddress` and `Comment` are blank, I'm guessing the update fails (or at least is a no-op).  do you want to catch that case here, potentially prompt for any necessary info or just return an error?

makesitgo (changes requested):
a lot of repeated comments (sorry for the heft here!)

overall this looks great... just various cleanup/organizational things and only a few things left to do

internal/commands/ip_access/create.go
asudhakal: This is what was decided on the spec (for now). @christineschen tagging you for visibility

internal/commands/ip_access/create_inputs.go
asudhakal: `UseCurrent` uses the users current IP address during creation. `AllowAll` allows access from anywhere (i.e `0.0.0.0/0`)

asudhakal: Just got this resolved, had to change/update the SDK as well. Thanks for pointing it out!

internal/commands/ip_access/create.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/spf13/pflag"
+)
+
+// CommandMetaCreate is the command meta for the `ip-access create` command
+var CommandMetaCreate = cli.CommandMeta{
+	Use:     "create",
+	Aliases: []string{"add"},
+	Display: "allowed IP create",
+	Hidden:  true,
```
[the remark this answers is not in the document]
asudhakal: will create a ticket and a `TODO`

internal/commands/ip_access/delete.go
asudhakal: will create a ticket and a `TODO`

internal/commands/ip_access/list.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/spf13/pflag"
+)
+
+// CommandMetaList is the command meta for the `ip-access list` command
+var CommandMetaList = cli.CommandMeta{
+	Use:     "list",
+	Aliases: []string{"ls"},
+	Display: "allowed IPs list",
```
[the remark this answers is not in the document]
asudhakal: will create a ticket and a `TODO`

internal/commands/ip_access/update.go
asudhakal: will create a ticket and a `TODO`.

internal/commands/accesslist/create.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (i *createInputs) Resolve(profile *user.Profile, ui terminal.UI) error {
+	if err := i.ProjectInputs.Resolve(ui, profile.WorkingDirectory, false); err != nil {
+		return err
+	}
+
+	if i.IPAddress == "" {
+		if (!i.UseCurrent && !i.AllowAll) || (i.UseCurrent && i.AllowAll) {
+			return errors.New("Must provide an IP Address or one of use-current and allow-all.")
```
asudhakal, line 75: Will work with design to figure out appropriate error messages.

internal/commands/accesslist/create.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+	} else {
+		if i.UseCurrent || i.AllowAll {
+			return errors.New("Cannot provide both an IP Address and one of use-current or allow-all.")
```
asudhakal, line 79: Will work with design to figure out appropriate error messages.

makesitgo:
I would strongly recommend breaking this PR up into smaller tickets... one for each command would be great (with `cloud/realm` client changes included as need be).  the first one can still include the `Hidden` stuff as well, but this will help the conversations be much more focused/detailed

PR #159, opened 2021-06-02
REALMC-9100: Allow CLI to create from a remote app a new app in a different project

jandersongo:
…ferent project

### DIFF
--- a/internal/cli/inputs.go
+++ b/internal/cli/inputs.go
@@ -123,7 +123,7 @@ func ResolveApp(ui terminal.UI, client realm.Client, filter realm.AppFilter) (re
 }
 
 // ResolveGroupID will use the provided MongoDB Cloud Atlas client to resolve the user's group id
-func ResolveGroupID(ui terminal.UI, client atlas.Client) (string, error) {
+func ResolveGroupID(ui terminal.UI, client atlas.Client, defaultGroupID string) (string, error) {
 	groups, groupsErr := client.Groups()
 	if groupsErr != nil {
 		return "", groupsErr
@@ -136,6 +136,7 @@ func ResolveGroupID(ui terminal.UI, client atlas.Client) (string, error) {
 		return groups[0].ID, nil
 	}
 
+	var defaultSelection string
 	groupIDsByOption := make(map[string]string, len(groups))
 	groupIDOptions := make([]string, len(groups))
 	for i, group := range groups {
@@ -143,12 +144,21 @@ func ResolveGroupID(ui terminal.UI, client atlas.Client) (string, error) {
 
 		groupIDsByOption[option] = group.ID
 		groupIDOptions[i] = option
+
+		if group.ID == defaultGroupID {
+			defaultSelection = option
+		}
+	}
+
+	if ui.AutoConfirm() && defaultGroupID != "" {
+		return groupIDsByOption[defaultSelection], nil
 	}
 
 	var selection string
 	if err := ui.AskOne(&selection, &survey.Select{
 		Message: "Atlas Project",
 		Options: groupIDOptions,
+		Default: defaultSelection,
 	}); err != nil {
 		return "", fmt.Errorf("failed to select group id: %s", err)
 	}
--- a/internal/cli/inputs.go
+++ b/internal/cli/inputs.go
@@ -123,7 +123,7 @@ func ResolveApp(ui terminal.UI, client realm.Client, filter realm.AppFilter) (re
 }
 
 // ResolveGroupID will use the provided MongoDB Cloud Atlas client to resolve the user's group id
-func ResolveGroupID(ui terminal.UI, client atlas.Client, defaultGroupID string) (string, error) {
+func ResolveGroupID(ui terminal.UI, client atlas.Client) (string, error) {
 	groups, groupsErr := client.Groups()
 	if groupsErr != nil {
 		return "", groupsErr
@@ -136,7 +136,6 @@ func ResolveGroupID(ui terminal.UI, client atlas.Client, defaultGroupID string) 
 		return groups[0].ID, nil
 	}
 
-	var defaultSelection string
 	groupIDsByOption := make(map[string]string, len(groups))
 	groupIDOptions := make([]string, len(groups))
 	for i, group := range groups {
@@ -144,21 +143,12 @@ func ResolveGroupID(ui terminal.UI, client atlas.Client, defaultGroupID string) 
 
 		groupIDsByOption[option] = group.ID
 		groupIDOptions[i] = option
-
-		if group.ID == defaultGroupID {
-			defaultSelection = option
-		}
-	}
-
-	if defaultGroupID != "" {
-		return groupIDsByOption[defaultSelection], nil
 	}
 
 	var selection string
 	if err := ui.AskOne(&selection, &survey.Select{
 		Message: "Atlas Project",
 		Options: groupIDOptions,
-		Default: defaultSelection,
 	}); err != nil {
 		return "", fmt.Errorf("failed to select group id: %s", err)
 	}
--- a/internal/cli/inputs.go
+++ b/internal/cli/inputs.go
@@ -150,7 +150,7 @@ func ResolveGroupID(ui terminal.UI, client atlas.Client, defaultGroupID string) 
 		}
 	}
 
-	if ui.AutoConfirm() && defaultGroupID != "" {
+	if defaultGroupID != "" {
 		return groupIDsByOption[defaultSelection], nil
 	}
 
--- a/internal/cli/inputs_test.go
+++ b/internal/cli/inputs_test.go
@@ -1,6 +1,7 @@
 package cli_test
 
 import (
+	"bytes"
 	"errors"
 	"os"
 	"path/filepath"
@@ -191,10 +192,17 @@ func TestResolveGroupID(t *testing.T) {
 		Name: "eggcorn",
 	}
 
+	testGroup2 := atlas.Group{
+		ID:   "some-other-id",
+		Name: "eggcorn2",
+	}
+
 	for _, tc := range []struct {
 		description     string
 		groups          []atlas.Group
 		procedure       func(c *expect.Console)
+		defaultGroupID  string
+		autoConfirm     bool
 		expectedGroupID string
 		expectedErr     error
 	}{
@@ -218,6 +226,14 @@ func TestResolveGroupID(t *testing.T) {
 			},
 			expectedGroupID: testGroup.ID,
 		},
+		{
+			description:     "Should select default group when FlagAutoConfirm is set",
+			groups:          []atlas.Group{testGroup, testGroup2},
+			procedure:       func(c *expect.Console) {},
+			autoConfirm:     true,
+			defaultGroupID:  testGroup2.ID,
+			expectedGroupID: testGroup2.ID,
+		},
 	} {
 		t.Run(tc.description, func(t *testing.T) {
 			atlasClient := mock.AtlasClient{}
@@ -225,7 +241,9 @@ func TestResolveGroupID(t *testing.T) {
 				return tc.groups, nil
 			}
 
-			_, console, _, ui, consoleErr := mock.NewVT10XConsole()
+			console, _, ui, consoleErr := mock.NewVT10XConsoleWithOptions(
+				mock.UIOptions{AutoConfirm: tc.autoConfirm},
+				new(bytes.Buffer))
 			assert.Nil(t, consoleErr)
 			defer console.Close()
 
@@ -234,8 +252,7 @@ func TestResolveGroupID(t *testing.T) {
 				defer close(doneCh)
 				tc.procedure(console)
 			}()
-
-			groupID, err := cli.ResolveGroupID(ui, atlasClient)
+			groupID, err := cli.ResolveGroupID(ui, atlasClient, tc.defaultGroupID)
 
 			console.Tty().Close() // flush the writers
 			<-doneCh              // wait for procedure to complete
@@ -251,7 +268,7 @@ func TestResolveGroupID(t *testing.T) {
 			return nil, errors.New("something bad happened")
 		}
 
-		_, err := cli.ResolveGroupID(nil, atlasClient)
+		_, err := cli.ResolveGroupID(nil, atlasClient, "")
 		assert.Equal(t, errors.New("something bad happened"), err)
 	})
 }
--- a/internal/cli/inputs_test.go
+++ b/internal/cli/inputs_test.go
@@ -1,7 +1,6 @@
 package cli_test
 
 import (
-	"bytes"
 	"errors"
 	"os"
 	"path/filepath"
@@ -192,17 +191,10 @@ func TestResolveGroupID(t *testing.T) {
 		Name: "eggcorn",
 	}
 
-	testGroup2 := atlas.Group{
-		ID:   "some-other-id",
-		Name: "eggcorn2",
-	}
-
 	for _, tc := range []struct {
 		description     string
 		groups          []atlas.Group
 		procedure       func(c *expect.Console)
-		defaultGroupID  string
-		autoConfirm     bool
 		expectedGroupID string
 		expectedErr     error
 	}{
@@ -226,22 +218,6 @@ func TestResolveGroupID(t *testing.T) {
 			},
 			expectedGroupID: testGroup.ID,
 		},
-		{
-			description:     "should select default group when provided and ui is set to auto confirm",
-			groups:          []atlas.Group{testGroup, testGroup2},
-			procedure:       func(c *expect.Console) {},
-			autoConfirm:     true,
-			defaultGroupID:  testGroup2.ID,
-			expectedGroupID: testGroup2.ID,
-		},
-		{
-			description:     "should select default group when provided and ui is not set to auto confirm",
-			groups:          []atlas.Group{testGroup, testGroup2},
-			procedure:       func(c *expect.Console) {},
-			autoConfirm:     false,
-			defaultGroupID:  testGroup2.ID,
-			expectedGroupID: testGroup2.ID,
-		},
 	} {
 		t.Run(tc.description, func(t *testing.T) {
 			atlasClient := mock.AtlasClient{}
@@ -249,10 +225,7 @@ func TestResolveGroupID(t *testing.T) {
 				return tc.groups, nil
 			}
 
-			console, _, ui, consoleErr := mock.NewVT10XConsoleWithOptions(
-				mock.UIOptions{AutoConfirm: tc.autoConfirm},
-				new(bytes.Buffer),
-			)
+			_, console, _, ui, consoleErr := mock.NewVT10XConsole()
 			assert.Nil(t, consoleErr)
 			defer console.Close()
 
@@ -261,7 +234,7 @@ func TestResolveGroupID(t *testing.T) {
 				defer close(doneCh)
 				tc.procedure(console)
 			}()
-			groupID, err := cli.ResolveGroupID(ui, atlasClient, tc.defaultGroupID)
+			groupID, err := cli.ResolveGroupID(ui, atlasClient)
 
 			console.Tty().Close() // flush the writers
 			<-doneCh              // wait for procedure to complete
@@ -277,7 +250,7 @@ func TestResolveGroupID(t *testing.T) {
 			return nil, errors.New("something bad happened")
 		}
 
-		_, err := cli.ResolveGroupID(nil, atlasClient, "")
+		_, err := cli.ResolveGroupID(nil, atlasClient)
 		assert.Equal(t, errors.New("something bad happened"), err)
 	})
 }
--- a/internal/cli/inputs_test.go
+++ b/internal/cli/inputs_test.go
@@ -33,7 +33,7 @@ func TestProjectAppInputsResolve(t *testing.T) {
 		test        func(t *testing.T, i cli.ProjectInputs)
 	}{
 		{
-			description: "Should not prompt for app when set by flag already",
+			description: "should not prompt for app when set by flag already",
 			inputs:      cli.ProjectInputs{App: "some-app"},
 			wd:          testRoot,
 			procedure:   func(c *expect.Console) {},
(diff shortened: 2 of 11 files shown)


internal/commands/app/create_inputs.go
```diff
@@ -82,9 +82,9 @@ func (i *createInputs) Resolve(profile *user.Profile, ui terminal.UI) error {
 	return nil
 }
 
-func (i *createInputs) resolveName(ui terminal.UI, client realm.Client, r appRemote) error {
+func (i *createInputs) resolveName(ui terminal.UI, client realm.Client, filter realm.AppFilter) error {
```
jandersongo, line 85: Changed this because the FindApps methods was looking for **client app id** while we were trying to find a match for an **app id**

internal/commands/app/create_inputs.go
```diff
@@ -108,13 +108,16 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
 		return fullPath, nil
 	}
-	_, appOK, err := local.FindApp(fullPath)
+	_, _, err = local.FindApp(fullPath)
 	if err != nil {
 		return "", err
 	}
-	if appOK {
-		return "", errProjectExists{fullPath}
+
```
jandersongo, line 115: instead of returning error due to existing path, prompt for alternative path

internal/commands/app/create.go
```diff
@@ -65,19 +65,18 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
+		if err != nil {
+			return err
 		}
-	} else {
-		groupID = appRemote.GroupID
+		groupID = id
 	}
 
-	err = cmd.inputs.resolveName(ui, clients.Realm, appRemote)
+	err = cmd.inputs.resolveName(ui, clients.Realm, realm.AppFilter{
```
jandersongo, line 77: Changed this because the resolveName->ResolveApp->FindApps methods was looking for **client app id** while we were trying to find a match for an **app id** through appRemote

internal/cli/inputs.go
```diff
@@ -136,19 +136,29 @@ func ResolveGroupID(ui terminal.UI, client atlas.Client) (string, error) {
...
 
 		groupIDsByOption[option] = group.ID
 		groupIDOptions[i] = option
+
+		if group.ID == defaultGroupID {
+			defaultSelection = option
+		}
+	}
+
+	if ui.AutoConfirm() && defaultGroupID != "" {
```
makesitgo, line 153: [nit] this may read easier if the comparison is `defaultSelection != ""`

internal/cli/inputs_test.go
```diff
@@ -218,14 +226,24 @@ func TestResolveGroupID(t *testing.T) {
...
 		t.Run(tc.description, func(t *testing.T) {
 			atlasClient := mock.AtlasClient{}
 			atlasClient.GroupsFn = func() ([]atlas.Group, error) {
 				return tc.groups, nil
 			}
 
-			_, console, _, ui, consoleErr := mock.NewVT10XConsole()
+			console, _, ui, consoleErr := mock.NewVT10XConsoleWithOptions(
+				mock.UIOptions{AutoConfirm: tc.autoConfirm},
+				new(bytes.Buffer))
```
makesitgo, line 246: [style] to keep things balanced, you can end this line with `,` and have the last trailing `)` on its own newline

internal/cli/inputs_test.go
```diff
@@ -218,14 +226,24 @@ func TestResolveGroupID(t *testing.T) {
 			},
 			expectedGroupID: testGroup.ID,
 		},
+		{
+			description:     "Should select default group when FlagAutoConfirm is set",
```
makesitgo, line 230: [style] for test descriptions, try to keep things all lower-cased (and without punctuation)

```go
description:     "should select default group when provided and ui is set to auto confirm",
```

internal/cli/inputs_test.go
```diff
@@ -218,14 +226,24 @@ func TestResolveGroupID(t *testing.T) {
 			},
 			expectedGroupID: testGroup.ID,
 		},
+		{
```
makesitgo, line 229: can we also write a test that proves the default selection is present when auto confirm is not true?

internal/commands/app/create.go
```diff
@@ -65,19 +65,18 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 	}
 
 	var groupID = cmd.inputs.Project
```
makesitgo, line 67: [drive-by] not sure why we used this syntax, but could you change this to `groupID := cmd.inputs.Project`?

---

alternatively, we could just mutate `cmd.inputs.Project` itself here and not have a `groupID` variable at all.  if you go that route, then i'd also make the (aesthetic) change to L#70 below and rename `id` to `groupID`

internal/commands/app/create.go
```diff
@@ -65,19 +65,18 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
-	if appRemote.IsZero() {
-		if groupID == "" {
-			id, err := cli.ResolveGroupID(ui, clients.Atlas)
-			if err != nil {
-				return err
-			}
-			groupID = id
+
```
makesitgo, line 68: [nit] omit this newline

internal/commands/app/create_inputs.go
makesitgo, line 85: lets change this function signature to just take `groupID, appID string` arguments, and you can construct the `realm.AppFilter` inside this function

internal/commands/app/create.go
makesitgo: in the `resolveRemoteApp` implementation, would it help us/make sense to pull `app.ClientAppID` instead of `app.ID`?  honestly the `remoteApp` struct could maybe just go away entirely and we deal with the `realm.App` itself (so we have all that data).  what do you think?

internal/commands/app/create.go
```diff
@@ -65,19 +65,18 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
 		}
-	} else {
-		groupID = appRemote.GroupID
+		groupID = id
 	}
 
-	err = cmd.inputs.resolveName(ui, clients.Realm, appRemote)
+	err = cmd.inputs.resolveName(ui, clients.Realm, realm.AppFilter{
+		GroupID: appRemote.GroupID,
+		App:     cmd.inputs.RemoteApp})
```
makesitgo, line 79: i think we still want to leverage the resolved remote app somehow here.  probably related to my question above about potentially eliminating `remoteApp` as a return value, but just wanted to leave a comment on this line as well

---

[also style nit] keep this line ending with a `,` and move the `)` to its own newline

internal/commands/app/create_inputs.go
makesitgo: I think we might still want this check?  What this prevents is a user from creating an app directory inside another app directory.  I think thats something we still can enforce for now, but i might also be missing something here.  let me know what you think

internal/commands/app/create_inputs.go
```diff
@@ -108,13 +108,16 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
+	if ui.AutoConfirm() {
+		fullPath = path.Join(wd, i.LocalPath+"-1")
```
makesitgo, line 117: it would be nice if we could somehow verify the `-1` doesn't already exist too, and potentially keep incrementing up until we find a directory that _is_ available on disk.

also, this unique path could then be set to `Default` on the input prompt below

internal/commands/app/create_inputs_test.go
```diff
@@ -111,7 +112,8 @@ func TestAppCreateInputsResolveName(t *testing.T) {
 				return []realm.App{testApp}, nil
 			}
 
-			err := tc.inputs.resolveName(nil, rc, tc.appRemote)
+			err := tc.inputs.resolveName(nil, rc, realm.AppFilter{
+				GroupID: tc.appRemote.GroupID, App: tc.appRemote.AppID})
```
makesitgo, line 116: same style nit throughout this file, its fairly common in Go to have that trailing `)` on its own newline (if/when you put each arg/field on its own newline).  i think here you can go one of two routes:

```go
err := tc.inputs.resolveName(nil, rc, realm.AppFilter{tc.appRemote.GroupID, tc.appRemote.AppID})
```

or

```go
err := tc.inputs.resolveName(nil, rc, realm.AppFilter{
  GroupID: tc.appRemote.GroupID,
  App: tc.appRemote.AppID,
})
```

that being said, depending on how we refactor the function signature for `resolveName` and/or what we do with the return value from `resolveRemoteApp`, maybe we can just edit the `tc` struct here (`appRemote` could be split into `groupID string` and `appID string` and/or changed to `realm.App`

internal/commands/app/create_inputs_test.go
```diff
@@ -127,7 +129,7 @@ func TestAppCreateInputsResolveName(t *testing.T) {
 			return nil, errors.New("realm client error")
 		}
 		inputs := createInputs{}
-		err := inputs.resolveName(nil, rc, appRemote{testApp.GroupID, testApp.ID})
+		err := inputs.resolveName(nil, rc, realm.AppFilter{GroupID: testApp.GroupID, App: testApp.ID})
```
makesitgo, line 132: I think its best if we keep `AppFilter` to be using either `Name` or `ClientAppID` (as you saw, `FindApps` looks for matches on those rather than the actual ID values).  i know this is a test and is mocked, so things are likely working fine here, but best to try and keep it used in a way that helps document things

internal/commands/app/create_inputs_test.go
```diff
@@ -209,14 +211,53 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
 		assert.Equal(t, "new-app", inputs.LocalPath)
 	})
 
-	t.Run("should error when path specified is another realm app", func(t *testing.T) {
+	t.Run("should create default local directory name when FlagAutoConfirm is set", func(t *testing.T) {
```
makesitgo, line 214: same thing here about capital letters in the test description
```suggestion
	t.Run("should create default local directory name when ui is set to auto confirm", func(t *testing.T) {
```

internal/commands/app/create_inputs_test.go
```diff
@@ -209,14 +211,53 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
 		profile, teardown := mock.NewProfileFromTmpDir(t, "app_create_test")
 		defer teardown()
 
-		specifiedDir := "test-dir"
-		inputs := createInputs{LocalPath: specifiedDir}
+		console, _, ui, err := mock.NewVT10XConsoleWithOptions(
```
makesitgo, line 218: do we need the `V10XConsole` here if we're not testing anything interactively here?

internal/commands/app/create_test.go
```diff
@@ -393,6 +393,76 @@ func TestAppCreateHandler(t *testing.T) {
 		}, "\n"), out.String())
 	})
 
+	t.Run("should create a new app with a structure based on the specified remote app and prompt for project", func(t *testing.T) {
+		profile, teardown := mock.NewProfileFromTmpDir(t, "app_create_test")
+		defer teardown()
+		profile.SetRealmBaseURL("http://localhost:8080")
```
makesitgo, line 399: was this needed?

internal/commands/app/create_test.go
```diff
@@ -393,6 +393,76 @@ func TestAppCreateHandler(t *testing.T) {
...
+		doneCh := make(chan (struct{}))
+		go func() {
+			defer close(doneCh)
+			console.ExpectString("Atlas Project")
+			console.Send(newGroupID)
+			console.SendLine(" ")
+			console.ExpectEOF()
+		}()
+
+		zipPkg, err := zip.OpenReader("testdata/project.zip")
```
makesitgo, line 422: do we need a `defer` close on this?

makesitgo (commented):
just some initial thoughts and questions here, it was working great locally though, nice work here

internal/commands/app/create.go
jandersongo: I also thought about adding ClientAppID to the remoteApp struct but I prefer the idea of returning the realm.App from resolveRemoteApp instead as you suggested.

internal/commands/app/create_test.go
jandersongo: on the zipPkg? I tried to follow the same usage [here](https://github.com/10gen/realm-cli/blob/7437d6edb804f7700b4cd92ffbf9dfb3cddd0bcf/internal/commands/app/create_test.go#L326) which did not close. Should I add something like [this](https://github.com/10gen/realm-cli/blob/7437d6edb804f7700b4cd92ffbf9dfb3cddd0bcf/internal/commands/pull/command_test.go#L191-L193) to both instances?

internal/commands/app/create_inputs.go
jandersongo: is that specific case being check [here](https://github.com/10gen/realm-cli/blob/7437d6edb804f7700b4cd92ffbf9dfb3cddd0bcf/internal/commands/app/init_inputs.go#L21-L23)?

internal/commands/app/create_test.go
makesitgo: yeah, seems like something I must've missed the first time around.  The return value of `zip.OpenReader` is a `zip.ReadCloser`, which has:

```go
// A ReadCloser is a Reader that must be closed when no longer needed.
type ReadCloser struct {
	f *os.File
	Reader
}
```

We probably want to add a `defer zipPkg.Close()` after the `err` check in both of those spots

internal/commands/app/create_inputs.go
makesitgo: I think that link you provided is doing the same check, yes.  Those are for the `app init` inputs though, so something similar would still need to live here as well (unless I'm missing something here)

And while I actually really like the idea of having that check again, I think the "true fix" there would be to say: "let's support auto-complete in a way that the user's input can be tabbed here and the list of matching directories (if any) would show"

Something we can definitely punt on for now.  A lot of words for basically saying "great idea, yes we'd want that, but we'll tackle it in a different ticket" :)

jandersongo: Could you please explain the auto-complete idea? 

I think the errProjectExists is a good catch here in case the user, when prompted, inputs a local path that already exists.

makesitgo: Yeah so definitely don't worry about the auto-complete here, but the idea is similar to when you're working in a terminal, and say you try to `cd` into a directory.  As you type out a name, you can hit "tab" and the shell generally will either auto-fill the name to what it knows is present, and/or show you a list of matches afterwards.

That way, the user is given more immediate feedback about the directories that are present on disk at the time, rather than typing in a name, hitting enter, and letting another cycle of the same validations run.

Again, I think what we have here with the first initial check is fine.  No need to over-complicate it with recursive validations or auto-complete just yet.

internal/cli/inputs_test.go
```diff
@@ -191,41 +192,67 @@ func TestResolveGroupID(t *testing.T) {
...
+			procedure:       func(c *expect.Console) {},
+			autoConfirm:     true,
+			defaultGroupID:  testGroup2.ID,
+			expectedGroupID: testGroup2.ID,
+		},
+		{
+			description:     "should select default group when provided and ui is not set to auto confirm",
+			groups:          []atlas.Group{testGroup, testGroup2},
+			procedure:       func(c *expect.Console) {},
+			autoConfirm:     false,
```
makesitgo, line 241: can omit this line

internal/cloud/realm/app.go
```diff
@@ -41,6 +41,11 @@ func (app App) Option() string {
 	return fmt.Sprintf("%s (%s)", app.ClientAppID, app.GroupID)
 }
 
+//IsZero checks for app GroupID and ID
```
makesitgo, line 44: [nit] for the docstrings, make sure to place a space between `// ` and the start of the message

---

also, how many places is this actually used?  if its just a few times, you might be able to fix my bad code and just eliminate this method all together

internal/commands/app/create.go
```diff
@@ -64,20 +64,16 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
-	if appRemote.IsZero() {
-		if groupID == "" {
-			id, err := cli.ResolveGroupID(ui, clients.Atlas)
-			if err != nil {
-				return err
-			}
-			groupID = id
+	groupID := cmd.inputs.Project
+	if groupID == "" {
+		id, err := cli.ResolveGroupID(ui, clients.Atlas, appRemote.GroupID)
```
makesitgo, line 69: instead of adding the `defaultGroupID` as an argument to `ResolveGroupID`, thoughts on using the following approach:

```go
groupID := cmd.inputs.Project
if groupID == "" {
  groupID = appRemote.GroupID
}
if groupID == "" {
  groupID, err = cli.ResolveGroupID(ui, clients.Atlas)
  if err != nil { return err }
}
```

I _think_ that removes the need to modify `ResolveGroupID`, thoughts?

internal/commands/app/create_inputs.go
```diff
@@ -108,23 +112,41 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
 	}
 	if appOK {
-		return "", errProjectExists{fullPath}
+		return "", errProjectExists{wd}
+	}
+
+	defaultLocalPath := getDefaultPath(wd, i.LocalPath)
+	if ui.AutoConfirm() {
+		fullPath = path.Join(wd, defaultLocalPath)
+		return fullPath, nil
```
makesitgo, line 128: [nit] can just `return path.Join(wd, defaultLocalPath), nil` here

internal/commands/app/create_inputs.go
```diff
@@ -217,3 +239,15 @@ func (i createInputs) args(omitDryRun bool) []flags.Arg {
 	}
 	return args
 }
+
+func getDefaultPath(wd string, localPath string) string {
+	i := 1
+	for {
```
makesitgo, line 245: a couple of thoughts here:
  - should assign `localPath + "-" + strconv.Itoa(i)` to a var within this loop
  - the error check seems off here, doesn't it want to be `if err == nil && !found`?
  - it would be nice if we had a failsafe here (maybe append a `primitive.NewObjectID().Hex()` if we exhaust file options `1-9`?)
  - we should add some unit tests for this as well

internal/commands/app/create_inputs_test.go
```diff
@@ -209,14 +210,52 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
 	})
 
-	t.Run("should error when path specified is another realm app", func(t *testing.T) {
+	t.Run("should create default local directory name when ui is set to auto confirm", func(t *testing.T) {
 		profile, teardown := mock.NewProfileFromTmpDir(t, "app_create_test")
 		defer teardown()
 
-		specifiedDir := "test-dir"
-		inputs := createInputs{LocalPath: specifiedDir}
+		_, _, ui, err := mock.NewVT10XConsoleWithOptions(
```
makesitgo, line 217: [question] would `mock.NewConsoleWithOptions` suffice here?  doesn't look we test anything interactive here

internal/commands/app/create_inputs_test.go
```diff
@@ -209,14 +210,52 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
+		newDefaultName := "test-app-1"
+		inputs := createInputs{newAppInputs: newAppInputs{Name: testAppName}}
+
+		err = os.Mkdir(path.Join(profile.WorkingDirectory, testAppName), os.ModePerm)
+		assert.Nil(t, err)
+
+		dir, err := inputs.resolveLocalPath(ui, profile.WorkingDirectory)
+		assert.Nil(t, err)
+		assert.Equal(t, path.Join(profile.WorkingDirectory, newDefaultName), dir)
+		assert.Equal(t, testAppName, inputs.LocalPath)
```
makesitgo, line 232: I guess I would have expected `LocalPath` to reflect the `dir` we returned from `resolveLocalPath`, so my question is do we need this assert?  If we don't need to mutate `LocalPath` for this command to work then we probably don't need to test its behavior for now

internal/commands/app/create_inputs_test.go
```diff
@@ -209,14 +210,52 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
+	})
+
+	t.Run("should request new path when path specified is another realm app", func(t *testing.T) {
```
makesitgo, line 235: can we add a test that it fails when the new path specified is also a realm app?  and is it feasible to test the case when `wd` is a realm app itself?  (basically, the two `errProjectExists` cases)

makesitgo (commented):
I think just the last few things from me on this, should be ready to go after this.  really nice work on this stuff!

internal/commands/app/create.go
jandersongo: The default group was added to account for the auto confirm flag. 
We pass in the group that the remote app belonged to to ResolveGroupID so it could be selected in case the auto confirm flag is set.

internal/commands/app/create_inputs_test.go
```diff
@@ -209,14 +215,138 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
+			defer close(doneCh)
+			console.ExpectString("Local path './existing-app' already exists, writing app contents to that destination may result in file conflicts.")
+			console.ExpectString("Would you still like to write app contents to './existing-app'? ('No' will prompt you to provide another destination)")
+			console.SendLine("no")
+			console.ExpectString("Local Path")
+			console.SendLine("existing-app")
+			console.ExpectEOF()
+		}()
+
+		dir, err := inputs.resolveLocalPath(ui, profile.WorkingDirectory)
```
makesitgo, line 294: not a big deal and no need to make changes, but generally (here and the test above) if you're testing the `err` case, we usually don't care about the other return values (e.g. `_, err := inputs.resolveLocalPath(...)`

internal/commands/app/create_inputs_test.go
```diff
@@ -346,3 +475,29 @@ func TestAppCreateInputsResolveDataLake(t *testing.T) {
 		assert.Equal(t, "123", expectedGroupID)
 	})
 }
+
+func TestFindDefaultPath(t *testing.T) {
+	t.Run("should return new incremented directory if provided directory already exists", func(t *testing.T) {
```
makesitgo, line 480: I think its worth adding test cases here that we return `name-1` as a default path, and something like `name-9`.  we could probably convert this test code to a test table to reuse the test implementation

makesitgo (approved):
I think there's just a missing test or two here, but once those are in and the commits are in good order, this should be good to go!

🆒  beans

PR #160, opened 2021-06-07
REALMC-8823: Use Realm log type filters when listing logs

jandersongo:


internal/cloud/realm/logs.go
```diff
@@ -95,7 +95,11 @@ type logsResponse struct {
 func (c *client) Logs(groupID, appID string, opts LogsOptions) (Logs, error) {
 	query := map[string]string{}
 	if len(opts.Types) > 0 {
-		query[logsQueryType] = strings.Join(opts.Types, ",")
+		formattedTypes := make([]string, len(opts.Types))
```
makesitgo, line 98: I think there may be a different fix for this, now that I look more closely here...

in [`logs/list.go`](https://github.com/10gen/realm-cli/blob/master/internal/commands/logs/list.go#L67), we should probably be using `cmd.inputs.logTypes()` as the value for `opts.Types` (rather than just `cmd.inputs.Types` directly)

then I think this method remains unchanged, it would be nice to add a test somewhere in the `logs` package that made sure we passed the expected, formatted strings as an arg to the (mocked) client method

makesitgo (changes requested):
we might be able to tackle this more directly in the command handler file, let me know what you think here

makesitgo:
Also, just a note on the commit message, something a bit more descriptive here would be useful: `REALMC-8823: Use Realm formatted log types based on user flags`

internal/commands/logs/list_inputs_test.go
```diff
@@ -27,5 +29,24 @@ func TestLogTypes(t *testing.T) {
...
 			assert.Equal(t, tc.logTypes, i.logTypes())
 		})
+
+		t.Run("should fail if log type "+tc.logType+" is not formatted before passed to client", func(t *testing.T) {
+			realmClient := mock.RealmClient{}
+			realmClient.FindAppsFn = func(filter realm.AppFilter) ([]realm.App, error) {
+				return []realm.App{{}}, nil
+			}
+			realmClient.LogsFn = func(groupID, appID string, opts realm.LogsOptions) (realm.Logs, error) {
+				assert.Equal(t, tc.logTypes, opts.Types)
```
makesitgo, line 39: it's generally a good idea to keep asserts out of mock function implementations like this, its a common practice to have a `var logsOpts realm.LogsOptions` outside of the scope of the `LogsFn`, and then assign to it inside the function implementation, so that it can be asserted on downstream in the test

internal/commands/logs/list_inputs_test.go
```diff
@@ -27,5 +29,24 @@ func TestLogTypes(t *testing.T) {
 
 			assert.Equal(t, tc.logTypes, i.logTypes())
 		})
+
+		t.Run("should fail if log type "+tc.logType+" is not formatted before passed to client", func(t *testing.T) {
```
makesitgo, line 33: I think this test would be better suited for `list_test.go`, since its running the `cmd.Handler` function (it can be based on just one of these test cases)

---

also [nit] while string concatenation like this is fairly performant in go, considering we're dealing with test descriptions, i'd just use `fmt.Sprintf(...)` here as it reads a bit easier

makesitgo (commented):
just a test re-organization thought here, then should be good to go

internal/commands/logs/list_test.go
```diff
@@ -386,3 +386,24 @@ func TestLogTypeDisplay(t *testing.T) {
 	}
 	t.Logf("the max width for log type display is: %d", maxWidth)
 }
+
+func TestLogTypeFormat(t *testing.T) {
+	t.Run(fmt.Sprintf("should fail if log type %s is not formatted before passed to client", logTypeSchema), func(t *testing.T) {
```
makesitgo, line 391: considering we aren't actually asserting any failures here, maybe we re-word this test description?

`with schema log type set should query for all types of schema logs`  (<-- I don't love this either, but you get the idea)

internal/commands/logs/list_test.go
```diff
@@ -386,3 +386,24 @@ func TestLogTypeDisplay(t *testing.T) {
...
+			logsOpts = opts
+			return realm.Logs{}, nil
+		}
+
+		typeInputs := listInputs{Types: []string{logTypeSchema}}
+
+		cmd := &CommandList{typeInputs}
+		assert.Nil(t, cmd.Handler(nil, nil, cli.Clients{Realm: realmClient}))
+
+		assert.Equal(t, typeInputs.logTypes(), logsOpts.Types)
```
makesitgo, line 407: I think having `[]string{realm.LogTypeSchema..., ...}` would be fine/more readable here than relying on constructing new inputs and calling `logTypes()` on it

makesitgo (approved):
last few things but this is good to go afterwards.  just make sure to clean up the commits, and perhaps edit the main commit message/pr title here:

`REALMC-8823: Use Realm log type filters when listing logs` (or something along those lines -- just try to avoid vague messages such as "Fix bug" and the like)

PR #166, opened 2021-06-10
REALMC-8830: Enable linking to multiple clusters/datalakes in one command

jandersongo:


internal/commands/app/create_inputs.go
```diff
@@ -153,56 +154,69 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
-		}
+
+	existingClusters := map[string]atlas.Cluster{}
+	for _, c := range clusters {
+		existingClusters[c.Name] = c
 	}
-	if clusterName == "" {
-		return dataSourceCluster{}, errors.New("failed to find Atlas cluster")
+
+	var dsClusters []dataSourceCluster
```
makesitgo, line 168: initialize this with 0 length and capacity to fit `len(i.Clusters)`

internal/commands/app/create_inputs.go
```diff
@@ -153,56 +154,69 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
+	for _, clusterName := range i.Clusters {
+		if _, ok := existingClusters[clusterName]; !ok {
+			continue
+		}
+		dsClusters = append(dsClusters,
+			dataSourceCluster{
+				Name: "mongodb-atlas",
+				Type: "mongodb-atlas",
```
makesitgo, line 176: we should probably move this to a `const` somewhere, I'm thinking probably in `realm` package...

internal/commands/app/create_inputs.go
```diff
@@ -153,56 +154,69 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
-	dsCluster := dataSourceCluster{
-		Name: "mongodb-atlas",
-		Type: "mongodb-atlas",
-		Config: configCluster{
-			ClusterName:         clusterName,
-			ReadPreference:      "primary",
-			WireProtocolEnabled: false,
-		},
+	if len(dsClusters) == 0 {
+		return []dataSourceCluster{}, errors.New("failed to find Atlas cluster")
```
makesitgo, line 185: i think we want to tweak this logic a bit... previously when we only allowed one cluster, we would return this error if we failed to match that name from the returned list of clusters.

preserving that behavior would entail returning this error where we `continue` above (which would only tell the user about the first cluster not found, not potentially all of them not found).  so maybe we want to extend the error to find all missing first and communicate that back to the user.

my last idea here is: we should check with product to see if this warrants a "failure" at all.  perhaps we want to collect any missing specified cluster/datalake names and print that out to the user as a warning.  I can help follow up on that with you, but feel free to start a thread with drew/sumedha on that one

internal/commands/app/create_inputs.go
```diff
@@ -153,56 +154,69 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
-	return dsCluster, nil
+	return dsClusters, nil
 }
 
-func (i *createInputs) resolveDataLake(client atlas.Client, groupID string) (dataSourceDataLake, error) {
+func (i *createInputs) resolveDataLakes(client atlas.Client, groupID string) ([]dataSourceDataLake, error) {
 	dataLakes, err := client.DataLakes(groupID)
 	if err != nil {
-		return dataSourceDataLake{}, err
+		return []dataSourceDataLake{}, err
```
makesitgo, line 193: no need to allocate an array here, just `return nil, err`

internal/commands/app/create_inputs.go
```diff
@@ -153,56 +154,69 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
+	var dsDataLakes []dataSourceDataLake
+	for _, dataLakeName := range i.DataLakes {
+		if _, ok := existingDataLakes[dataLakeName]; !ok {
+			continue
+		}
+
+		dsDataLakes = append(dsDataLakes,
+			dataSourceDataLake{
+				Name: "mongodb-datalake",
+				Type: "datalake",
```
makesitgo, line 210: same thing about consts, also I missed this in the first implementation but is this really accurate that a datalake would have a different name than type?  (cc: @nathantfrank)

if so, then only the `Type` here needs to be a const

internal/commands/app/create_inputs.go
```diff
@@ -153,56 +154,69 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
+			})
 	}
-	dsDataLake := dataSourceDataLake{
-		Name: "mongodb-datalake",
-		Type: "datalake",
-		Config: configDataLake{
-			DataLakeName: dataLakeName,
-		},
+	if len(dsDataLakes) == 0 {
+		return []dataSourceDataLake{}, errors.New("failed to find Atlas data lake")
```
makesitgo, line 217: same thing about error handling here as clusters above

internal/commands/app/create_inputs.go
```diff
@@ -228,11 +242,11 @@ func (i createInputs) args(omitDryRun bool) []flags.Arg {
 	if i.Environment != realm.EnvironmentNone {
 		args = append(args, flags.Arg{flagEnvironment, i.Environment.String()})
 	}
-	if i.Cluster != "" {
-		args = append(args, flags.Arg{flagCluster, i.Cluster})
+	if len(i.Clusters) > 0 {
+		args = append(args, flags.Arg{flagCluster, strings.Join(i.Clusters, ",")})
```
makesitgo, line 246: also probably want to preserve the `!= ""` check somehow.  although now this likely depends on where we land about errors/warnings for any unresolved, specified data source names

makesitgo (commented):
i didn't look too too closely at tests here, I think we want to sync up with product first.  first batch of comments/cleanup here

internal/commands/app/create.go
```diff
@@ -201,11 +211,11 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 	rows = append(rows, map[string]interface{}{"Info": "Client App ID", "Details": appRealm.ClientAppID})
 	rows = append(rows, map[string]interface{}{"Info": "Realm Directory", "Details": dir})
 	rows = append(rows, map[string]interface{}{"Info": "Realm UI", "Details": fmt.Sprintf("%s/groups/%s/apps/%s/dashboard", profile.RealmBaseURL(), appRealm.GroupID, appRealm.ID)})
-	if dsCluster.Name != "" {
-		rows = append(rows, map[string]interface{}{"Info": "Data Source (Cluster)", "Details": dsCluster.Name})
+	if len(dsClusters) > 0 {
+		rows = append(rows, map[string]interface{}{"Info": "Data Source (Clusters)", "Details": strings.Join(clusterNames[:], ", ")})
```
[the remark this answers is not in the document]
jandersongo: these cluster names are checked [here](https://github.com/10gen/realm-cli/blob/362361d1ee6a774e88a3451237ec8dc9ec1332b6/internal/commands/app/create.go#L169) while filling clusterNames []string

internal/commands/app/create.go
```diff
@@ -201,11 +211,11 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
 	}
-	if dsDataLake.Name != "" {
-		rows = append(rows, map[string]interface{}{"Info": "Data Source (Data Lake)", "Details": dsDataLake.Name})
+	if len(dsDataLakes) > 0 {
+		rows = append(rows, map[string]interface{}{"Info": "Data Source (Data Lakes)", "Details": strings.Join(dataLakeNames[:], ", ")})
```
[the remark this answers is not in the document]
jandersongo: these data lake names are checked [here](https://github.com/10gen/realm-cli/blob/362361d1ee6a774e88a3451237ec8dc9ec1332b6/internal/commands/app/create.go#L185) while filling dataLakeNames []string

internal/commands/app/create_inputs.go
nathantfrank: Yes this is the case.

internal/commands/app/create_inputs.go
jandersongo: at the end of this file maybe? https://github.com/10gen/realm-cli/blob/362361d1ee6a774e88a3451237ec8dc9ec1332b6/internal/cloud/realm/realm.go

makesitgo: that could work, or just thinking forward a little I think creating a `services.go` file wouldn't be the worst thing here either (we have a similarly small `auth_providers.json` file)

internal/cloud/realm/services.go
```diff
@@ -0,0 +1,7 @@
+package realm
+
+// set of supported data source types
+const (
+	ClusterType  = "mongodb-atlas"
```
makesitgo, line 5: [nit] rename this to `ServiceTypeCluster`

internal/cloud/realm/services.go
```diff
@@ -0,0 +1,7 @@
...
+	DataLakeType = "datalake"
```
makesitgo, line 6: [nit] rename this to `ServiceTypeDatalake` (let's start to make consistent the casing of "datalake" here ... I think we're settled on it being all one word/no hyphens/no extra camel cases)

internal/commands/app/create.go
```diff
@@ -228,25 +231,34 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
-			"type": dsCluster.Type,
-			"config": map[string]interface{}{
-				"clusterName":         dsCluster.Config.ClusterName,
-				"readPreference":      dsCluster.Config.ReadPreference,
-				"wireProtocolEnabled": dsCluster.Config.WireProtocolEnabled,
-			},
-		})
+	clusterNames := make([]string, 0, len(dsClusters))
+	for _, dsCluster := range dsClusters {
+		if dsCluster.Name != "" {
```
makesitgo, line 236: is this condition possible still?  seems like not

if it is, i would only suggest inverting the `if` clause here and using `continue`, so we can un-indent the `local.AddDataSource` block below

```go
if dsCluster.Name == "" { continue }
local.addDataSource(...)
```

internal/commands/app/create.go
```diff
@@ -228,25 +231,34 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
-			"name": dsDataLake.Name,
-			"type": dsDataLake.Type,
-			"config": map[string]interface{}{
-				"dataLakeName": dsDataLake.Config.DataLakeName,
-			},
-		})
+
+	dataLakeNames := make([]string, 0, len(dsDataLakes))
+	for _, dsDataLake := range dsDataLakes {
+		if dsDataLake.Name != "" {
```
makesitgo, line 252: likewise here, is this condition still applicable?  same comments from above if so

internal/commands/app/create_inputs.go
```diff
@@ -156,56 +166,117 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
+		return nil, err
 	}
-	var clusterName string
-	for _, cluster := range clusters {
-		if i.Cluster == cluster.Name {
-			clusterName = cluster.Name
-			break
-		}
+
+	existingClusters := map[string]interface{}{}
```
makesitgo, line 175: oops, sorry I should have clarified this one better.  so the purpose of `struct{}` is to have the smallest allocations possible (useful for when you're only tracking unique key values).  so something like this:

```go
existingClusters := map[string]struct{}{} // another pattern you'll see is people using `bool` here instead
for _, c := range clusters {
  existingClusters[c.Name] = struct{}{} // but `struct{}` has just slightly a smaller memory footprint
}
// can check out for more info: https://gist.github.com/davecheney/3be245c92b61e5045f75
```

now full disclosure, i may be missing something here.  if we actually do need the value from this map at some point then this won't work (but considering you got away with `interface{}` here already I'm guessing that's not the case)

internal/commands/app/create_inputs.go
```diff
@@ -156,56 +166,117 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
+			nonExistingClusters = append(nonExistingClusters, clusterName)
+			continue
+		}
+
+		serviceName := clusterName
+		if len(i.ClusterServiceNames) > idx {
+			serviceName = i.ClusterServiceNames[idx]
+		} else {
+			if !ui.AutoConfirm() {
+				if err := ui.AskOne(&serviceName, &survey.Input{Message: fmt.Sprintf("Enter a Service Name for Cluster '%s'", clusterName), Default: serviceName}); err != nil {
```
makesitgo, line 193: [style nit] might be helpful to put a few line breaks in here...

```go
if err := ui.AskOne(&serviceName, &survey.Input{
  Message: fmt.Sprintf("Enter a Service Name for Cluster '%s'", clusterName),
  Default: serviceName,
}); err != nil {
  return nil, err
}
```

internal/commands/app/create_inputs.go
```diff
@@ -156,56 +166,117 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
-		Name: "mongodb-atlas",
-		Type: "mongodb-atlas",
-		Config: configCluster{
-			ClusterName:         clusterName,
-			ReadPreference:      "primary",
-			WireProtocolEnabled: false,
-		},
+
+	if len(nonExistingClusters) > 0 {
+		ui.Print(terminal.NewWarningLog("Please note, the following Atlas clusters '%s' were not linked because they could not be found", strings.Join(nonExistingClusters[:], ", ")))
```
makesitgo, line 212: ok so this is almost there, just a couple of notes:

* start the warning message as: `Note: The following Atlas clusters...` (no "Please", capital "The")
* let's invert the structure a bit so the "dynamic" portion of the message always is at the end
  * `Note: The following Atlas clusters were not linked because they could not be found: %s`
  * we don't need to wrap the whole list of separated cluster names with `''`
* I'd like to move this Warning and Confirm outside the scope of this function and to the command `Handler`
  * We should instead just add `nonExistingClusters` (a `[]string`) to the return signature of this function
  * We can do the same for `nonExistingDatalakes`
  * The warning message can then account for both `clusters` and `datalakes`, with a single confirm for the user, before potentially proceeding
* in the case of `!proceed`, we don't need to return an error, just `nil` is fine

makesitgo (commented):
held off on the review of the tests for now.  there's one fairly significant refactor I'm suggesting here that probably shifts a bunch of that test code around

let me know if you have any questions here.  but outside of this this is looking to be in a great spot

internal/commands/app/create.go
```diff
@@ -98,20 +101,39 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 		return err
 	}
 
-	var dsCluster dataSourceCluster
-	if cmd.inputs.Cluster != "" {
-		dsCluster, err = cmd.inputs.resolveCluster(clients.Atlas, groupID)
+	var nonExistingDataSources []string
+
+	var dsClusters []dataSourceCluster
```
makesitgo, line 106: [meganit] micro-optimization here...

if you `var dsClustersMissing []string` alongside this here (and do the same with `var dsDatalakesMissing []string`), then further down (around L#126 right here) you could initialize `nonExistingDataSources` with `make([]string, 0, len(dsClustersMissing) + len(dsDatalakesMissing))` (and ultimately preventing any re-allocations with `nonExistingDataSources`)

internal/commands/app/create.go
```diff
@@ -98,20 +101,39 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
+	if len(cmd.inputs.Datalakes) > 0 {
+		var nonExistingDatalakes []string
+		dsDatalakes, nonExistingDatalakes, err = cmd.inputs.resolveDatalakes(ui, clients.Atlas, groupID)
 		if err != nil {
 			return err
 		}
+		nonExistingDataSources = append(nonExistingDataSources, nonExistingDatalakes...)
+	}
+
+	if len(nonExistingDataSources) > 0 {
```
makesitgo, line 126: after initializing the `nonExistingDataSources` with an initial capacity here, for display purposes I would then loop through each "missing" array:

```go
for _, missingCluster := range dsClustersMissing {
  nonExistingDataSources = append(nonExistingDataSources, fmt.Sprintf("'%s'", missingCluster))
  // likewise for datalakes
}
```

i don't think we allow spaces in cluster or datalake names, but if we did those quotes will go a long way here

internal/commands/app/create.go
```diff
@@ -98,20 +101,39 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
+		ui.Print(terminal.NewWarningLog("Note: The following data sources were not linked because they could not be found: %s", strings.Join(nonExistingDataSources[:], ", ")))
```
makesitgo, line 127: you should be able to omit the `[:]` here

internal/commands/app/create.go
```diff
@@ -98,20 +101,39 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
+		if !ui.AutoConfirm() {
```
makesitgo, line 128: I missed this earlier but `ui.Confirm` already takes into account `ui.AutoConfirm`, so this wrapping `if` is not necessary

internal/commands/app/create_test.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/10gen/realm-cli/internal/local"
 	"github.com/10gen/realm-cli/internal/utils/test/assert"
 	"github.com/10gen/realm-cli/internal/utils/test/mock"
+	"go.mongodb.org/mongo-driver/bson/primitive"
```
makesitgo, line 20: [nit] move this down to the lower import group

internal/commands/app/create_test.go
```diff
@@ -599,39 +600,55 @@ func TestAppCreateHandler(t *testing.T) {
...
 					return []atlas.Cluster{{Name: "test-cluster"}}, nil
 				},
 			},
 		},
 		{
-			description: "should create minimal project with a data lake data source when data lake is set",
-			dataLake:    "test-datalake",
+			description:         "should create minimal project with multiple cluster data sources when clusters are set",
+			clusters:            []string{"test-cluster", "test-cluster-2"},
+			clusterServiceNames: []string{"mongodb-atlas", "mongodb-atlas"},
```
makesitgo, line 623: for the sake of mimic-ing real behavior, can we have unique names across these?

internal/commands/app/create_test.go
```diff
@@ -1059,6 +1084,92 @@ func TestAppCreateHandler(t *testing.T) {
...
+
+		for _, tc := range []struct {
+			description string
+			response    string
+			appCreated  bool
+		}{
+			{
+				description: "and quit if not confirmed",
+				response:    "no",
+				appCreated:  false,
```
makesitgo, line 1135: [meganit] can just remove this line

internal/commands/app/create_test.go
```diff
@@ -1059,6 +1084,92 @@ func TestAppCreateHandler(t *testing.T) {
...
+				go func() {
+					defer close(doneCh)
+					console.ExpectString(fmt.Sprintf("Note: The following data sources were not linked because they could not be found: %s, %s, %s, %s",
+						dummyClusters[0], dummyClusters[1], dummyDatalakes[0], dummyDatalakes[1]))
+					console.ExpectString("Would you still like to create the app?")
+					console.SendLine(tc.response)
+					console.ExpectEOF()
+				}()
+
+				appCreated := false
```
makesitgo, line 1161: [style nit] `var appCreated bool` suffices here as well

internal/commands/app/create_inputs_test.go
```diff
@@ -372,111 +373,659 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
+		inputs := createInputs{
+			newAppInputs:        newAppInputs{Name: testApp.Name, Project: testApp.GroupID},
+			Clusters:            []string{"test-cluster-1", dummyClusters[0], dummyClusters[1]},
+			ClusterServiceNames: []string{"mongodb-atlas"},
+		}
+
+		doneCh := make(chan (struct{}))
+		go func() {
+			defer close(doneCh)
+			console.ExpectString(fmt.Sprintf("Note: The following data sources were not linked because they could not be found: %s", strings.Join(dummyClusters[:], ", ")))
```
makesitgo, line 492: [nit] shouldn't need the `[:]` here, and considering its a test, you don't necessarily _need_ the `fmt.Sprintf` here and could just write the expected string.  your call on that last one though

internal/commands/app/create_inputs_test.go
```diff
@@ -372,111 +373,659 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
+				defer teardown()
+
+				_, console, _, ui, err := mock.NewVT10XConsole()
+				assert.Nil(t, err)
+				defer console.Close()
+
+				doneCh := make(chan (struct{}))
+				go func() {
+					defer close(doneCh)
+					console.ExpectString(fmt.Sprintf("Note: The following data sources were not linked because they could not be found: %s", strings.Join(dummyClusters[:], ", ")))
```
makesitgo, line 569: same nit here about `[:]` and/or just writing the expected string as is (without `fmt.Sprintf`)

internal/commands/app/create_inputs_test.go
```diff
@@ -372,111 +373,659 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
+					defer close(doneCh)
+					tc.procedure(console)
+				}()
+
+				inputs := createInputs{newAppInputs: newAppInputs{Name: "test-app"},
+					Clusters:            tc.clusterNames,
+					ClusterServiceNames: tc.clusterServiceNames,
+				}
+
+				ds, _, _ := inputs.resolveClusters(ui, ac, "123")
```
makesitgo, line 665: generally still advisable to handle `err` here even if its expected to be `nil`... just a `assert.Nil(t, err)` on the next line suffices

internal/commands/app/create_inputs_test.go
```diff
@@ -372,111 +373,659 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
+		inputs := createInputs{
+			newAppInputs:         newAppInputs{Name: testApp.Name, Project: testApp.GroupID},
+			Datalakes:            []string{"test-datalake-1", dummyDatalakes[0], dummyDatalakes[1]},
+			DatalakeServiceNames: []string{"mongodb-datalake"},
+		}
+
+		doneCh := make(chan (struct{}))
+		go func() {
+			defer close(doneCh)
+			console.ExpectString(fmt.Sprintf("Note: The following data sources were not linked because they could not be found: %s", strings.Join(dummyDatalakes[:], ", ")))
```
makesitgo, line 821: usual comment on `[:]` here and opting to just write the string out if you feel it makes things clearer

internal/commands/app/create_inputs_test.go
```diff
@@ -372,111 +373,659 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
+
+		for _, tc := range []struct {
+			description string
+			response    string
+			appCreated  bool
+		}{
+			{
+				description: "and quit if not confirmed",
+				response:    "no",
+				appCreated:  false,
```
makesitgo, line 880: can just remove this line

internal/commands/app/create_inputs_test.go
```diff
@@ -372,111 +373,659 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
+					tc.procedure(console)
+				}()
+
+				inputs := createInputs{
+					newAppInputs:         newAppInputs{Name: "test-app"},
+					Datalakes:            tc.datalakeNames,
+					DatalakeServiceNames: tc.datalakeServiceNames,
+				}
+
+				ds, _, _ := inputs.resolveDatalakes(ui, ac, "123")
```
makesitgo, line 996: should assert on the `err` here

makesitgo (commented):
nothing major as a comment from me, but there's probably enough cleanup to warrant one last look before we ✅  this

cannot say it enough, but excellent turnaround on all of this!

internal/commands/app/create_inputs_test.go
```diff
@@ -372,111 +372,661 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
+		inputs := createInputs{
+			newAppInputs:        newAppInputs{Name: testApp.Name, Project: testApp.GroupID},
+			Clusters:            []string{"test-cluster-1", dummyClusters[0], dummyClusters[1]},
+			ClusterServiceNames: []string{"mongodb-atlas"},
+		}
+
+		doneCh := make(chan (struct{}))
+		go func() {
+			defer close(doneCh)
+			console.ExpectString(("Note: The following data sources were not linked because they could not be found: 'test-cluster-dummy-1', 'test-cluster-dummy-2'"))
```
makesitgo, line 491: [meganit] stray double parens here (leftover from the `fmt.Sprintf` likely)

internal/commands/app/create_inputs_test.go
```diff
@@ -372,111 +372,661 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
...
+					tc.procedure(console)
+				}()
+
+				inputs := createInputs{
+					newAppInputs:         newAppInputs{Name: "test-app"},
+					Datalakes:            tc.datalakeNames,
+					DatalakeServiceNames: tc.datalakeServiceNames,
+				}
+
+				ds, _, _ := inputs.resolveDatalakes(ui, ac, "123")
```
makesitgo, line 995: missed actually setting `err` here ... here's the edge-case argument _in favor of_ `namedErr` (vs. `err`) 😬

makesitgo (approved):
two super minor things, but with that, a clean evergreen build, and fixed up commits (we can go over the rebase/squash/fixup sequence if you'd like) this is good to go 🎉 

🆒  beans

internal/commands/app/create_inputs_test.go
jandersongo: ah! good catch!

PR #167, opened 2021-06-12
REALMC-8260: Fix up disabling of template selection

UnicodeSnowman:
whoops, wasn't paying attention when I commented this out in the last PR... fail.


andrewzlchen (approved):
lgtm!

PR #169, opened 2021-06-17
REALMC-8832: Update command flag descriptions

makesitgo:


makesitgo:
worked with @kpham1231 on this one... things to look out for while reviewing:
 - consistency with capitalization ("Realm app", "Service", "Function") and punctuation (don't end usages with `.`, use `;` b/w value descriptions, `[]` around the docs links)
 - consistency across commands for similar flags
 - typos or any other oddities

internal/commands/function/run_inputs.go
```diff
@@ -14,13 +14,14 @@ import (
 
 const (
 	flagFunctionName      = "name"
-	flagFunctionNameUsage = "specify the name of the function to run"
+	flagFunctionNameUsage = "Specify the name of the function to run"
 
-	flagFunctionArgs      = "args"
-	flagFunctionArgsUsage = "specify the args to pass to your function"
+	flagFunctionArgs = "args"
+	// TODO(REALMC-XXXX): Add docs link to the end of the usage string here: [Learn more: http://docs.link]
```
makesitgo, line 20: ⚠️  need to reach out to docs to get a JIRA to link to

internal/commands/logs/list_inputs.go
```diff
@@ -14,19 +14,21 @@ import (
...
+	flagTypeUsage = `Specify the type(s) of logs to list (Default value: <none>; Allowed values: "auth", "function", "push", "service", "trigger", "graphql", "sync", "schema")`
 
 	flagErrors      = "errors"
-	flagErrorsUsage = "specify to view only error logs"
+	flagErrorsUsage = "View your Realm app's error logs"
 
-	flagStartDate      = "start"
-	flagStartDateUsage = "specify the start date to begin listing logs from"
+	flagStartDate = "start"
+	// TODO(REALMC-XXXX): Include docs link at the end of this usage: [Learn more: http://docs.link]
```
makesitgo, line 23: ⚠️  need to reach out to docs to get a JIRA to link to

internal/commands/logs/list_inputs.go
```diff
@@ -14,19 +14,21 @@ import (
...
+	flagStartDateUsage = `Specify when to begin listing logs (Allowed format: "2006-01-02[T15:04:05.000-0700]")`
 
-	flagEndDate      = "end"
-	flagEndDateUsage = "specify the end date to finish listing logs from"
+	flagEndDate = "end"
+	// TODO(REALMC-XXXX): Include docs link at the end of this usage: [Learn more: http://docs.link]
```
makesitgo, line 27: ⚠️  need to reach out to docs to get a JIRA to link to

internal/cli/inputs.go
```diff
@@ -17,13 +17,13 @@ import (
...
-	flagAppUsage = "the remote Realm app name or id"
+	flagAppUsage = "Specify the name or ID of a Realm app"
 
 	flagProject      = "project"
-	flagProjectUsage = "the MongoDB cloud project id"
+	flagProjectUsage = "Specify the ID of a MongoDB Atlas project"
 
 	flagProduct      = "product"
-	flagProductUsage = "the Realm app product (can be specified multiple times)"
+	flagProductUsage = `Specify the Realm app product(s) (Allowed values: "standard", "atlas")`
```
jandersongo, line 26: the quotes around flagProductUsage `str` are different than flagProjectUsage & flagAppUsage "str"
can we keep them consistent?

internal/commands/app/create_inputs.go
```diff
@@ -20,26 +20,26 @@ import (
...
+	flagDatalakeServiceNameUsage = "Specify the Realm app Service name to reference your Atlas data lake"
 
 	flagTemplate      = "template"
-	flagTemplateUsage = "create your app from an available template"
+	flagTemplateUsage = "Create your Realm app from an available template"
 
 	flagDryRun      = "dry-run"
 	flagDryRunShort = "x"
-	flagDryRunUsage = "include to run without writing any changes to the file system nor deploying any changes to the Realm server"
+	flagDryRunUsage = "Run without neither writing any changes to the local filepath nor pushing any changes to the Realm server"
```
jandersongo, line 42: [grammar nit] "without neither" is a double negative. Maybe something like this would be more clear:
"Run without writing any changes to the local filepath or pushing any changes to the Realm server"

makesitgo: hahah so @kpham1231 and i waffled on this for a while...

because it's in the negative sense, it stands to reason `nor` is our operative word in the middle (as opposed to `or`).  omitting "neither" ends up sounding bit odd too: "Run without writing changes ... nor pushing changes ...".

i'll ultimately defer to what kevin wants here.  maybe he'll have second thoughts on this after seeing your comment?

kpham1231: ah good catch LOL I think then let's go with `"Run without writing any changes to the local filepath nor pushing any changes to the Realm server"` since both "without... or" and "without... nor" are both grammatically correct, though the former is likely used more frequently

jandersongo (approved):

internal/commands/function/run_inputs.go
nlarew: https://docs.mongodb.com/realm/functions/call-a-function/#call-from-realm-cli

internal/commands/logs/list_inputs.go
nlarew: https://docs.mongodb.com/realm/logs/cli/#view-logs-for-a-date-range

internal/commands/logs/list_inputs.go
nlarew: https://docs.mongodb.com/realm/logs/cli/#view-logs-for-a-date-range

internal/cli/inputs.go
```diff
@@ -17,13 +17,13 @@ import (
 const (
 	flagApp      = "app"
 	flagAppShort = "a"
-	flagAppUsage = "the remote Realm app name or id"
+	flagAppUsage = "Specify the name or ID of a Realm app"
```
nathantfrank, line 20: [q] Should we mention Client App ID vs the Realm App ID?

makesitgo: Nowhere in the CLI is the App ID (as you know the, the ObjectID) really a part of the user experience.  So for brevity, we generally (and this is existing behavior) refer to what we call the client app id as just the app id (notice how we rename it `"app_id"` when exporting out the `realm_config.json`/`config.json`/etc.

Going to leave this as is, until user feedback indicates we need to do otherwise

nathantfrank: Yeah, just didn't want to introduce confusion with using the term `ID` also for a `MongoDB Atlas project`, which is the `ObjectID`.

PR #172, opened 2021-06-22
REALMC-8257 Add templates fetching for export / pull 

jo-tse-mongo:


internal/commands/pull/inputs.go
```diff
@@ -109,3 +118,46 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
 
 	return app, nil
 }
+
+func (i *inputs) resolveTemplate(ui terminal.UI, realmClient realm.Client, groupID, appID string) (*zip.Reader, []string,  error) {
+	if i.TemplateID == "" {
+		return nil, nil, nil
+	}
+
+	var paths []string
```
makesitgo, line 127: check me on this one, but I think you could initialize this with `paths := []string{frontendPath}`, and still overwrite it with `AskOne` (and ultimately avoid the `else` down below)

internal/commands/pull/inputs.go
```diff
@@ -109,3 +118,46 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
...
+	}
+
+	compatibleTemplates, err := realmClient.CompatibleTemplates(groupID, appID)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	for _, template := range compatibleTemplates {
+		// This is the ID that they selected
+		if template.ID == i.TemplateID {
```
makesitgo, line 152: [question] do we ever give the user the ability to _select_ the template id?  in `app create`, we more or less always prompt them for a selection (unless `-y` is set).  especially with things like IDs, i think its helpful to have a way where the user does not need to know the specific value, but can otherwise pick it

internal/commands/pull/inputs.go
```diff
@@ -109,3 +118,46 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
...
+			// Fetch the template
+			templateZip, err := realmClient.ClientTemplate(groupID, appID, template.ID)
+			if err != nil {
+				return nil, nil, err
+			}
+			return templateZip, paths, nil
+		}
+	}
+
+	return nil, nil, fmt.Errorf("templateID %s is not compatible with this app", i.TemplateID)
```
makesitgo, line 162: try to keep code-like words out of the user-facing messaging
```suggestion
	return nil, nil, fmt.Errorf("template '%s' is not compatible with this app", i.TemplateID)
```

makesitgo (changes requested):
just did a pass through of the code files (no tests yet).  i have a few questions, probably all related to my unfamiliarity with the TSA project, so I imagine once I have a bit better context here I can dive into the rest of the code

internal/cloud/realm/client.go
```diff
@@ -72,6 +72,7 @@ type Client interface {
 
 	Templates() ([]Template, error)
 	ClientTemplate(groupID, appID, templateID string) (*zip.Reader, error)
+	CompatibleTemplates(groupID, appID string) ([]Template, error)
```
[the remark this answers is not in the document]
jo-tse-mongo: `Templates` _is_ fetching all templates, and `CompatibleTemplates` is a filtered list of templates that are compatible with the app. They both return only 
```
        ID   string `json:"id"`
	Name string `json:"name"`
```
`ClientTemplate` is the most granular of the endpoints, which fetches 
 everything about the template specified from `templateID`, and for realm-cli returns a zip package containing the template information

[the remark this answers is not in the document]
andrewzlchen: Yeah, I'm cool with renaming `Templates` to be `AllTemplates` to make that more clear. 

I think it should be relatively apparent that the endpoints return the same data given the interface definition

internal/cloud/realm/templates.go
```diff
@@ -48,6 +50,10 @@ func (c *client) ClientTemplate(groupID, appID, templateID string) (*zip.Reader,
 	if resErr != nil {
 		return nil, resErr
 	}
+	if res.StatusCode == http.StatusNoContent {
+		// No client exists for this template so there is nothing to return
```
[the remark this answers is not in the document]
jo-tse-mongo: From my understanding, we return a `204` when there is no `ClientPath` in the template that we're trying to fetch from the `ClientTemplate` endpoint. 
The `ClientPath` is used to get the template content from Github, which is why we need one attached to the template itself. 

So, if there is no `ClientPath`, then we can't get the information for the Template from Github, and therefore, the return for this fetch should be `nil`.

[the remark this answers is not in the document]
andrewzlchen: I don't think we actually need this check in this function. It should be for the `ClientTemplate` function.

This is referring to the changes made here: https://github.com/10gen/baas/pull/4381/files

Basically, in the process of fetching a client template, sometimes the template that we want to fetch the client for does not have an associated frontend client (like in the case of the database triggers template).

We would return a `http.StatusNoContent` to denote that the template that you're trying to find a client for does not have a template.

internal/commands/pull/command.go
```diff
@@ -60,17 +68,36 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+		return err
+	}
+
 	pathTarget, zipPkg, err := cmd.doExport(profile, clients.Realm, app.GroupID, app.ID)
 	if err != nil {
 		return err
 	}
 
+	var templatePathTargets []string
+	if templateZipPkg != nil {
```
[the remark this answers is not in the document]
jo-tse-mongo: 👀 I think I had something else in that block before
Will remove it in an update

internal/commands/pull/command.go
```diff
@@ -60,17 +68,36 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+		}
+	}
+
 	pathRelative, err := filepath.Rel(profile.WorkingDirectory, pathTarget)
 	if err != nil {
 		return err
 	}
 
-	proceed, err := checkAppDestination(ui, pathTarget)
+	proceed, err := checkPathDestination(ui, pathTarget)
```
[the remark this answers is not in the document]
jo-tse-mongo: No, since it'd make the directory for the `app` in making the template directories. 
I could move this above the template `checkPathDestination`s or only do this if `templateZipPkg` is nil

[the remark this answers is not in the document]
makesitgo: I probably still don't have the full picture of what we expect/want to show the user if a template is involved vs when its not.  at least for when there's no template, we need to show this single prompt to tell them where we're writing contents.

by the sounds of it, it seems that when a template _is_ involved, we're pushing the "write logic" to potential 2 spots (`appDir/backend` and `appDir/frontend`) and so it would stand to reason with me that we'd want either:
  * a single confirm prompting the user about these two new spots that we're writing to
  * or two prompts confirming individually each spot

i think the question then becomes, do want to allow the user to ever partially write a template?  as in, if it had both frontend and backend paths, do you still want the user to have the ability to opt into one or both of them?  in other words, can a template still be "usable" if its only partially written ("partially" here being omitting either frontend, backend, or both)?  the answer to those questions likely drives what type of prompts we ultimately want to write for the user to see...

we should strive to keep the amount of expected prompts to a minimum (think "less button clicks" in the UI, make it streamlined).  so at the very least, we should not be prompting a third time here

internal/cloud/realm/client.go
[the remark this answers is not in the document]
makesitgo: cool, and just to clarify I think that also means we can then remove `Compatible` from this new function name?

internal/cloud/realm/templates.go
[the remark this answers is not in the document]
makesitgo: hahaha i think I thought it was in a different function too 🙄 ... checking for `NoContent` makes more sense to me here.

my only thought is: it leaves us a bit more flexible (and probably more go-like) if we add `bool` to list of return values to this function (before `error`, after `*zip.Reader`) so that we're not relying on a "is nil" check on the returned object to otherwise confirm a valid response.  id rather we have a `zipPkg, ok, err := realmClient.ClientTemplate` type of usage

internal/commands/pull/command.go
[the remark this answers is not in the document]
jo-tse-mongo: See @andrewzlchen 's comment below, but I was mostly following what was [written on the ticket](https://jira.mongodb.org/browse/REALMC-8257) in terms of where to put the templates

internal/commands/pull/command.go
```diff
@@ -90,6 +117,15 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 	}
 	ui.Print(terminal.NewTextLog("Saved app to disk"))
 
+	if templateZipPkg != nil {
+		for _, templatePathTarget := range templatePathTargets {
+			if err := local.WriteZip(templatePathTarget, templateZipPkg); err != nil {
```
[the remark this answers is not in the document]
jo-tse-mongo: See @andrewzlchen 's comment below; I'm going to update the PR to do what's in the comment

internal/cloud/realm/templates.go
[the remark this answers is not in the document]
andrewzlchen: i agree. that's a good idea

internal/commands/pull/inputs.go
jo-tse-mongo: @sumedham  Does this look fine from a design perspective? 
We're going with a new flow of work as Andrew detailed above

sumedham: this sgtm

internal/cloud/realm/client.go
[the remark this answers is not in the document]
jo-tse-mongo: @andrewzlchen I'm iffy from making this just `Templates` because the templates that we fetch for this app aren't necessarily the templates that this app is currently using; they're _potential_ templates that the app _could_ use because they're compatible

internal/cloud/realm/templates.go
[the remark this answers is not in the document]
jo-tse-mongo: I forgot to add this, but I will in just a bit

internal/cloud/realm/client.go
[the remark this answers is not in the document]
andrewzlchen: yeah i think we should call them:

`AllTemplates` and `CompatibleTemplates`

If `CompatibleTemplates` is not preferable, maybe we can call them `SiblingTemplates`, `SupportedTemplates` or something similar

internal/cloud/realm/templates.go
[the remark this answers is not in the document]
andrewzlchen: hmm i dont see this change for some reason. maybe there was some miscommunication?

internal/commands/pull/command.go
```diff
@@ -60,23 +61,47 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		return err
 	}
 
-	pathTarget, zipPkg, err := cmd.doExport(profile, clients.Realm, app.GroupID, app.ID)
+	templateZipPkgs, ok, err := cmd.inputs.resolveClient(ui, clients.Realm, app.GroupID, app.ID)
```
andrewzlchen, line 64: nit: `clientZipPkgs`

internal/commands/pull/inputs.go
```diff
@@ -109,3 +117,61 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
 
 	return app, nil
 }
+
+func (i *inputs) resolveClient(ui terminal.UI, realmClient realm.Client, groupID, appID string) (map[string]*zip.Reader, bool, error) {
+	compatibleTemplates, err := realmClient.CompatibleTemplates(groupID, appID)
```
andrewzlchen, line 122: i think this was where nick was trying to add the `bool` return param.

I think this should be something like

```go
// its possible that the request was successful, but there was no template for the current app.
// checking ok allows us to have more granular context for what went wrong
compatibleTemplates, ok, err := realmClient.CompatibleTemplates(groupID, appID)
if err != nil {
	// something went wrong
}
if !ok {
	// request succesful, but nothing returned
	return nil
}
// client exists so we handle it
```

andrewzlchen:
one nit that i've noticed is that the templates format seems to be a bit different between pull and create:

realm-cli pull
image


VS

realm-cli app create

image

Let's try to align with the create command where we add the template id to the selection

andrewzlchen (changes requested):
nice job! this is really close!

i only had nits and once they're resolved, I think this is pretty much good

andrewzlchen (approved):
should look good pending tests!

internal/cloud/realm/client.go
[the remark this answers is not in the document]
makesitgo: fine with it as `All`/`Compatible` here

internal/commands/pull/command.go
```diff
@@ -60,23 +62,47 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
 	}
 
-	pathRelative, err := filepath.Rel(profile.WorkingDirectory, pathTarget)
+	pathTarget, zipPkg, err := cmd.doExport(profile, clients.Realm, app.GroupID, app.ID)
 	if err != nil {
 		return err
 	}
 
-	proceed, err := checkAppDestination(ui, pathTarget)
+	if clientZipPkgs != nil {
```
makesitgo, line 75: i'd prefer _not_ to use the `nil`-presence of a map to decide to something.  could this just be `len(clientZipPkgs) == 0`?  this would technically allow the user to not select anything from the multi-prompt and then subsequently prompt them if they're ok to write contents to a directory, at which point they'll write nothing to that spot.  seems like a clunky UX to me

internal/commands/pull/command.go
```diff
@@ -60,23 +62,47 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
 	}
 
-	pathTarget, zipPkg, err := cmd.doExport(profile, clients.Realm, app.GroupID, app.ID)
+	clientZipPkgs, err := cmd.inputs.resolveClient(ui, clients.Realm, app.GroupID, app.ID)
 	if err != nil {
 		return err
 	}
 
-	pathRelative, err := filepath.Rel(profile.WorkingDirectory, pathTarget)
+	pathTarget, zipPkg, err := cmd.doExport(profile, clients.Realm, app.GroupID, app.ID)
```
makesitgo, line 70: maybe instead of `pathTarget` we can call this `pathBackend` now?  and maybe do one even better, we can have:

```go
pathApp, zipPkg, err := cmd.doExport(...)
pathBackend := pathApp // new reference allows us to keep things a bit more "immutable"/distinct

if clientZipPkgs != nil {
  pathFrontend := filepath.Join(pathApp, local.FrontendPath)

  // write zip files

  // now we update pathBackend, and pathApp can still remain accurate
  // if this line of code were to move for w/e reason, we don't risk creating a bad `pathFrontend` above
  pathBackend = filepath.Join(pathApp, local.BackendPath)
}
```

internal/commands/pull/command.go
```diff
@@ -60,23 +62,47 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
-	proceed, err := checkAppDestination(ui, pathTarget)
+	if clientZipPkgs != nil {
+		templatePath := filepath.Join(pathTarget, local.FrontendPath)
+		if proceed, err := checkPathDestination(ui, templatePath); err != nil {
+			return err
+		} else if !proceed {
+			return nil
+		}
+
+		for templateID, templateZipPkg := range clientZipPkgs {
```
makesitgo, line 83: just pointing out there'll be no deterministic order for how we write these templates, which would ultimately lead to subsequent runs of this command to have potentially different outputs.  not a *huge* concern, but also not "perfect".  might be worth returning some order from `resolveClientTemplates` as well, via a key slice and/or some sorting

internal/commands/pull/command.go
```diff
@@ -60,23 +62,47 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+			if err := local.WriteZip(templatePath, templateZipPkg); err != nil {
+				return err
+			}
+			ui.Print(terminal.NewTextLog(fmt.Sprintf("Saved template %s to disk", templateID)))
```
makesitgo, line 87: the `fmt.Sprintf` is unecessary here

internal/commands/pull/command.go
```diff
@@ -60,23 +62,47 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+		}
+
+		pathTarget = filepath.Join(pathTarget, local.BackendPath)
+	}
+
+	// App path
+	proceed, err := checkPathDestination(ui, pathTarget)
```
makesitgo, line 94: if you go with the suggested change above, make sure to now use `pathBackend` here

also, we end up using the same confirm prompt for both backend and frontend templates, and the messaging only changes with the filepath shown.  does it make sense to adjust the language on each to make it clear when we're writing the "backend stuff" or "frontend stuff" (I would imagine that would also only really want to happen if templates were involved in the first place)

if so, i think breaking out `checkPathDestination` to two funcs, one for `checkBackend` and one for `checkFrontend` would be a good idea here.  and `checkBackend` could then either use the existing code for when templates arent involved, or mention that this would be writing "backend stuff"...

internal/commands/pull/command.go
```diff
@@ -60,23 +62,47 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
 	if err != nil {
 		return err
 	} else if !proceed {
 		return nil
 	}
 
+	pathRelative, err := filepath.Rel(profile.WorkingDirectory, pathTarget)
```
makesitgo, line 101: `pathRelative` here should probably remain as the relative "app path" (and not the backend path, which it is now).

additionally, we should update the `dryRun` output here to talk about templates if they're involved.  ideally keep the same existing behavior if templates arent used, but if so, we could add on to the debug logs indicating that from the "app's relative path", backend code would live in `./backend` as well as list the various front end template paths that would have been written

internal/commands/pull/command_test.go
```diff
@@ -317,6 +334,204 @@ Successfully pulled app down: app
 		assert.Nil(t, err)
 		assert.Equal(t, "<html><body>hello world!</body></html>", string(modifiedData))
 	})
+
+	t.Run("with a template ID to export with", func(t *testing.T) {
+
```
makesitgo, line 339: [style nit] no need for the newline, could also just inline the `// needs to have...` alongside the `zipPkg, err := ...` line

internal/commands/pull/command_test.go
```diff
@@ -317,6 +334,204 @@ Successfully pulled app down: app
...
+				templateID:  "template_2",
+				templateZip: &templateZipPkg2.Reader,
+				expectedTemplate: `{
+  "egg": "scrambled",
+  "pancake": "plain",
+  "toast": "none"
+}`,
+			},
+		} {
+			t.Run(fmt.Sprintf("should export the corresponding template when the template ID %s is passed in", tc.templateID), func(t *testing.T) {
```
makesitgo, line 386: [nit] avoid punctuation and capitalization in test descriptions

```suggestion
			t.Run(fmt.Sprintf("should export the corresponding template when the template id %s is passed in", tc.templateID), func(t *testing.T) {
```

also, [question]: why is this a test table?  i wanted to comment that we shouldn't need the actual assert value in the test description ("template id" here), and other than different template contents I don't understand what different logic these tests are proving.  could/should this be reduced down to one test?  seems like you'd still be able to prove the multiple templates were written

internal/commands/pull/command_test.go
```diff
@@ -317,6 +334,204 @@ Successfully pulled app down: app
...
+				_, err := os.Stat(destination)
+				assert.Nil(t, err)
+
+				testData, readErr := ioutil.ReadFile(filepath.Join(destination, fmt.Sprintf("%s.json", tc.templateID)))
+				assert.Nil(t, readErr)
+				assert.Equal(t, tc.expectedTemplate, string(testData))
+			})
+		}
+
+		t.Run("should export all selected compatible templates if no template ID is passed in and the app is made with a template", func(t *testing.T) {
```
makesitgo, line 414:
```suggestion
		t.Run("should export all selected compatible templates if no template id is passed in and the app is made with a template", func(t *testing.T) {
```

this should also probably call out that "auto confirm is set to true" somehow, b/c the behavior of this command would be different if not (and if we don't have test coverage for that scenario, seems like we'd want it)

internal/commands/pull/command_test.go
```diff
@@ -317,6 +334,204 @@ Successfully pulled app down: app
...
+
+			out := new(bytes.Buffer)
+			ui := mock.NewUIWithOptions(mock.UIOptions{AutoConfirm: true}, out)
+
+			cmd := &Command{inputs{Project: "elsewhere", LocalPath: "app", TemplateID: "some-template-id"}}
+
+			err := cmd.Handler(profile, ui, cli.Clients{Realm: realmClient})
+			assert.Equal(t, "some kind of error", err.Error())
+		})
+		t.Run("should return an error if the template ID is not compatible with the app", func(t *testing.T) {
```
makesitgo, line 491:
```suggestion
		t.Run("should return an error if the template id is not compatible with the app", func(t *testing.T) {
```

internal/commands/pull/command_test.go
```diff
@@ -317,6 +334,204 @@ Successfully pulled app down: app
...
+			profile, teardown := mock.NewProfileFromTmpDir(t, "should_fail_also")
+			defer teardown()
+
+			out := new(bytes.Buffer)
+			ui := mock.NewUIWithOptions(mock.UIOptions{AutoConfirm: true}, out)
+
+			cmd := &Command{inputs{Project: "elsewhere", LocalPath: "app", TemplateID: "some-template-id"}}
+
+			err := cmd.Handler(profile, ui, cli.Clients{Realm: realmClient})
+			assert.Equal(t, "template some-template-id is not compatible with this app", err.Error())
```
makesitgo, line 505: prefer asserting equality on `errors.New("...")` and not using `err.Error()`

internal/commands/pull/command_test.go
```diff
@@ -317,6 +334,204 @@ Successfully pulled app down: app
...
+
+			// Check to see that export went as normal
+			destination := filepath.Join(profile.WorkingDirectory, "app")
+
+			_, err = os.Stat(destination)
+			assert.Nil(t, err)
+
+			testData, readErr := ioutil.ReadFile(filepath.Join(destination, "test.json"))
+			assert.Nil(t, readErr)
+			assert.Equal(t, "{\"egg\":\"corn\"}\n", string(testData))
```
makesitgo, line 532: [style nit] I want to say use backticks for string literals when writing valid json (or anything that needs `"`), but that forces you into adding a literal newline to this assert... so carry on as you see fit

internal/commands/pull/inputs.go
```diff
@@ -109,3 +117,62 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
...
+	if err != nil {
+		return nil, err
+	}
+	if !ok {
+		return nil, nil
+	}
+
+	if i.TemplateID != "" {
+		for _, template := range compatibleTemplates {
+			// This is the ID that they selected
```
makesitgo, line 132: this comment feels unnecessary

internal/commands/pull/inputs.go
```diff
@@ -109,3 +117,62 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
...
+			if template.ID == i.TemplateID {
+				// Fetch the template
```
makesitgo, line 134: also can probably just remove this comment

internal/commands/pull/inputs.go
```diff
@@ -109,3 +117,62 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
...
+				templateZip, err := realmClient.ClientTemplate(groupID, appID, template.ID)
+				if err != nil {
+					return nil, err
+				}
+				return map[string]*zip.Reader{template.ID: templateZip}, nil
+			}
+		}
+		return nil, fmt.Errorf("template %s is not compatible with this app", i.TemplateID)
```
makesitgo, line 142: should probably wrap this with `'%s'`

internal/commands/pull/inputs.go
```diff
@@ -109,3 +117,62 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
...
+	}
+
+	if len(compatibleTemplates) != 0 {
```
makesitgo, line 145: I asked a similar question up top, but is this case (an empty template array) different than when the `400` is returned from our server?  if not, then we probably don't need `bool` returned from the client `CompatibleTemplates` method.

also, you should invert this if to return early and then un-indent the following block

internal/commands/pull/inputs.go
```diff
@@ -109,3 +117,62 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
...
+		if proceed, err := ui.Confirm("Would you like to export with a template?"); err != nil {
```
makesitgo, line 146: we can also invert the `ok` check here to remove some layers of indenting

```go
if proceed, err := ui.Confirm(...); err != nil {
  return nil, err
} else if !proceed {
  return nil, nil
}

nameOptions := ...
```

internal/commands/pull/inputs.go
```diff
@@ -109,3 +117,62 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
...
+			}
+
+			templateIDs := make([]string, len(selectedTemplateIdxs))
+			for idx, selectedTemplateIdx := range selectedTemplateIdxs {
+				templateIDs[idx] = compatibleTemplates[selectedTemplateIdx].ID
+			}
+
+			result := make(map[string]*zip.Reader, len(templateIDs))
+			for _, templateID := range templateIDs {
+				templateZip, err := realmClient.ClientTemplate(groupID, appID, templateID)
```
makesitgo, line 168: this is the client method where we're returning `nil, nil`, and so now we're populating a map entry with a `nil` `*zip.Reader`.  make sure this is ok with the consuming code for this method (where we're writing the frontend templates to disk).

and if we did add the `bool` return here, then it would simply be a `if !ok { continue }` (and it would seem you'd also want to warn the user about such a situation somehow ... if that's the case please try to keep all the warning/debug messages printed in the command `Handler` code and not here in the inputs ... also make sure to point that out in the `dryRun` stuff too)

sorry for the run-on comment there, let me know if i can clarify anything there

internal/commands/pull/inputs_test.go
```diff
@@ -159,3 +160,169 @@ func TestPullInputsResolveRemoteApp(t *testing.T) {
...
+		}
+
+		_, ui := mock.NewUI()
+		input := inputs{}
+		templateZipMap, err := input.resolveClient(ui, realmClient, "some-group-id", "some-app-id")
+		assert.Nil(t, templateZipMap)
+		assert.Equal(t, "something went wrong", err.Error())
+	})
+
+	t.Run("with a template ID passed in", func(t *testing.T) {
```
makesitgo, line 187:
```suggestion
	t.Run("with a template id passed in", func(t *testing.T) {
```

internal/commands/pull/inputs_test.go
```diff
@@ -159,3 +160,169 @@ func TestPullInputsResolveRemoteApp(t *testing.T) {
...
+		t.Run("should return an error if fetching the ClientTemplate for this template fails", func(t *testing.T) {
```
makesitgo, line 188: in addition to avoiding capitalization/punctuation, avoid "code words" like variable or function names in the test descriptions as well

```suggestion
		t.Run("should return an error if fetching the client template for this template fails", func(t *testing.T) {
```

internal/commands/pull/inputs_test.go
```diff
@@ -159,3 +160,169 @@ func TestPullInputsResolveRemoteApp(t *testing.T) {
...
+
+			realmClient.CompatibleTemplatesFn = func(groupID, appID string) ([]realm.Template, bool, error) {
+				return []realm.Template{{ID: "some-template-id", Name: "should not be there"}}, true, nil
+			}
+			realmClient.ClientTemplateFn = func(groupID, appID, templateID string) (*zip.Reader, error) {
+				return nil, errors.New("something went wrong with client")
+			}
+			_, ui := mock.NewUI()
+			input := inputs{TemplateID: "some-template-id"}
+			result, err := input.resolveClient(ui, realmClient, "some-group-id", "some-app-id")
```
makesitgo, line 199: there should be no need to assert the `result` in an error case... which comes back to me asking if we can switch to the `if len(clientZipPkgs) == 0` in the pull command handler

internal/commands/pull/inputs_test.go
```diff
@@ -159,3 +160,169 @@ func TestPullInputsResolveRemoteApp(t *testing.T) {
...
+			assert.Nil(t, result)
+			assert.Equal(t, "something went wrong with client", err.Error())
```
makesitgo, line 201: assert equality on the `Error` rather than just the error string

internal/commands/pull/inputs_test.go
```diff
@@ -159,3 +160,169 @@ func TestPullInputsResolveRemoteApp(t *testing.T) {
...
+		t.Run("should return an error if the template ID is not compatible with the app", func(t *testing.T) {
+			var realmClient mock.RealmClient
+
+			realmClient.CompatibleTemplatesFn = func(groupID, appID string) ([]realm.Template, bool, error) {
+				return []realm.Template{{ID: "some-template-id", Name: "some template name"}}, true, nil
+			}
+
+			_, ui := mock.NewUI()
+			input := inputs{TemplateID: "wrong-template-id"}
+			result, err := input.resolveClient(ui, realmClient, "some-group-id", "some-app-id")
```
makesitgo, line 213: same thing about not needing to assert on `result` in an error case

internal/commands/pull/inputs_test.go
```diff
@@ -159,3 +160,169 @@ func TestPullInputsResolveRemoteApp(t *testing.T) {
...
+			assert.Nil(t, result)
+			assert.Equal(t, "template wrong-template-id is not compatible with this app", err.Error())
```
makesitgo, line 215: same thing on asserting equality on the `Error` rather than just the error string

internal/commands/pull/inputs_test.go
```diff
@@ -159,3 +160,169 @@ func TestPullInputsResolveRemoteApp(t *testing.T) {
...
+
+			_, ui := mock.NewUI()
+			input := inputs{TemplateID: "some-template-id"}
+			result, err := input.resolveClient(ui, realmClient, "some-group-id", "some-app")
+			assert.Nil(t, err)
+			assert.Equal(t, 1, len(result))
+			compareZipPackages(t, templateZipPkg1, result["some-template-id"])
+		})
+	})
+	t.Run("without a template ID passed in", func(t *testing.T) {
```
makesitgo, line 236:
```suggestion
	t.Run("without a template id passed in", func(t *testing.T) {
```

internal/commands/pull/inputs_test.go
```diff
@@ -159,3 +160,169 @@ func TestPullInputsResolveRemoteApp(t *testing.T) {
...
+		input := inputs{}
+		t.Run("should return nothing if the app is not made with a template", func(t *testing.T) {
+			var realmClient mock.RealmClient
+
+			realmClient.CompatibleTemplatesFn = func(groupID, appID string) ([]realm.Template, bool, error) {
+				return nil, false, nil
+			}
+			_, ui := mock.NewUI()
+			result, err := input.resolveClient(ui, realmClient, "some-group-id", "some-app-id")
+			assert.Nil(t, result)
```
makesitgo, line 246: assert on the `err` first in a success case, and then if we switch that `if` statement then assert on the `len(result)` here

internal/commands/pull/inputs_test.go
```diff
@@ -159,3 +160,169 @@ func TestPullInputsResolveRemoteApp(t *testing.T) {
...
+			assert.Nil(t, err)
+		})
+		t.Run("should return nothing if the user refuses to export with compatible templates", func(t *testing.T) {
```
makesitgo, line 249: [general style nit] add a newline between `t.Run` blocks (there are a few spots in this file, just do an overall check for this please)

internal/commands/pull/inputs_test.go
```diff
@@ -159,3 +160,169 @@ func TestPullInputsResolveRemoteApp(t *testing.T) {
...
+			var realmClient mock.RealmClient
+			realmClient.CompatibleTemplatesFn = func(groupID, appID string) ([]realm.Template, bool, error) {
+				return []realm.Template{{ID: "some-template-id", Name: "some name"}}, true, nil
+			}
+			_, console, _, ui, err := mock.NewVT10XConsole()
```
makesitgo, line 254: [nit] newline here to separate ui setup and client setup

internal/commands/pull/inputs_test.go
```diff
@@ -159,3 +160,169 @@ func TestPullInputsResolveRemoteApp(t *testing.T) {
...
+			doneCh := make(chan struct{})
+			go func() {
+				defer close(doneCh)
+				console.ExpectString("Would you like to export with a template?")
+				console.SendLine("n")
+				console.ExpectEOF()
+			}()
+
+			result, err := input.resolveClient(ui, realmClient, "some-group-id", "some-group-app")
+			assert.Nil(t, result)
```
makesitgo, line 267: same thing about:
  * asserting on the `err` first in a success case
  * assert on the `len(result)` rather than its nil-ness

internal/commands/pull/inputs_test.go
```diff
@@ -159,3 +160,169 @@ func TestPullInputsResolveRemoteApp(t *testing.T) {
...
+			assert.Nil(t, err)
+		})
+
+		t.Run("should prompt the user to export with compatible templates and return the templates selected", func(t *testing.T) {
```
makesitgo, line 271: it seems that this test (and potentially others) are already covered in some of your `pull_test` command Handler tests.  if that is the case, let's remove the duplicate ones from this file

internal/local/app.go
```diff
@@ -15,6 +15,13 @@ const (
 	maxDirectoryContainSearchDepth = 8
 )
 
+const (
+	// BackendPath is to specify the path for the app when we have templates
```
makesitgo, line 19:
```suggestion
	// BackendPath is the relative path to write app contents to when we have templates
```

internal/local/app.go
```diff
@@ -15,6 +15,13 @@ const (
...
+	BackendPath = "backend"
+	// FrontendPath is to specify the path for templates
```
makesitgo, line 21:
```suggestion
	// FrontendPath is the relative path to write frontend templates' contents to
```

makesitgo (commented):
a bunch of various thoughts/concerns/questions.  in general, i think the functionality mostly there, but we could do a lot to improve/focus on the UX of this.

i don't want to "Request changes" and throw a wrench in being able to get stuff out the door.  if you think (or the project needs) some of these things to be followed up on in separate tickets I'm ok with that.

let me know if you want to talk about any more of these in depth, happy to hop on a slack or call if need be

internal/cloud/realm/templates.go
```diff
@@ -65,3 +70,24 @@ func (c *client) ClientTemplate(groupID, appID, templateID string) (*zip.Reader,
 
 	return zipPkg, nil
 }
+
+func (c *client) CompatibleTemplates(groupID, appID string) ([]Template, bool, error) {
```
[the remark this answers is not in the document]
andrewzlchen: hmm i think both places are probably valid in some sense. I was in a hurry when I reviewed last time, so I misinterpreted your comment. This is on me.

for this endpoint, we return 400 if the current app was not created using a template app. The client template handler does this too.

Because 400's are "bad requests", it makes sense that we should stop the command and tell the user that their command is incorrect. We should halt wherever we encounter 400 errors.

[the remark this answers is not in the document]
makesitgo: then it would seem to me we want to error the whole command in that case, and instruct the user what went wrong, rather than proceed with a command to behave in the way they didn't specify

much like in the UI if we see any status code returned outside of `[200, 300)` we throw an error, I would expect the same here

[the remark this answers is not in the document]
jo-tse-mongo: I think that us returning a 400 from `baas` makes sense because we shouldn't be trying to get a ClientTemplate for an app that wasn't made by a template. But in the CLI, we don't know if the app was made with or without a template unless we ask `baas`, which is why we're using the 400 response code to check.  

I think if we had a call where we could check, like `isTemplateApp` or something that returns 200 with either true or false, then we can avoid making the ClientTemplate call check for it.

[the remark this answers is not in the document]
makesitgo: so just to clarify, I am less questioning the use of `400` for this response, but more so why are we swallowing that error here in the CLI?

if its a true error (and it sounds like it is from both descriptions here), then we should treat it as such.  provide feedback to the user about what went wrong: `"the app you requested was not created via a template..."` (you have a better idea on the messaging for something like that than i).  we can even go as far to provide suggested fixes: "remove the `--template` flag", "please use a templated app", etc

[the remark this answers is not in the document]
jo-tse-mongo: Yeah, I see. I think during some iterations ago, I might've used the `nil` return value for something, but looking at how the `resolveClientTemplates` evolved now, it might not be useful to have the `nil` value. 

I think it wouldn't be so bad in the current workflow to have the error being thrown, but @andrewzlchen let me know what you think.

internal/commands/pull/command.go
jo-tse-mongo: I think it's fine to leave it as is, since if there is no template, then there is no concept of `frontend` or `backend` paths

internal/commands/pull/command.go
jo-tse-mongo: 👍 on the dry run updating

`pathTarget` _is_ the path where the app is located. So if there's a template involved, we put the app in a `backend` folder

internal/commands/pull/command.go
makesitgo: feel free to settle on namings how you see fit, but what I don't love with the current implementation is that when templates are involved, you are mutating the value of `pathTarget`, and the order of operations matters significantly within that if block.  fwiw, i'm totally fine with us calling it "backend" even when a template isn't involved

that is why I want us to use 3 `var`s here for "app" (or could call it "root", as it effectively the same as `local.App.RootDir`), "backend" (which _is_ just the "app" when templates aren't involved, so maybe we call that the `pathApp`), and then "frontend" (which is where we write one or many client templates)

if/when we go down that route, we just need to be very careful we are then using the right versions of these 3 `var`s (and for what we use to derive `pathRelative`) to keep things working as expected

internal/commands/pull/command.go
makesitgo: my point being: we should be showing the user the top of the directory we've created for them with this `pathRelative`, not _just_ the backend directory.  then with the dry run updates, you can add further detail explaining what the `./backend` and`./frontend` directories would contain

let me know if you want to hop on a call or slack thread to discuss these things all at once, i think a lot of these comments will end up being related and might be helpful to tackle them all at once

internal/commands/pull/command_test.go
jo-tse-mongo: I removed it! I was going to use it to test something else, but I ended up putting it somewhere else

internal/commands/pull/command_test.go
jo-tse-mongo: The only difference when AutoConfirm is true is there is no extra prompt for this, but I'll add it to the test description
`Would you like to export with a template?`

internal/commands/pull/inputs.go
jo-tse-mongo: 👍 on the invert 

 I think this is where I was confused when I made the reply to your initial comment. For CompatibleTemplates, I made the `bool` return here to check whether the app was made with a template or not.

internal/commands/pull/inputs_test.go
jo-tse-mongo: I'm iffy on removing the tests even if they duplicate in the areas that they cover because the PullHandler tests are concerned with how `resolveClientTemplates` works in regards to the pull function, but these tests are for the different cases that _can_ happen in `resolveClientTemplates`.

internal/cloud/realm/templates.go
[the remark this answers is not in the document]
andrewzlchen: yeah throwing the error sounds good to me!

internal/commands/pull/inputs.go
andrewzlchen: `%q` works as well!

makesitgo: sort of an arbitrary distinction, but go with single quotes as that what most of our error messages are currently formatted with (when needed).  we use the double-quotes everywhere in help text.  i imagine at some point it will be normalized, but best to keep the different variants consistent within themselves at first

internal/commands/pull/inputs_test.go
makesitgo: definitely don't remove coverage, but in particular it seems to me this is specifically covered with the `"should export all selected compatible templates if no template id is passed in and the app is made with a template"` test (verifying the contents an all).  at the very least, we might be able to remove this one (and the `compareZipPkg` helper method along with it)

internal/cloud/realm/templates.go
```diff
@@ -43,25 +44,49 @@ func (c *client) Templates() ([]Template, error) {
...
+	return zipPkg, true, nil
+}
+
+func (c *client) CompatibleTemplates(groupID, appID string) ([]Template, error) {
+	res, resErr := c.do(http.MethodGet, fmt.Sprintf(compatibleTemplatesPathPattern, groupID, appID), api.RequestOptions{})
+	if resErr != nil {
+		return nil, resErr
+	}
+	if res.StatusCode == http.StatusBadRequest {
+		return nil, nil
```
makesitgo, line 80:
```suggestion
		return nil, nil // 400 implies app was not created with template so we return no compatible templates
```

internal/cloud/realm/templates.go
```diff
@@ -43,25 +44,49 @@ func (c *client) Templates() ([]Template, error) {
...
-func (c *client) ClientTemplate(groupID, appID, templateID string) (*zip.Reader, error) {
+func (c *client) ClientTemplate(groupID, appID, templateID string) (*zip.Reader, bool, error) {
 	res, resErr := c.do(http.MethodGet, fmt.Sprintf(clientTemplatePathPattern, groupID, appID, templateID), api.RequestOptions{})
 	if resErr != nil {
-		return nil, resErr
+		return nil, false, resErr
+	}
+	if res.StatusCode == http.StatusNoContent {
+		// No client exists for this template so there is nothing to return
+		return nil, true, nil
```
makesitgo, line 54: this should be `return nil, false, nil`

internal/commands/pull/command.go
```diff
@@ -60,32 +61,58 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
 	}
 
-	pathTarget, zipPkg, err := cmd.doExport(profile, clients.Realm, app.GroupID, app.ID)
+	clientZipPkgs, err := cmd.inputs.resolveClientTemplates(ui, clients.Realm, app.GroupID, app.ID)
 	if err != nil {
 		return err
 	}
 
-	pathRelative, err := filepath.Rel(profile.WorkingDirectory, pathTarget)
+	projectPath, zipPkg, err := cmd.doExport(profile, clients.Realm, app.GroupID, app.ID)
```
makesitgo, line 69: can we keep this in the same style/format as the other paths?
```suggestion
	pathProject, zipPkg, err := cmd.doExport(profile, clients.Realm, app.GroupID, app.ID)
```

internal/commands/pull/command.go
```diff
@@ -60,32 +61,58 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
 	if err != nil {
 		return err
 	}
 
-	proceed, err := checkAppDestination(ui, pathTarget)
+	var pathFrontend string
+	pathBackend := projectPath
+	if len(clientZipPkgs) != 0 {
+		pathFrontend = filepath.Join(projectPath, local.FrontendPath)
+		if proceed, err := checkPathDestination(ui, pathFrontend); err != nil {
```
makesitgo: i don't think we need this check here, instead let's check the `pathProject` down below on L#87`

internal/commands/pull/command.go
```diff
@@ -60,32 +61,58 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+			return err
+		} else if !proceed {
+			return nil
+		}
+		pathBackend = filepath.Join(projectPath, local.BackendPath)
+	}
+
+	// App path
+	proceed, err := checkPathDestination(ui, pathBackend)
```
makesitgo: let's use `pathProject` instead of `pathBackend` here.  this gives use one check/prompt to the user prior to the potential dry-run exit.  if the `pathProject` exists, chances are the `./backend` and `./frontend` paths already exist as well (unless they somehow switched their app from not using templates to using templates)

internal/commands/pull/command.go
```diff
@@ -60,32 +61,58 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+	}
+
 	if cmd.inputs.DryRun {
-		ui.Print(
-			terminal.NewTextLog("No changes were written to your file system"),
-			terminal.NewDebugLog("Contents would have been written to: %s", pathRelative),
-		)
+		if len(cmd.inputs.TemplateID) != 0 {
+			ui.Print(
+				terminal.NewTextLog("No changes were written to your file system"),
```
makesitgo, line 97: [DRY] we can build the list of logs to print (to ensure a consistent first message)

```go
logs := make([]terminal.Log, 0, 3)
logs = append(logs, terminal.NewTextLog("No changes were written to your file system"))

if len(clientZipPkgs) != 0 { // NOTE: I think its important we stick with the same check throughout here
  logs = append(logs,
    terminal.NewDebugLog("App contents would have been written to: %s", filepath.Join(pathRelative, local.BackendPath)),
    terminal.NewDebugLog("Template contents would have been written to: %s", filepath.Join(pathRelative, local.FrontendPath)),
  )
} else {
  logs = append(logs, terminal.NewDebugLog("Contents would have been written to: %s", pathRelative))
}
ui.Print(logs...)
return nil
```

internal/commands/pull/command.go
```diff
@@ -60,32 +61,58 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+				terminal.NewDebugLog("Contents would have been written to: %s", pathRelative),
+			)
+		}
 		return nil
 	}
 
-	if err := local.WriteZip(pathTarget, zipPkg); err != nil {
+	if err := local.WriteZip(pathBackend, zipPkg); err != nil {
```
makesitgo, line 110: if we wanted, right before this would be where I'd add `checkPathDestination` for the `pathBackend` filepath

internal/commands/pull/command.go
```diff
@@ -138,6 +165,14 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		ui.Print(terminal.NewDebugLog("Fetched hosting assets"))
 	}
 
+	for templateID, templateZipPkg := range clientZipPkgs {
```
makesitgo, line 163: if we wanted, right before this would be where I'd add `checkPathDestination` for the `pathFrontend` filepath

internal/commands/pull/command.go
```diff
@@ -60,32 +61,58 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+				terminal.NewTextLog("No changes were written to your file system"),
+				terminal.NewDebugLog("Contents would have been written to: %s", pathRelative),
+			)
+		}
 		return nil
 	}
 
-	if err := local.WriteZip(pathTarget, zipPkg); err != nil {
+	if err := local.WriteZip(pathBackend, zipPkg); err != nil {
 		return err
```
makesitgo, line 111:
```suggestion
		return fmt.Errorf("unable to write app to disk: %v", err)
```

internal/commands/pull/command.go
```diff
@@ -138,6 +165,14 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		ui.Print(terminal.NewDebugLog("Fetched hosting assets"))
 	}
 
+	for templateID, templateZipPkg := range clientZipPkgs {
+		if err := local.WriteZip(pathFrontend, templateZipPkg); err != nil {
```
makesitgo, line 164: I'd rather see us do this:

```go
if err := local.WriteZip(pathFrontend, templateZipPkg); err != nil {
  // note also the lack of starting capital letter, single quotes around the template id, and `%s` for the error string
  return fmt.Errorf("Unable to save template '%s' to disk: %s", templateID, err)
} else {
  // then also add a TODO here...
  // TODO(REALMC-9452): defer printing the successfully saved templates until after the `Successfully pulled app down' log
  // we should be able to print the list of saved templates in one log message/list, rather than each on their own new line
  ui.Print(terminal.NewTextLog("Saved template %s to disk", templateID))
}
```

internal/commands/pull/command_test.go
```diff
@@ -82,6 +87,28 @@ Contents would have been written to: app
...
+			backendDestination := filepath.Join(profile.WorkingDirectory, "app", local.BackendPath)
+			frontendDestination := filepath.Join(profile.WorkingDirectory, "app", local.FrontendPath)
+
+			assert.Equal(t, `No changes were written to your file system
+App contents would have been written to: app/backend
+Template contents would have been written to: app/frontend
+`, out.String())
+
+			_, err := os.Stat(backendDestination)
+			assert.True(t, os.IsNotExist(err), "expected %s to not exist, but instead: %s", err)
```
makesitgo, line 107: I think you're missing formatting args here

internal/commands/pull/command_test.go
```diff
@@ -82,6 +87,28 @@ Contents would have been written to: app
...
+			_, err = os.Stat(frontendDestination)
+			assert.True(t, os.IsNotExist(err), "expected %s to not exist, but instead: %s", err)
```
makesitgo, line 109: I think you're missing formatting args here

internal/commands/pull/command_test.go
```diff
@@ -317,6 +357,187 @@ Successfully pulled app down: app
...
+
+			_, err := os.Stat(destination)
+			assert.Nil(t, err)
+
+			testData, readErr := ioutil.ReadFile(filepath.Join(destination, fmt.Sprintf("%s.json", templateID)))
+			assert.Nil(t, readErr)
+			assert.Equal(t, expectedTemplate, string(testData))
+		})
+
+		t.Run("should export all selected compatible templates if no template id is passed in and the app is made with a template and autoConfirm is true", func(t *testing.T) {
```
makesitgo, line 418:
```suggestion
		t.Run("should export all selected compatible templates if no template id is passed in and the app is made with a template and auto confirm is not set", func(t *testing.T) {
```

internal/commands/pull/inputs.go
```diff
@@ -109,3 +117,64 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
...
 	return app, nil
 }
+
+func (i *inputs) resolveClientTemplates(ui terminal.UI, realmClient realm.Client, groupID, appID string) (map[string]*zip.Reader, error) {
+	compatibleTemplates, err := realmClient.CompatibleTemplates(groupID, appID)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(compatibleTemplates) == 0 {
```
makesitgo, line 127: not sure why this block moved, we should probably be keeping this _after_ the `if i.TemplateID != ""` block.

if they specified a template id but we got none back (likely due to the fact it was a `400` b/c the app wasn't created with a template), then we'll want to error, not proceed silently

makesitgo (commented):
im seeing the dreaded "Refresh" at the top of my screen... going to submit this review as just a "Comment" before going back for an "Approve"

internal/commands/app/create.go
```diff
@@ -335,16 +330,18 @@ func createFromTemplate(realmClient realm.Client, appID, groupID, templateID, ba
...
 		)
 		if err != nil {
 			return err
 		}
-
-		if err := local.WriteZip(path.Join(rootDir, frontendPath, templateID), zipPkg); err != nil {
+		if !ok {
+			return fmt.Errorf("template '%s' does not have a frontend to pull", templateID)
+		}
+		if err := local.WriteZip(path.Join(rootDir, local.FrontendPath), zipPkg); err != nil {
```
makesitgo, line 344: we still dropped `templateID` from this filepath (hopefully a test needs updating here too)

internal/commands/pull/command.go
```diff
@@ -60,33 +62,58 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
 	}
 
-	pathRelative, err := filepath.Rel(profile.WorkingDirectory, pathTarget)
+	pathProject, zipPkg, err := cmd.doExport(profile, clients.Realm, app.GroupID, app.ID)
 	if err != nil {
 		return err
 	}
 
-	proceed, err := checkAppDestination(ui, pathTarget)
+	var pathFrontend string
```
makesitgo, line 79: [meganit] we probably don't need to do this pathing logic until after the first `checkPathDestination` (maybe even after the `pathRelative` logic too, just before `if dryRun`)

internal/commands/pull/command.go
```diff
@@ -60,33 +62,58 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+
+		if len(clientZipPkgs) != 0 {
+			logs = append(logs,
+				terminal.NewDebugLog("App contents would have been written to: %s", filepath.Join(pathRelative, local.BackendPath)),
+				terminal.NewDebugLog("Template contents would have been written to: %s", filepath.Join(pathRelative, local.FrontendPath)),
+			)
+		} else {
+			logs = append(logs, terminal.NewDebugLog("Contents would have been written to: %s", pathRelative))
+		}
+		ui.Print(logs...)
```
makesitgo, line 111: [stylenit] can we add a newline b/w here and the closing bracket above

internal/commands/pull/command.go
```diff
@@ -138,6 +165,18 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		ui.Print(terminal.NewDebugLog("Fetched hosting assets"))
 	}
 
+	successfulTemplateWrites := make([]string, 0, len(clientZipPkgs))
+	for templateID, templateZipPkg := range clientZipPkgs {
```
makesitgo, line 169: you should add a `TODO(REALMC-9452)` here about `iterating over clientZipPkgs in a deterministic order`

internal/commands/pull/inputs.go
```diff
@@ -107,5 +115,72 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
...
+		return nil, nil
+	}
+
+	if proceed, err := ui.Confirm("Would you like to export with a template?"); err != nil {
+		return nil, err
+	} else if !proceed {
+		return nil, nil
+	}
+
+	nameOptions := make([]string, len(compatibleTemplates))
```
makesitgo, line 158: we should just file a followup ticket (and add the TODO here) but this behavior should probably align with `app create`, where instead of prompting the user above ("Would you like to export...") we otherwise just present them a list of templates, with `No template` (confirm the formatting on this) being the first option

makesitgo (commented):
last few things to get right (that `templateID` in the path is still nabbing us).

I also still want to say I'd like to remove the `pull/inputs` test where we're comparing zip packages.  if that's functionality not already covered in the `Handler` test then we should push it there (so we can just confirm file contents on disk instead), and then remove that test here

i promise you have my 👍  after this, but I just want to make sure we're covered on frontend paths...

internal/commands/pull/inputs.go
jo-tse-mongo: Added to this ticket 
https://jira.mongodb.org/browse/REALMC-9474

andrewzlchen:
bump

internal/commands/pull/command.go
```diff
@@ -138,6 +166,19 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+	successfulTemplateWrites := make([]string, 0, len(clientZipPkgs))
+	// TODO(REALMC-9452) fix to iterate in a deterministic order
+	for templateID, templateZipPkg := range clientZipPkgs {
+		if err := local.WriteZip(filepath.Join(pathFrontend, templateID), templateZipPkg); err != nil {
+			return fmt.Errorf("unable to save template '%s' to disk: %s", templateID, err)
+		}
+		// TODO(REALMC-9452): defer printing the successfully saved templates until after the `Successfully pulled app down' log
+		successfulTemplateWrites = append(successfulTemplateWrites, templateID)
+	}
+	if len(successfulTemplateWrites) != 0 {
```
makesitgo, line 178: is it easy enough to move this down below `"Successfully pulled app down"`?  could probably remove the above `TODO` if so

internal/commands/pull/inputs.go
```diff
@@ -107,5 +115,73 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
...
+	var selectedTemplateIdxs []int
+	if err := ui.AskOne(
+		&selectedTemplateIdxs,
+		&survey.MultiSelect{Message: "Which template(s) would you like to export this app with", Options: nameOptions}); err != nil {
+		return nil, err
+	}
+
+	templateIDs := make([]string, len(selectedTemplateIdxs))
+	for idx, selectedTemplateIdx := range selectedTemplateIdxs {
+		templateIDs[idx] = compatibleTemplates[selectedTemplateIdx].ID
```
makesitgo, line 173: [question] is the double `for` loop necessary here?  could we just create the `result` map inside of this `range selectedTemplateIdxs`?

makesitgo (approved):
just two final thoughts, both non-blocking/non-critical.  really nice work here

🆒  beans

PR #173, opened 2021-06-22
REALMC-9256: Implement programmatic flags for commands

makesitgo:
there be linter errors a-plenty here, but hoping to at least get some early eyes and thoughts on the approach here

what this does is standardize the way we write flag usage strings.  this is a result of having done quite a few deep-dives on the messaging with design, and there are patterns and formats we settled on that I'd like to live in code rather than in my head and/or comments.


tsedgwick (approved):
nice, LGTM

jandersongo (approved):

UnicodeSnowman (approved):

nathantfrank (approved):
N🧊 ! I'm liking these changes!

PR #174, opened 2021-06-23
REALMC-9228: re-enable user template prompt

tsedgwick:


makesitgo (commented):
looks fine, but I think we want to hold off on merging this until we cut the GA release (I've found one typo in a flag usage string so far...)

andrewzlchen:
I noticed something with the realm-cli app create command that I didn’t catch with the earlier PR. (seems to be a correctness issue)

The current code seems to create the template app but then immediately imports a default app which overwrites the template app

To fix it, i think we need to change [this line](https://github.com/10gen/realm-cli/blob/8a21e4ee57cdcf5fafbe2b3e7249cf05b77b0760/internal/commands/app/create.go#L191) to be

```go
if appRemote.GroupID == "" && appRemote.ID == "" && templateId == "" {
// ...
}
```

so that we can Export instead of instantiating a new app via `local.NewApp`

image

tsedgwick:
> I noticed something with the realm-cli app create command that I didn’t catch with the earlier PR. (seems to be a correctness issue)
> 
> The current code seems to create the template app but then immediately imports a default app which overwrites the template app
> 
> To fix it, i think we need to change [this line](https://github.com/10gen/realm-cli/blob/8a21e4ee57cdcf5fafbe2b3e7249cf05b77b0760/internal/commands/app/create.go#L191) to be
> 
> ```go
> if appRemote.GroupID == "" && appRemote.ID == "" && templateId == "" {
> // ...
> }
> ```
> 
> so that we can Export instead of instantiating a new app via `local.NewApp`
> 
> image

Is the template not handled below with the following?
```
	if cmd.inputs.Template != "" {
		s := spinner.New(terminal.SpinnerCircles, 250*time.Millisecond)
		s.Suffix = " Downloading client template..."

		downloadAndWriteClient := func() error {
			s.Start()
			defer s.Stop()

			zipPkg, err := clients.Realm.ClientTemplate(
				appRealm.GroupID,
				appRealm.ID,
				cmd.inputs.Template,
			)
			if err != nil {
				return err
			}

			if err := local.WriteZip(path.Join(rootDir, frontendPath), zipPkg); err != nil {
				return err
			}

			return nil
		}

		if err := downloadAndWriteClient(); err != nil {
			return err
		}
	}
```

andrewzlchen:
> Is the template not handled below with the following?

That's only for the frontend client portion. The frontend client stuff is fine - the way the command is handling the backend realm app is the issue atm

internal/commands/app/create.go
```diff
@@ -187,8 +187,12 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 	}
 
 	var appLocal local.App
-
-	if appRemote.GroupID == "" && appRemote.ID == "" {
+	if cmd.inputs.Template != "" {
```
makesitgo, line 190: i'm actually fine with it laid out this way as its reads nice/clear/direct, but I just want to call out `createFromTemplate` does a lot of the same things as the `else` block here does.  at the cost of making the `else if` here a bit more complex (`if appRemote.GroupID == "" && appRemote.ID == "" && cmd.inputs.Template == ""`, you could then leave the `else` block as is and add the `downloadAndWriteClient` usage inside a `if cmd.inputs.Template != ""` block

the only reason I suggest this is should the process of setting `appLocal` ever change (export => write => load) then we'll need to remember two spots in this command to update.  not a huge deal, i could just see us missing that at some point down the line.

[tangent] I want to wait and see how templates evolve before fully going down this path, but it seems to me that getting the `frontendDir`/`backendDir` management (with respect to `appLocal.RootDir`) to be owned/controlled by the `local.App` itself would make writing these commands a bit easier.  If we added something like `TemplateID` to the `local.App` struct, I think we could then remove a lot of the directory/path management we do for templates in `app create`/`push`/`pull`/`app init`...

internal/commands/app/create_test.go
```diff
@@ -504,23 +511,22 @@ Check out your app: cd ./remote-app && realm-cli app describe
...
-		backendZipPkg, err := zip.OpenReader("testdata/bitcoin-miner.zip")
+		backendZipPkg, err := zip.OpenReader("testdata/project.zip")
 		assert.Nil(t, err)
 
 		client := mock.RealmClient{
 			FindAppsFn: func(filter realm.AppFilter) ([]realm.App, error) {
 				return []realm.App{}, nil
 			},
 			ExportFn: func(groupID, appID string, req realm.ExportRequest) (string, *zip.Reader, error) {
-				return "", &backendZipPkg.Reader, err
```
makesitgo, line 523: 👀

internal/commands/app/new_app_inputs.go
```diff
@@ -94,29 +93,27 @@ func (i *newAppInputs) resolveTemplateID(ui terminal.UI, client realm.Client) er
...
-	//
-	//	i.Template = templateIDs[selectedIndex]
-	//
-	//	return nil
+	options := make([]string, 0, len(templates)+1)
+	templateIDs := make([]string, 0, len(templates)+1)
+	options = append(options, "[No Template]: Do Not Use A Template")
+	templateIDs = append(templateIDs, "")
+	for _, template := range templates {
+		options = append(options, fmt.Sprintf("[%s]: %s", template.ID, template.Name))
```
makesitgo, line 101: [optional/futureproof] might be nice to functionalize this format so we ensure consistency with the `[No Template]` option above

makesitgo (approved):
just a few thoughts/musings from me... i still think this should go in after #175 (and its subsequent GA release)

internal/commands/app/create.go
tsedgwick: yeah it does do a lot of the same thing. I found it made more sense with the least amount of refactoring to duplicate some of that code so all template logic was in the same place.

I'm open to doing a larger refactor later, but since we are pushing to release template starter apps I'm going to leave as is.

internal/commands/app/create_test.go
tsedgwick: this err _is_ harmless to return, but there's no need

internal/commands/app/new_app_inputs.go
tsedgwick: I'll add a ticket to do this and your suggestion above as a follow up refactor

tsedgwick (commented):
> just a few thoughts/musings from me... i still think this should go in after #175 (and its subsequent GA release)

I think we're going to want to have this for GA - if not a very follow fast to it

andrewzlchen (approved):
The looks good to me, but after building the binary, I'm running into one more issue.

I'm getting this error:

image

which seems to be the error message that originates from the import handler. from where we call the `import` [here](https://github.com/10gen/realm-cli/blob/318e02f2b24206b435b5f6a0b7972aa0952d867c/internal/commands/app/create.go#L261)

I don't think there's anything to do in the CLI, but we either need to allow URIs to be empty or somehow make a prevConfig in the import handler for template apps

internal/commands/app/create_test.go
makesitgo: just saying, named errors would never let such a thing occur :bushes:

andrewzlchen (changes requested):
oh, one thing we missed - we need to add tracking for template app creation, along with the template id that was created.

tsedgwick:
> oh, one thing we missed - we need to add tracking for template app creation, along with the template id that was created as its field.
> 
> https://docs.google.com/document/d/1gmbjUjeSULW4aI3OV9ytBqU0sypQ78H2cpIKIHr1xWM/edit?disco=AAAANLtngUo

@andrewzlchen That should have been in its own ticket for proper tracking. Ill do the work here, but let’s keep that in mind for the future please.

tsedgwick:
@makesitgo, @andrewzlchen I rebased into two commits, the first commit was already reviewed. The second commit contains all of the tracking logic.

andrewzlchen (approved):
🔥

andrewzlchen:
I don't have much context on the tracking bit, and since nick seems to be out until Wednesday, I'll add Jordan to the review.

jo-tse-mongo (approved):
LGTM - if there's time, it'd be nice to add some unit tests for the new telemetry stuff

PR #175, opened 2021-06-24
REALMC-8828: Update README with instructions to build and run CLI locally

makesitgo:


README.md
```diff
@@ -1,5 +1,82 @@
...
+
+You may wish to run the CLI from source, in which case `go` provides a few different options.
+
+#### Using `go run`
+
+The entry point for the CLI is in `main.go`, so simply running `go run main.go` from the root of the project repo is equivalent to invoking the installed command `realm-cli`.  It accepts commands and flags as normal.
+
+The one caveat here is you may wish to be running the tool from a clean directory, so as not to write app configuration files to the project repo.  However, for commands that do not interact with the local filesystem, this is often the quickest and easiest way to run from source.
+
+#### Using `go build`
```
UnicodeSnowman, line 21: potentially worth adding some instructions for `go install` here as well. I personally find it more convenient to use for local development vs. `go build`, since `go install` will automatically build the binary and put it in `~/go/bin`, which most people will presumably already have added to their path.

UnicodeSnowman (approved):

README.md
makesitgo: Yeah that’s a good call out. It crossed my mind initially but I’ve frankly yet to use it that way locally, so I didn’t have a confident/confirmed command to copy/paste (I’m hoping the -ldflags still works the same?) I’ll play around with it and get something for that though

tsedgwick (approved):

README.md
```diff
@@ -41,6 +41,16 @@ This will create a CLI build that will print `0.0.0-local` when `--version` is i
...
 
+#### Using `go install`
+
+Lastly, you can use `go install` which will build the `realm-cli` executable and install it in `$GOPATH/bin` (making it readily accessible throughout your machine).  To do so, run:
+
+```cmd
+go install main.go
+```
+
+This will create a CLI executable that is, by default, named `main`.  You should be able to run `which main` and/or `main --version` to confirm this installation was successful.  **You may wish to rename this binary to something more descriptive like `realm-cli` or `realm-cli-local`, to do so just `mv` the resulting file in `$GOPATH/bin`.**  Additionally, the same `-ldflags` apply for `go install` as they do `go build` should you wish to configure this build further.
```
UnicodeSnowman, line 52: hmm not quite. you can just do `go install` and it'll create a `realm-cli` binary in `~go/bin`, so you don't have to rename it

PR #177, opened 2021-06-30
REALMC-9391: Add initial data source to app create body

andrewzlchen:
Since we added a new field for the app creation endpoint in the `baas` PR of this ticket, we need to update the app creation request body that we send to baas from the realm-cli to match what we do in the UI.

This PR handles the above change and it adds a requirement that makes users specify a data source if they want to make a template app.

I believe prompting users for the data source that they want to use if they're requesting a template would be an improvement, but the current log should work for the time being.

### DIFF
--- a/internal/cloud/realm/app.go
+++ b/internal/cloud/realm/app.go
@@ -18,10 +18,11 @@ const (
 
 // AppMeta is Realm application metadata
 type AppMeta struct {
-	Location        Location        `json:"location,omitempty"`
-	DeploymentModel DeploymentModel `json:"deployment_model,omitempty"`
-	Environment     Environment     `json:"environment,omitempty"`
-	Template        string          `json:"template_id,omitempty"`
+	Location          Location        `json:"location,omitempty"`
+	DeploymentModel   DeploymentModel `json:"deployment_model,omitempty"`
+	Environment       Environment     `json:"environment,omitempty"`
+	Template          string          `json:"template_id,omitempty"`
+	InitialDataSource interface{}     `json:"link_data_source_data,omitempty"`
 }
 
 // App is a Realm application
--- a/internal/cloud/realm/app.go
+++ b/internal/cloud/realm/app.go
@@ -18,11 +18,11 @@ const (
 
 // AppMeta is Realm application metadata
 type AppMeta struct {
-	Location          Location        `json:"location,omitempty"`
-	DeploymentModel   DeploymentModel `json:"deployment_model,omitempty"`
-	Environment       Environment     `json:"environment,omitempty"`
-	Template          string          `json:"template_id,omitempty"`
-	InitialDataSource interface{}     `json:"link_data_source_data,omitempty"`
+	Location        Location        `json:"location,omitempty"`
+	DeploymentModel DeploymentModel `json:"deployment_model,omitempty"`
+	Environment     Environment     `json:"environment,omitempty"`
+	Template        string          `json:"template_id,omitempty"`
+	DataSource      interface{}     `json:"data_source,omitempty"`
 }
 
 // App is a Realm application
--- a/internal/commands/app/create.go
+++ b/internal/commands/app/create.go
@@ -183,14 +183,31 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 		return nil
 	}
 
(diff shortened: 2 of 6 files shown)


internal/commands/app/create_inputs.go
```diff
@@ -206,6 +208,7 @@ func (i *createInputs) resolveClusters(ui terminal.UI, client atlas.Client, grou
 					ReadPreference:      "primary",
 					WireProtocolEnabled: false,
 				},
+				Version: 1,
```
andrewzlchen, line 211: I needed to make this 1 in order to pass validation for the import handler.

I'm guessing that manually setting 1 here is not optimal. I'm having trouble finding context for what this version is, and how its being used.

Is there a better way of getting this?

tsedgwick: Setting to 1 is fine, afaik this only relates to using promises.
https://github.com/10gen/baas/blob/master/service/models/version.go

https://github.com/10gen/baas/blob/master/function/execution/vm/stream.go#L33

internal/commands/app/create.go
```diff
@@ -183,14 +183,31 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 		return nil
 	}
 
+	// choose a data source to import template app schema data onto
```
tsedgwick, line 186: defaulting to the first cluster isn't ideal for anyone with more than one datasource. can we just add the prompt to let the user select?

internal/cloud/realm/app.go
```diff
@@ -18,10 +18,11 @@ const (
...
 type AppMeta struct {
-	Location        Location        `json:"location,omitempty"`
-	DeploymentModel DeploymentModel `json:"deployment_model,omitempty"`
-	Environment     Environment     `json:"environment,omitempty"`
-	Template        string          `json:"template_id,omitempty"`
+	Location          Location        `json:"location,omitempty"`
+	DeploymentModel   DeploymentModel `json:"deployment_model,omitempty"`
+	Environment       Environment     `json:"environment,omitempty"`
+	Template          string          `json:"template_id,omitempty"`
+	InitialDataSource interface{}     `json:"link_data_source_data,omitempty"`
```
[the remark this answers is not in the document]
andrewzlchen: okay sounds good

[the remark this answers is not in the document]
tsedgwick: https://github.com/10gen/baas/pull/4454

internal/commands/app/create.go
andrewzlchen: added a prompt if user is trying to link multiple data sources

andrewzlchen:
The rest of the tests seem to be failing because cloud-dev seems to be down

internal/commands/app/create.go
andrewzlchen: hmm now that im looking at it again, I think maybe it's better to move this to the _inputs.go file. I will do this tonight

internal/commands/app/create_test.go
```diff
@@ -510,7 +519,27 @@ Check out your app: cd ./remote-app && realm-cli app describe
...
+		assert.Nil(t, consoleErr)
+		defer console.Close()
+
+		doneCh := make(chan (struct{}))
+		go func() {
+			defer close(doneCh)
+			procedure(console)
+		}()
+
+		// out, ui := mock.NewUI()
```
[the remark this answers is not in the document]
andrewzlchen: ack, wasn't sure at first so I just commented this out, but I forgot to come back to remove this.

I'll remove this for the next round

internal/commands/app/create_inputs.go
```diff
@@ -257,6 +261,45 @@ func (i *createInputs) resolveDatalakes(ui terminal.UI, client atlas.Client, gro
 	return dsDatalakes, nonExistingDatalakes, nil
 }
 
+func (i createInputs) resolveInitialTemplateDataSource(ui terminal.UI, dataLakes []dataSourceDatalake, clusters []dataSourceCluster) (interface{}, error) {
+	if len(dataLakes)+len(clusters) == 0 {
```
tsedgwick, line 265: this is a lot cleaner 👏🏻

tsedgwick (approved):

PR #178, opened 2021-06-30
REALMC-9363 - Removed client-side transpilation.

gagik:
1. Replaced `POST` endpoint with the new `PUT` endpoint.
2. Removed client-side transpilation.
3. Modified the tests to be compatible with the changes.


internal/cloud/realm/dependencies.go
```diff
@@ -49,20 +81,43 @@ func (c *client) ImportDependencies(groupID, appID, uploadPath string) error {
...
 		},
 	)
-	if resErr != nil {
-		return resErr
+	if err != nil {
+		return err
 	}
 	if res.StatusCode != http.StatusNoContent {
-		return api.ErrUnexpectedStatusCode{"import dependencies", res.StatusCode}
+		return fmt.Errorf("import got invalid status code %d", res.StatusCode)
```
gagik, line 96: I could split this into a different command-based function like `deployDraftAndWait(...)` (e.g. `importDependenciesAndWait` but just having this in the function itself allows to be a drop-in replacement.

internal/cloud/realm/dependencies_test.go
```diff
@@ -78,7 +85,8 @@ func TestRealmDependencies(t *testing.T) {
 			assert.Nil(t, expectedDepsErr)
 			defer expectedDeps.Close()
 
-			assert.Equalf(t, parseZipArchive(t, expectedDeps), parseZipArchive(t, actualDeps), "expected archives to match")
+			// make sure same files are present; contents may be different because of transpilation
```
gagik, line 88: not sure if comment is necessary

internal/local/dependencies.go
```diff
@@ -64,11 +62,6 @@ func (d Dependencies) PrepareUpload() (string, error) {
 		return "", err
 	}
 
-	transpiler, err := newDefaultTranspiler()
-	if err != nil {
-		return "", err
-	}
-
 	dir, err := ioutil.TempDir("", "") // uses os.TempDir and guarantees existence and proper permissions
```
gagik, line 65: not sure if other parts of the `PrepareUpload(...)` are important for the server-side transpilation. Should other bits be removed as well?

internal/local/dependencies_test.go
```diff
@@ -100,7 +99,7 @@ func TestDependenciesPrepare(t *testing.T) {
 	testUpload, testUploadErr := os.Open(filepath.Join(wd, "testdata/dependencies/upload.zip"))
```
gagik, line 99: it seems like this is meant to be the "transpiled" version? Not sure if there's a better way to rewrite the test to make it more meaningful. Again, depends on whether PrepareUpload() is even necessary.

internal/cloud/realm/dependencies.go
```diff
@@ -49,20 +82,43 @@ func (c *client) ImportDependencies(groupID, appID, uploadPath string) error {
...
+	}
+
+	waitForTranspilation := func() error {
```
gagik, line 100: I used a function for this to be consistent with `deployDraftAndWait` but could be removed into a basic loop in same scope.

internal/local/dependencies.go
arahmanan: I believe you can get rid of this and the `PrepareDependencies` functions completely. We no longer need to modify the archive that's provided by the user. All of this is done to transpile the js files within the provided archive, but now the new endpoint will handle that for us.

internal/local/dependencies_test.go
arahmanan: You can get rid of this test if you manage to remove the `PrepareUpload` function.

internal/cloud/realm/dependencies_test.go
arahmanan: I think it's fine to keep it

arahmanan (commented):
Good start! 💪

internal/cloud/realm/dependencies.go
```diff
@@ -18,10 +19,42 @@ const (
...
+func (c *client) GetDependenciesStatus(groupID, appID string) (DependenciesStatus, error) {
+	res, err := c.do(
+		http.MethodGet,
+		fmt.Sprintf(dependenciesStatusPathPattern, groupID, appID),
+		api.RequestOptions{},
+	)
+	if err != nil {
+		return "", err
+	}
+	var transpilation AppDependenciesInfo
```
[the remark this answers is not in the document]
gagik: oh yes this was left over from when I (incorreclty) had this as a transpile function

internal/cloud/realm/dependencies.go
```diff
@@ -18,10 +19,42 @@ const (
 	dependenciesPathPattern        = appPathPattern + "/dependencies"
 	dependenciesArchivePathPattern = dependenciesPathPattern + "/archive"
 	dependenciesDiffPathPattern    = dependenciesPathPattern + "/diff"
+	dependenciesStatusPathPattern  = dependenciesPathPattern + "/status"
 
 	paramFile = "file"
 )
 
+// AppDependenciesInfo is used to get information from a dependencies status request
```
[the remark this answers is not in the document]
gagik: sounds good; as in have it be the same name but different struct (aka one with only Status and StatusMessage as opposed with a `data` that has all the information inside)?

internal/cloud/realm/dependencies.go
gagik: for example, `DeployDraft` doesn't by itself wait for the draft to be deployed but `deployDraftAndWait` does. In this case `ImportDependencies` does always wait for the import to end. I was asking whether it'd be meaningful to have this separate as well. 
With regards to the error, my bad, I must have thought the api object was not of type `error`.

internal/cloud/realm/dependencies.go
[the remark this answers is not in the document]
arahmanan: yes, just the name with only `Status` and `StatusMessage`. The `data` field is not necessary. Also, it doesn't seem like you need to export this struct, so you can make it lowercased. I.e.:
```
type dependencyInstallation struct {
	Status DependenciesStatus `json:"status"`
        StatusMessage string `json:"status_message"`
}
```

internal/cloud/realm/dependencies.go
gagik: Yes I'll change all the mentions of transpilation. Moving out of the `ImportDependencies` is what I meant above when talking about `importDependenciesAndWait`.

internal/cloud/realm/dependencies.go
```diff
@@ -49,20 +82,43 @@ func (c *client) ImportDependencies(groupID, appID, uploadPath string) error {
...
+	if err != nil {
+		return err
 	}
 	if res.StatusCode != http.StatusNoContent {
-		return api.ErrUnexpectedStatusCode{"import dependencies", res.StatusCode}
+		return fmt.Errorf("import got invalid status code %d", res.StatusCode)
+	}
+
+	waitForTranspilation := func() error {
+		status, err := c.GetDependenciesStatus(groupID, appID)
```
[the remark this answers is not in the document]
gagik: I originally had this but was worried this assumption may not necessarily be always true in case something somehow goes wrong although I suppose even if it did this wouldn't really change the outcome.

internal/cloud/realm/dependencies.go
arahmanan: Ah got it. I don't have a preference here. The way you have it now works for me.

gagik:
Still WIP just want to see how it runs without `PrepareDependencies`.

arahmanan:
> Still WIP just want to see how it runs without `PrepareDependencies`.

btw you can run evergreen patches through [this command line tool](https://github.com/evergreen-ci/evergreen/wiki/Using-the-Command-Line-Tool) as well.

internal/cloud/realm/dependencies.go
```diff
@@ -18,10 +19,43 @@ const (
...
 	dependenciesDiffPathPattern    = dependenciesPathPattern + "/diff"
+	dependenciesStatusPathPattern  = dependenciesPathPattern + "/status"
 
 	paramFile = "file"
 )
 
+// DependencyInstallation is used to get information from a dependencies status request
+type DependencyInstallation struct {
+	Status        DependenciesStatus `json:"status"`
+	StatusMessage DependenciesStatus `json:"status_message"`
```
makesitgo, line 30: [question] why do we have this field, it does not seem to be used anywhere?  if we do need it, I'm guessing it wants to be of type `string` rather than `DependenciesStatus`

internal/cloud/realm/dependencies.go
```diff
@@ -18,10 +19,43 @@ const (
 	dependenciesPathPattern        = appPathPattern + "/dependencies"
 	dependenciesArchivePathPattern = dependenciesPathPattern + "/archive"
 	dependenciesDiffPathPattern    = dependenciesPathPattern + "/diff"
+	dependenciesStatusPathPattern  = dependenciesPathPattern + "/status"
 
 	paramFile = "file"
 )
 
+// DependencyInstallation is used to get information from a dependencies status request
+type DependencyInstallation struct {
```
makesitgo, line 28: [question] would it make sense to rename this to `DependenciesInstallation` to match the plural form of "dependencies" we're using everywhere else?  or is this "Status" really just for the installation of one dependency?  (if the latter is true, then maybe we need to rethink the name for `DependenciesStatus`)

internal/cloud/realm/dependencies.go
```diff
@@ -18,10 +19,43 @@ const (
...
+	Status        DependenciesStatus `json:"status"`
+	StatusMessage DependenciesStatus `json:"status_message"`
+}
+
+// DependenciesStatus represents possible dependency statuses
```
makesitgo, line 33:
```suggestion
// DependenciesStatus represents the status of dependencies being installed
```

internal/cloud/realm/dependencies.go
```diff
@@ -18,10 +19,43 @@ const (
...
+
+func (c *client) GetDependenciesStatus(groupID, appID string) (DependenciesStatus, error) {
+	res, err := c.do(
+		http.MethodGet,
+		fmt.Sprintf(dependenciesStatusPathPattern, groupID, appID),
+		api.RequestOptions{},
+	)
+	if err != nil {
+		return "", err
+	}
```
makesitgo, line 51: two other things we'll want to do here (as you'll see in other client method implementations):

1. check the `res.StatusCode` is what we expect it to be
  - return `api.ErrUnexpectedStatus{"get dependencies status", res.StatusCode}` if it doesnt match
2. after that, you can `defer res.Body.Close()`
  - b/c we're then consuming the response body, we want to make sure we close it

internal/cloud/realm/dependencies.go
```diff
@@ -49,20 +83,43 @@ func (c *client) ImportDependencies(groupID, appID, uploadPath string) error {
 		return err
 	}
 
-	res, resErr := c.do(
-		http.MethodPost,
+	res, err := c.do(
+		http.MethodPut,
```
makesitgo, line 87: just for my own sanity: i imagine there was a corresponding change for this on the backend?  do you mind linking me the commit?

internal/cloud/realm/dependencies.go
```diff
@@ -49,20 +83,43 @@ func (c *client) ImportDependencies(groupID, appID, uploadPath string) error {
...
+				return err
+			}
+			if status == DependenciesStatusFailed {
+				return fmt.Errorf("failed to install dependencies")
+			}
+		}
+		return nil
+	}
+
+	if err := waitForTranspilation(); err != nil {
```
makesitgo, line 119: i may not be fully aware of the requirements on this, but I don't think this "Wait for" should live in the implementation of realm client's `ImportDependencies`.  i get the desire to make sure we're always waiting for the successful install (and report failures otherwise) whenever we call `ImportDependencies`, but so far in the CLI that operation is only called once (in the `push` command).

this logic can/should just live there for now.  if/when we ever need this to be abstracted away, it would then make sense to make a helper (similar to `cli.ResolveApp`) which encapsulated all of this.

the main reason this caught my eye is something andy already alluded to: using a spinner.  moreover, we should avoid placing any "ui logic" (or related package imports) in the `cloud/realm` package

internal/cloud/realm/dependencies.go
```diff
@@ -49,20 +82,43 @@ func (c *client) ImportDependencies(groupID, appID, uploadPath string) error {
...
+		return fmt.Errorf("import got invalid status code %d", res.StatusCode)
+	}
+
+	waitForTranspilation := func() error {
+		status, err := c.GetDependenciesStatus(groupID, appID)
+		if err != nil {
+			return err
+		}
+		for status == DependenciesStatusCreated {
+			time.Sleep(time.Second)
```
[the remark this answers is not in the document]
makesitgo: fwiw, we use 1 second when waiting for the deployment to complete.  I think that's fine here too, but either way let's keep it consistent.  (3 seconds feels like _maybe_ a bit long of a wait for a CLI command)

internal/cloud/realm/dependencies_test.go
```diff
@@ -28,8 +28,15 @@ func TestRealmDependencies(t *testing.T) {
 		assert.Nil(t, err)
 
 		uploadPath := filepath.Join(wd, "testdata/dependencies_upload.zip")
+
+		_, err = client.GetDependenciesStatus(groupID, app.ID)
+		assert.NotNilf(t, err, "dependency status must return error before import")
```
makesitgo, line 33: can we assert the equality on the actual error itself?

internal/cloud/realm/dependencies_test.go
```diff
@@ -28,8 +28,15 @@ func TestRealmDependencies(t *testing.T) {
...
 		assert.Nil(t, client.ImportDependencies(groupID, app.ID, uploadPath))
 
+		status, err := client.GetDependenciesStatus(groupID, app.ID)
+		assert.Equalf(t, status, realm.DependenciesStatusSuccessful, "must wait until dependency status is successful during import")
+		assert.Nil(t, err)
```
makesitgo, line 38: should always assert/test the `err` first (before `status)

internal/cloud/realm/dependencies_test.go
```diff
@@ -28,8 +28,15 @@ func TestRealmDependencies(t *testing.T) {
 		assert.Nil(t, err)
 
 		uploadPath := filepath.Join(wd, "testdata/dependencies_upload.zip")
+
+		_, err = client.GetDependenciesStatus(groupID, app.ID)
+		assert.NotNilf(t, err, "dependency status must return error before import")
 		assert.Nil(t, client.ImportDependencies(groupID, app.ID, uploadPath))
 
+		status, err := client.GetDependenciesStatus(groupID, app.ID)
+		assert.Equalf(t, status, realm.DependenciesStatusSuccessful, "must wait until dependency status is successful during import")
```
makesitgo, line 37: [optional] I don't think the extra message here adds much, feel free to just use `assert.Equal` here

internal/cloud/realm/dependencies_test.go
```diff
@@ -97,7 +105,7 @@ func TestRealmDependencies(t *testing.T) {
 	})
 }
 
-func parseZipArchive(t *testing.T, file *os.File) map[string]string {
+func getZipFileNames(t *testing.T, file *os.File) map[string]bool {
```
makesitgo, line 108: [meganit] use return value of `map[string]struct{}` instead (it may be _slightly_ more performant, but it also conveys there's no use for the values of the map here, we're simply keeping track of the key set)

internal/cloud/realm/dependencies_test.go
```diff
@@ -106,5 +114,13 @@ func parseZipArchive(t *testing.T, file *os.File) map[string]string {
 	zipPkg, err := zip.NewReader(file, fileInfo.Size())
 	assert.Nil(t, err)
 
-	return parseZipPkg(t, zipPkg)
+	fileNames := make(map[string]bool, len(zipPkg.File))
```
makesitgo, line 117: [naming nit] `filename`/`filenames` would be the preferred form of this `var` name within this codebase (something more general like `out` is totally fine within the scope of this helper func)

internal/commands/app/diff.go
```diff
@@ -97,10 +97,12 @@ func (cmd *CommandDiff) Handler(profile *user.Profile, ui terminal.UI, clients c
 	}
 
 	if cmd.inputs.IncludeDependencies {
-		uploadPath, err := local.PrepareDependencies(app, ui)
+		uploadDependencies, err := local.FindAppDependencies(app.RootDir)
```
makesitgo, line 100: [naming nit] maybe we just call this `localDependencies` or `appDependencies` now

internal/commands/app/diff.go
```diff
@@ -97,10 +97,12 @@ func (cmd *CommandDiff) Handler(profile *user.Profile, ui terminal.UI, clients c
...
 		if err != nil {
 			return err
 		}
+		uploadPath := uploadDependencies.ArchivePath
```
makesitgo, line 104: I don't think we need to declare this as its own `var` ... also I don't think we want to be calling `defer os.Remove(uploadPath)` down below anymore ... that was originally for cleanup of the temporary `.zip` we'd create for upload when calling `PrepareDependencies`.  this would now just remove the user's original dependencies each time

internal/commands/push/command.go
```diff
@@ -100,16 +100,17 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		return err
 	}
 
-	var uploadPathDependencies string
+	var uploadDependenciesPath string
```
makesitgo, line 103: [question] did this need to change?  no need to revert if not, just going forward its generally advisable to limit the amount of drive-bys in a PR to keep the diff focused and easier to read.

some instances where a drive-by may be worth it: if its fixing an issue or otherwise undoing a pattern we're trying to eliminate from a codebase

internal/commands/push/command.go
```diff
@@ -100,16 +100,17 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
 	var dependenciesDiffs realm.DependenciesDiff
 	if cmd.inputs.IncludeDependencies {
-		uploadPathDependencies, err = local.PrepareDependencies(app, ui)
+		uploadDependencies, err := local.FindAppDependencies(app.RootDir)
```
makesitgo, line 106: same name nit here about using `localDependencies` or `appDependencies` here

internal/commands/push/command.go
```diff
@@ -100,16 +100,17 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
 		if err != nil {
 			return err
 		}
-		defer os.Remove(uploadPathDependencies) //nolint:errcheck
+		uploadDependenciesPath := uploadDependencies.ArchivePath
```
makesitgo, line 110: i actually think there's a subtle bug here... you can see it easier [here](https://play.golang.org/p/INJtTpTqdIg) but in short, b/c you have `:=` here you are no longer assigning to the string `var` we declared outside this if block

internal/commands/push/command.go
```diff
@@ -100,16 +100,17 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+		defer os.Remove(uploadDependenciesPath) //nolint:errcheck
```
makesitgo, line 111: I may be missing something still with the new way we're handling dependencies, but I don't think we want/need this `defer` anymore.  these `os.Remove` calls were always to ensure the CLI cleaned up after itself while creating new files from the `PrepareDependencies` implementation

makesitgo (changes requested):
I think you'll need to make sure we're updating tests for `app diff` and `push` here as well

let me know if you have any questions/thoughts about my comments here, would be happy to discuss anything further.

i am incredibly excited to see we can remove all things transpiler-related from this codebase though!  nice work so far

internal/cloud/realm/dependencies.go
gagik: that's true it's meant to be a string and will be used in the spinner based on earlier feedback from Andy, was still working on it so this was more of in an intermediary stage.

internal/cloud/realm/dependencies.go
gagik: seems like it'd represent all dependencies given so I'll make sure plural is consistent.

internal/cloud/realm/dependencies.go
gagik: Yes, [this is the commit that adds the new endpoint with job queues](https://github.com/10gen/baas/commit/2a22b089090c18168a841d396d1c240d03c2df5f),.

internal/cloud/realm/dependencies_test.go
gagik: Yes although the error message essentially comes from the `baas` server so if this will have to be somewhat hard-coded and potentially break if it changes on the server side.

internal/commands/push/command.go
gagik: I just found it to be confusing to be called `PathDependencies` as opposed to `DependenciesPath` and worthy of changing to reflect a causal connection with the `uploadDependencies` variable I declared below. Will take the note about drive by-s in mind in future PRs, thank you! I think it will be good to change this to `appDependenciesPath`, again to be consistent with the other suggested variable change.

internal/cloud/realm/dependencies_test.go
makesitgo: you're right in pointing out a brittle test, but in this case I think it's fine to have.  for one, these tests have been this repo's main line of defense when it comes to server-side changes that unknowingly have an impact on the CLI.  obviously for this error message, its probably not critical to the functionality of the CLI, however its a good practice to get into.

internal/commands/push/command.go
makesitgo: naming in go can get weird at times (as with all languages).  you'll notice we use `uploadPath` around the codebase, and the intent here was to specify/denote the flavor of `uploadPath` we were dealing with, in this case a dependencies one.

either way, feel free to leave as you have it now!  just wanted to chime in about drive-bys in general here

internal/cloud/realm/dependencies_test.go
```diff
@@ -97,7 +113,23 @@ func TestRealmDependencies(t *testing.T) {
 	})
 }
 
-func parseZipArchive(t *testing.T, file *os.File) map[string]string {
+func waitForInstallation(realmClient realm.Client, groupID, appID string) error {
```
gagik, line 116: For testing, this seemed like the best way to go about this to keep `cloud/realm` and `command` parts of the code separate. Not sure if there might be a better approach

internal/commands/push/command.go
```diff
@@ -444,3 +444,40 @@ func deployDraftAndWait(ui terminal.UI, realmClient realm.Client, remote appRemo
...
+		defer s.Stop()
+
+		status := realm.DependenciesStatusCreated
+		for status == realm.DependenciesStatusCreated {
+			time.Sleep(time.Second)
+			install, err := realmClient.DependenciesInstallation(remote.GroupID, remote.AppID)
+			if err != nil {
+				return err
+			}
+			s.Suffix = fmt.Sprintf(" %s...", install.StatusMessage)
```
gagik, line 468: At the moment, StatusMessage just ends up being `processing` so the spinner that displays is like `[] processing...`. I'm not sure if this is intentional but does not seem user friendly since other spinners have messages like `[] Installing dependencies...` so perhaps capitalization / more descriptive message would be nice to have?

internal/commands/push/command.go
```diff
@@ -444,3 +444,40 @@ func deployDraftAndWait(ui terminal.UI, realmClient realm.Client, remote appRemo
...
+			status = install.Status
+			if status == realm.DependenciesStatusFailed {
+				return fmt.Errorf("failed to install dependencies")
```
gagik, line 471: should the error be handled differently? any logging necessary?

internal/commands/push/command.go
```diff
@@ -444,3 +444,40 @@ func deployDraftAndWait(ui terminal.UI, realmClient realm.Client, remote appRemo
...
+			}
+		}
+		return nil
+	}
+
+	if err := waitForInstallation(); err != nil {
+		return err
+	}
+
+	ui.Print(terminal.NewTextLog("Dependencies installation complete"))
```
gagik, line 481: is this a good message to end with?

internal/commands/push/command_test.go
```diff
@@ -1429,6 +1431,89 @@ func TestPushCommandDeployDraftAndWait(t *testing.T) {
 	})
 }
 
+func TestPushCommandImportDependenciesAndWait(t *testing.T) {
+	groupID, appID, uploadPath := "groupID", "appID", "node_modules.zip"
+	testError := errors.New("something bad happened")
```
gagik, line 1436: other tests just kept re-introducing this variable which I thought was a bit, I'm not sure if this is better and can just do it like the rest if that's better for readability.

internal/commands/push/command_test.go
```diff
@@ -1429,6 +1431,89 @@ func TestPushCommandDeployDraftAndWait(t *testing.T) {
...
+		realmClient.ImportDependenciesFn = func(groupID, appID, uploadPath string) error {
+			capturedGroupID = groupID
+			capturedAppID = appID
+			capturedUploadPath = uploadPath
+			return testError
+		}
+		err := importDependenciesAndWait(nil, realmClient, appRemote{groupID, appID}, uploadPath)
+		assert.Equal(t, testError, err)
+
+		t.Log("and should properly pass through the expected inputs")
```
gagik, line 1451: I could do this as another nested `t.Run` instead of a log if that's better.

gagik (commented):
Just some things I am not so sure about.

internal/commands/push/command.go
```diff
@@ -444,3 +444,40 @@ func deployDraftAndWait(ui terminal.UI, realmClient realm.Client, remote appRemo
...
+		status := realm.DependenciesStatusCreated
+		for status == realm.DependenciesStatusCreated {
+			time.Sleep(time.Second)
+			install, err := realmClient.DependenciesInstallation(remote.GroupID, remote.AppID)
+			if err != nil {
+				return err
+			}
+			s.Suffix = fmt.Sprintf(" %s...", install.StatusMessage)
+			status = install.Status
+			if status == realm.DependenciesStatusFailed {
```
arahmanan, line 470: [opt] you could move this if statement outside of the for loop.

internal/commands/push/command_test.go
arahmanan: I don't have a preference here. I'm ok with even removing this log.

internal/commands/push/command_test.go
arahmanan: this works for me.

internal/cloud/realm/dependencies_test.go
```diff
@@ -97,7 +113,23 @@ func TestRealmDependencies(t *testing.T) {
...
+	status := realm.DependenciesStatusCreated
+	for status == realm.DependenciesStatusCreated {
```
makesitgo, line 118: we should have a mechanism here to prevent a test from potentially hanging... instead of using a `status` var to check conditions, we could switch this loop to be `for idx < 100` (or w/e you think an appropriate limit is), and then each time check `status`, breaking out or returning early from the loop as necessary

note: this is less of a concern for the actual implementation of this code, since users can/should use ctrl+c to kill any potentially hanging commands

internal/commands/push/command.go
```diff
@@ -444,3 +444,40 @@ func deployDraftAndWait(ui terminal.UI, realmClient realm.Client, remote appRemo
 	ui.Print(terminal.NewTextLog("Deployment complete"))
 	return nil
 }
+
+func importDependenciesAndWait(ui terminal.UI, realmClient realm.Client, remote appRemote, uploadPath string) error {
+	err := realmClient.ImportDependencies(remote.GroupID, remote.AppID, uploadPath)
+	if err != nil {
```
makesitgo, line 450: [nit] can reduce the scope footprint of `err` by inlining the `if` statement here

also maybe it makes sense to include all of this within the "spinner scope", thoughts?

internal/commands/push/command.go
```diff
@@ -444,3 +444,40 @@ func deployDraftAndWait(ui terminal.UI, realmClient realm.Client, remote appRemo
...
+	s.Suffix = " Installing dependencies..."
+
+	waitForInstallation := func() error {
+		s.Start()
+		defer s.Stop()
+
+		status := realm.DependenciesStatusCreated
+		for status == realm.DependenciesStatusCreated {
+			time.Sleep(time.Second)
+			install, err := realmClient.DependenciesInstallation(remote.GroupID, remote.AppID)
```
makesitgo, line 464: same `var` naming comment here

internal/commands/push/command.go
makesitgo: alternatively could just use `"Installed dependencies"` (it sort of matches the rough pattern with the other spinners we've used in for importing hosting assets, resetting the cdn cache, etc)

internal/commands/push/command.go
```diff
@@ -444,3 +444,40 @@ func deployDraftAndWait(ui terminal.UI, realmClient realm.Client, remote appRemo
 	ui.Print(terminal.NewTextLog("Deployment complete"))
 	return nil
 }
+
+func importDependenciesAndWait(ui terminal.UI, realmClient realm.Client, remote appRemote, uploadPath string) error {
```
makesitgo, line 448: does this need to be its own function?

internal/commands/push/command_test.go
makesitgo: personally not a huge fan of this, but its not a big deal either way.  in my mind, its just one more hop when reading test code to figure out what exactly is being asserted... in generally DRYing up test code often leads to just more complexity and difficulty to read

internal/commands/push/command_test.go
```diff
@@ -1429,6 +1431,89 @@ func TestPushCommandDeployDraftAndWait(t *testing.T) {
...
+			return nil
+		}
+
+		t.Run("but fails to finish the installation", func(t *testing.T) {
+			realmClient.DependenciesInstallationFn = func(groupID, appID string) (realm.DependenciesInstallation, error) {
+				return realm.DependenciesInstallation{}, testError
+			}
+
+			out, ui := mock.NewUI()
+			err := importDependenciesAndWait(ui, realmClient, appRemote{groupID, appID}, uploadPath)
```
makesitgo, line 1469: I imagine you introduced the `importDependenciesAndWait` helper for testing purposes, but this should be something thats just as easily testable with `cmd.Handler(...)` (if we choose to get rid of the helper all together)

internal/commands/push/command_test.go
```diff
@@ -1429,6 +1431,89 @@ func TestPushCommandDeployDraftAndWait(t *testing.T) {
...
+			t.Log("and should fail with error")
```
makesitgo, line 1470: [test style nit] I would get rid of the `t.Log`s and instead just add these to the encompassing test descriptions

internal/commands/push/command_test.go
```diff
@@ -1429,6 +1431,89 @@ func TestPushCommandDeployDraftAndWait(t *testing.T) {
...
+			assert.Equal(t, err, testError)
+			assert.Equal(t, out.String(), "")
```
makesitgo, line 1472: again not super critical to test the empty values on error/failure tests, can just remove this assert

internal/commands/push/command_test.go
```diff
@@ -1429,6 +1431,89 @@ func TestPushCommandDeployDraftAndWait(t *testing.T) {
...
+				polls++
+				return realm.DependenciesInstallation{Status: status, StatusMessage: "processing"}, nil
+			}
+
+			out, ui := mock.NewUI()
+
+			t.Log("and should fail with installation error")
+			err := importDependenciesAndWait(ui, realmClient, appRemote{groupID, appID}, uploadPath)
+			assert.Equal(t, err, errors.New("failed to install dependencies"))
+			assert.Equal(t, out.String(), "")
```
makesitgo, line 1512: similar note about not needing to test the empty case in a failure scenario

internal/commands/push/command_test.go
```diff
@@ -1429,6 +1431,89 @@ func TestPushCommandDeployDraftAndWait(t *testing.T) {
...
+				}
+				polls++
+				return realm.DependenciesInstallation{Status: status, StatusMessage: "processing"}, nil
+			}
+
+			out, ui := mock.NewUI()
+
+			t.Log("and should fail with installation error")
+			err := importDependenciesAndWait(ui, realmClient, appRemote{groupID, appID}, uploadPath)
+			assert.Equal(t, err, errors.New("failed to install dependencies"))
```
makesitgo, line 1511: also I'm just noticing this now, but with the test code here we actually want `assert.Equal(t, expected, actual)`

do you mind going through all your tests and here ensuring we use that order?  it just aligns better with some of the auto-generated test failure messages this way

makesitgo (commented):
lookin good here, just have another batch of comments/questions for you.  really nice turnaround on a lot of this stuff!

internal/cloud/realm/client.go
```diff
@@ -30,6 +30,7 @@ type Client interface {
 	ImportDependencies(groupID, appID, uploadPath string) error
 	Diff(groupID, appID string, appData interface{}) ([]string, error)
 	DiffDependencies(groupID, appID, uploadPath string) (DependenciesDiff, error)
+	DependenciesInstallation(groupID, appID string) (DependenciesInstallation, error)
```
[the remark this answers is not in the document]
gagik: what about `DependenciesImportStatus` or `DependenciesInstallStatus`? `DependenciesImportStatus` seems to make most sense to me considering this is only being used for `ImportDependencies` right now. @arahmanan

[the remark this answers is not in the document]
makesitgo: my one caution here would be that we shouldn't name the client method based on how its used, we should name it for what it does.

I finally got up to speed took a look over on the `baas` side... I would vote one of the following:
* keep it as `DependenciesInstallation` ... I didn't realize that's how we referred to the `struct` in the backend, I can figure out how to reason what an installation means eventually
* use `DependenciesStatus` since that's what the API handler is called (`getDependenciesStatus`)

I think ultimately I like the latter, although then it introduces a stutter with the `Status` field (which could ultimately renamed to `State` or `Type` while keeping the `json` mapping the same `"status"`)

---

at the very least, I think this encourages us to remove the `type DependencyStatus string` as its overloading more of these words, and really doesn't add anything that a simple `string` couldn't do

install.js
```diff
@@ -105,17 +101,4 @@ try {
...
-} catch (err) {
-  console.error('Realm CLI installation failed:', err);
-  process.exit(1);
-}
-
-requstBinary(transpilerdownloadURL, 'transpiler').catch(err => {
-  console.error('failed to download Realm CLI:', err);
-  process.exit(1);
-});
+});
```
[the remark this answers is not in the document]
gagik: should I remove the new line?

internal/cloud/realm/client.go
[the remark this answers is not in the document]
gagik: on a different note I guess simply naming it `DependenciesStatus` makes sense considering it's a request to the `dependencies/status` endpoint so seems most fitting for that.

internal/commands/push/command_test.go
gagik: I took the same approach / structure as the `deployDraftAndWait` function [here to stay consistent.](https://github.com/10gen/realm-cli/blob/master/internal/commands/push/command.go#:~:text=func-,deployDraftAndWait,-(ui%20terminal.UI). I did introduce the [waitForInstallation function for testing](https://github.com/gagik/realm-cli/commit/11961946d9ab94177a0f5c31856630764840d823#:~:text=func-,waitForInstallation,-(realmClient%20realm.Client) which I will remove since it's only being used once.

makesitgo: so we have patterns in the `if cmd.inputs.IncludeHosting` and `if cmd.inputs.ResetCDNCache` blocks as well that I think equally apply, at the very least I want to make sure we print logs in a similar manner/style to those blocks

if you find this extra function dramatically helps testing then I'm totally fine leaving it as is (I'll be the first one to tell you I was a little "helper function crazy" earlier on in this codebase 😬 )

gagik: ah okay, did not notice how `cmd.Handler(...)` works, will give this a try.

internal/commands/push/command.go
```diff
@@ -198,7 +198,35 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		if err := clients.Realm.ImportDependencies(appRemote.GroupID, appRemote.AppID, uploadPathDependencies); err != nil {
 			return err
 		}
-		ui.Print(terminal.NewTextLog("Uploaded dependencies archive"))
+
+		ui.Print(terminal.NewTextLog("Uploaded dependencies"))
```
gagik, line 202: Is this message necessary? If so, I thought it would be good to make it more consistent with the "Installed dependencies" message that should occur after it.

internal/cloud/realm/dependencies_test.go
```diff
@@ -28,8 +28,32 @@ func TestRealmDependencies(t *testing.T) {
...
+		startStatus, err := client.DependenciesStatus(groupID, app.ID)
+		assert.Nil(t, err)
+		assert.Equal(t, realm.DependenciesStatus{
+			State:   realm.DependenciesStateCreated,
+			Message: "processing request",
+		}, startStatus)
+
+		// wait for dependencies to finish installation
+		var currentStatus realm.DependenciesStatus
+		for i := 0; i < 50; i++ {
```
gagik, line 46: in my computer it usually takes around <15 runs but I imagine with all sorts of delays and all, `50` seems to be a good balance between making sure it isn't flaky but also not running for too long; could be lowered.

internal/cloud/realm/dependencies.go
```diff
@@ -18,10 +18,44 @@ const (
...
 	dependenciesArchivePathPattern = dependenciesPathPattern + "/archive"
 	dependenciesDiffPathPattern    = dependenciesPathPattern + "/diff"
+	dependenciesStatusPathPattern  = dependenciesPathPattern + "/status"
 
 	paramFile = "file"
 )
 
+// DependenciesStatus is used to get information from a dependencies status request
+type DependenciesStatus struct {
+	State   string `json:"status"`
```
gagik, line 28: I think this may be the best way to avoid stutter but I can see how the idea of a State of Status is a bit odd.

internal/cloud/realm/dependencies.go
```diff
@@ -18,10 +18,44 @@ const (
...
+	Message string `json:"status_message"`
+}
+
+// set of known dependencies status states
+const (
+	DependenciesStateCreated    = "created"
```
gagik, line 34: I updated these to make them more related to the `status.State` as opposed to the `status` itself.

gagik (commented):
Hopefully this addresses many of the issues so far.

.evg.yml
```diff
@@ -190,8 +169,6 @@ tasks:
     exec_timeout_secs: 3600
     commands:
       - func: "fetch_go114"
-      - func: "fetch_yarn"
-      - func: "fetch_baas_artifacts"
```
makesitgo, line 194: `test_with_cloud` task will still want the `fetch_baas_artifacts` step included... inside of `setup_baas_server` we have this command:

```cmd
cp baas_artifacts/transpiler baas/etc/transpiler/bin/
```

internal/cloud/realm/dependencies_test.go
```diff
@@ -28,8 +28,32 @@ func TestRealmDependencies(t *testing.T) {
...
+			time.Sleep(time.Second)
```
makesitgo, line 47: [minor nit] maybe move this `Sleep` to the end of the for loop

internal/commands/push/command.go
makesitgo: It does seem like it can go away now... it previously existed to mark the completion of this block, which as you allude to is now done with the "Installed dependencies" message below

I think the key here is we want to make sure this behaves similarly (read: prints the same amount of logs) as the hosting and/or reset cdn blocks do below (which from the looks of it appears to be the case)

Have you tested this locally yet?  I'm guessing there's a test case below that can also confirm this...

internal/commands/push/command.go
```diff
@@ -198,7 +198,35 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+		s := spinner.New(terminal.SpinnerCircles, 250*time.Millisecond)
+		s.Suffix = " Installing dependencies: starting..."
+
+		s.Start()
```
makesitgo, line 206: these start/stop calls should be inside the `waitForInstallation` function scope (to ensure the `defer` behaves correctly)

internal/commands/push/command.go
```diff
@@ -198,7 +198,35 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+		defer s.Stop()
+		waitForInstallation := func() error {
```
makesitgo, line 208: [naming nit] lets rename this to `installDependencies`/`importDependencies`/something along those lines

also, I think I mentioned this earlier but we should include the `clients.Realm.ImportDependencies(...)` call inside this function as well.  this way, we have the spinner present (and only present) for the duration of these dependencies actions)

internal/commands/push/command_test.go
```diff
@@ -581,12 +618,12 @@ Successfully pushed app up: eggcorn-abcde
 
 			assert.Nil(t, cmd.Handler(nil, ui, cli.Clients{Realm: realmClient}))
 			assert.Equal(t, `Determining changes
-Transpiled dependency sources
 Creating draft
 Pushing changes
 Deploying draft
 Deployment complete
-Uploaded dependencies archive
+Uploaded dependencies
```
makesitgo, line 625: yeah, so to your question earlier I think we should definitely remove this to avoid the double-printing of "dependencies" messages

internal/commands/push/command.go
```diff
@@ -198,7 +198,35 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+			status := realm.DependenciesStatus{State: realm.DependenciesStateCreated}
+			for status.State == realm.DependenciesStateCreated {
+				time.Sleep(time.Second)
```
makesitgo, line 211: also, I doubt it affects much more than just readability, but let's push this `Sleep` as well to the end of the for loop

makesitgo (commented):
last few things from me, but this is lookin 🔥

internal/commands/push/command.go
gagik: Yes, as you saw below in the tests it ends up being 2 dependencies-related messages in the tests so I will remove it.

internal/commands/push/command.go
gagik: I see; I misinterpreted that comment and changed the spinner to outside the scope the nested function instead.

internal/commands/push/command.go
```diff
@@ -195,10 +195,38 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+			}
+
+			status := realm.DependenciesStatus{State: realm.DependenciesStateCreated}
+			for status.State == realm.DependenciesStateCreated {
+				var err error
+				status, err = clients.Realm.DependenciesStatus(appRemote.GroupID, appRemote.AppID)
+				if err != nil {
+					return err
+				}
+				s.Suffix = fmt.Sprintf(" Installing dependencies: %s...", status.Message)
```
makesitgo, line 216: [question] maybe you can try to test this out (either by running the command and watching closely or mocking the realm client to suit your needs), but do we experience any "flicker" on this spinner/message when it's being set on every for loop iteration?

i ask b/c if so, then maybe we only want to be calling `s.Suffix = ...` if/when `status.Message` changes (which would then force us to remove the `var err error` declaration and use something like `s, err := clients.Realm.DependenciesStatus(...)` so we can have reference to both past and current statuses

makesitgo (approved):
just one final question I'd like for you to answer via local testing, lmk if you want any help in setting that up

otherwise this looks terrific, really nice work here

🆒   beans

internal/commands/push/command.go
gagik: I did not see any flicker when doing my testing so I don't think it's a problem. Unrelated to this, but I did notice a case of `" Installing dependencies: ...` during testing but this is because of the mock functions not providing any status messages during some tests. I could add status messages to the mock functions if this matters for tests.

makesitgo: Awesome, and I wouldn't both with crafting messages for the tests, that text doesn't persist anyways (we don't assert on it)

I'd say all of this is good to go then!

PR #179, opened 2021-07-06
REALMC-9314: Create accesslist create command

asudhakal:
Implementation of the `create` command for `accessList` and also added `Hidden` attribute to the `CommandMeta`


internal/commands/accesslist/create.go
```diff
@@ -0,0 +1,85 @@
...
+	ui.Print(terminal.NewTextLog("Successfully created allowed IP, id: %s", allowedIP.ID))
+	return nil
+}
+
+func (i *createInputs) Resolve(profile *user.Profile, ui terminal.UI) error {
+	if err := i.ProjectInputs.Resolve(ui, profile.WorkingDirectory, false); err != nil {
+		return err
+	}
+
+	if i.Address == "" {
```
asudhakal, line 75: @kpham1231 I added a pretty straightforward error message here to indicate the logic behind this `if` statement. Let me know how we can rephrase this!

kpham1231: This makes sense to me! I don't think you need to be more specific :)

jandersongo (approved):

internal/commands/accesslist/create.go
makesitgo, line 75: we have a fairly complicated `if` block here, and I think we can dumb it down a bit, at the very least make it a little more direct to read

```go
if i.AllowAll {
  if i.Address != "" { return errTooManyAddresses }
  i.Address = "0.0.0.0"
}

if i.Address != "" && i.UseCurrent { return errTooManyAddresses }
```

i might be off here, but do you think this achieves the same as what you have here?

internal/commands/accesslist/create.go
```diff
@@ -0,0 +1,85 @@
...
+		if i.UseCurrent && i.AllowAll {
+			return errors.New("when you are using this command, you can only provide one IP address or CIDR block at a time")
```
makesitgo, line 77: remove some of the unnecessary language for this error message, and I would also recommend moving this into its own `var errTooManyAddresses = ` (or some name along the same lines)
```suggestion
			return errors.New("must only provide one IP address or CIDR block at a time")
```

internal/commands/accesslist/inputs.go
```diff
@@ -0,0 +1,16 @@
+package accesslist
+
+// Flag names and usages across the accessList commands
+const (
+	flagIP            = "ip"
```
makesitgo, line 5: looking at this, I'd like to merge my stuff and help you through the conflicts here as opposed to the other way around, it will help shape/change a few of these strings.  ultimately, this file should go away and we'll only add it back if/when a flag _does_ need to be re-used or kept consistent across these `accesslist` commands

makesitgo (changes requested):
Just a few various thoughts/comments here, its lookin very good though

I'd like to merge in the programmatic flag stuff (#173) before this just to get you experience working with the new paradigm (considering we're going to be adding more of these commands soon)

internal/commands/commands.go
```diff
@@ -177,4 +178,19 @@ var (
 			},
 		},
 	}
+
+	AccessList = cli.CommandDefinition{
+		CommandMeta: cli.CommandMeta{
+			Use:         "accessList",
```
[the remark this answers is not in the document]
asudhakal: I do believe this was discussed and we want to stay consistent with Atlas naming conventions for this one and stick with `accessList` as the default and `accesslist` as the alias. Here's the [discussion](https://docs.google.com/document/d/1foVVydfAxTsA4rVPLz3x4YDlHcnXBgeDKgD3a2RPBtU/edit?disco=AAAAMLbjyu4) behind it.

internal/cloud/realm/access_list_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+// TODO(REALMC-9207): Add tests once backend is fully supported
+func TestRealmIPAccess(t *testing.T) {
+	u.SkipUnlessRealmServerRunning(t)
+
+	t.Run("should fail without an auth client", func(t *testing.T) {
+		client := realm.NewClient(u.RealmServerURL() + "")
```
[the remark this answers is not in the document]
asudhakal: error on my part! I believe `secrets` was doing this and figured it was required. I'll get rid of it there too.

internal/commands/commands.go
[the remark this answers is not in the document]
makesitgo: I'd love to hear the rationale that was used "per project check in meeting", b/c it seemed like that whole discussion up to that point was agreeing with what I suggested here.

you can put the blame on me, but can we just have one last thread on slack about the naming here?  b/c 2 days after that discussion I know we had zoom call (christine sumedha and i) clearing up some confusion about the naming of these commands

internal/commands/accesslist/create.go
asudhakal: Yes! awesome, this is much cleaner.

internal/commands/commands.go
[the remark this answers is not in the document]
asudhakal: Just had a discussion with Sumedha, seems like we're going to go with `accesslist` as the default and `access-list` as an alias for now!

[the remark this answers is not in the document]
tsedgwick: Since this is a security based command, we should stay consistent with atlas (`accessList`) as the default. I'm fine with others as aliases. The last thing we want to do is add confusion or cognitive overhead to a security control.

best case scenario here would be to have mcli change their default, but that's a long shot.

internal/commands/accesslist/inputs.go
asudhakal: Sounds good!

internal/commands/accesslist/create.go
```diff
@@ -0,0 +1,111 @@
...
+	ui.Print(terminal.NewTextLog("Successfully created allowed IP, id: %s", allowedIP.ID))
+	return nil
+}
+
+func (i *createInputs) Resolve(profile *user.Profile, ui terminal.UI) error {
+	if err := i.ProjectInputs.Resolve(ui, profile.WorkingDirectory, true); err != nil {
+		return err
+	}
+
+	if i.Address == "" && !(i.AllowAll || i.UseCurrent) {
```
makesitgo, line 94: if you move this check down to the bottom (at least after `if i.AllowAll`, but might read nicer if at the very end of this function), then you could simplify this if logic slightly to be:

```go
if i.Address == "" && !i.UseCurrent { ... }
```

internal/commands/accesslist/create.go
```diff
@@ -0,0 +1,111 @@
...
+		if err := ui.AskOne(&i.Address, &survey.Input{Message: "IP Address"}); err != nil {
```
makesitgo, line 95: can you also add a `// TODO(REALMC-9532): validate the user does not enter an empty string` above this line?

no action needed now (especially since this command is hidden).  this should likely be done in a few other places, but it will give the ticket-doer at least one actionably place to look for first

internal/commands/accesslist/create_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+func TestCreateInputs(t *testing.T) {
+	for _, tc := range []struct {
+		description string
+		inputs      createInputs
+		procedure   func(c *expect.Console)
+		test        func(t *testing.T, i createInputs)
+	}{
+		{
+			description: "Should not prompt for inputs when flags provide the data",
```
makesitgo, line 21: we're avoiding capitalization and punctuation in test descriptions, make this `"should not prompt..." (and similar adjustments for the others in this file)

(you'll see plenty of instances where this pattern is broken, it was instituted a bit late in the game.  but it's what we're doing going forward)

---

also, can we split this test table out into at least two variants: one for interactivity and one for not?

the non-interactive tests can be written so much simpler that its worth keeping them separate from the ones where we need the fancy pseudo-terminal

internal/commands/commands.go
```diff
@@ -177,4 +178,19 @@ var (
...
+			Aliases:     []string{"accesslist", "access-list"},
+			Description: "Manage your allowed IP addresses and CIDR blocks",
+			//Hidden:      true, // TODO(REALMC-9164): Remove 'Hidden'
```
makesitgo, line 187: related to my other comment about `Hidden`, but by not commenting this out we can now hide all `accesslist` commands in one spot, and not worry about setting `Hidden` individually on each subcommand

internal/commands/commands.go
```diff
@@ -177,4 +178,19 @@ var (
 			},
 		},
 	}
+
+	AccessList = cli.CommandDefinition{
+		CommandMeta: cli.CommandMeta{
+			Use:         "accessList",
+			Aliases:     []string{"accesslist", "access-list"},
+			Description: "Manage your allowed IP addresses and CIDR blocks",
```
makesitgo, line 186: going to bump this comment one more time, it was resolved but nothing changed...  make this `Description`:

```go
Description: "Manage the allowed IP addresses and CIDR blocks of your Realm app",
```

internal/commands/accesslist/create_test.go
```diff
@@ -0,0 +1,120 @@
...
+			assert.Nil(t, tc.inputs.Resolve(profile, ui))
+
+			console.Tty().Close() // flush the writers
+			<-doneCh              // wait for procedure to complete
+
+			tc.test(t, tc.inputs)
+		})
+	}
+
+	t.Run("Should error when more than one address given", func(t *testing.T) {
```
makesitgo, line 86: likewise for capitalization here

also in general its best to avoid code-like words in the test descriptions (if they ever change, these can get out of date quickly).  considering the code-like words here also have punctuation (`-`), let's find a different way to specify these.  could be something like:

* with allow all specified and an address provided
* with use current specified and an address provided
* with both allow all and use current specified
* with both allow all and use current specified and an address provided

apologies for the nits like this, it just tends to improve the readability of the test output

---

also super not critical but these all could fit nicely into a test table of their own

internal/commands/accesslist/create_test.go
```diff
@@ -0,0 +1,120 @@
...
+		t.Run("with both address and allow-all", func(t *testing.T) {
+			profile := mock.NewProfile(t)
+
+			inputs := createInputs{Address: "0.0.0.0", AllowAll: true}
+			err := inputs.Resolve(profile, nil)
+
+			assert.Equal(t, err, errors.New("must only provide one IP address or CIDR block at a time"))
```
makesitgo, line 93: two comments on this:

* I wasn't very careful looking at this until now, but the necessary form for these asserts is `assert.Equal(t, expected, actual)`
  - you have it backwards here, and by flipping it the test failure messages will actually make sense
  - can you do an audit of all your other test asserts and ensure these are also in the right order
* when you make the change to `errTooManyAddresses`, then you can just assert equality with that

internal/commands/accesslist/create_test.go
```diff
@@ -0,0 +1,120 @@
...
+		})
+		t.Run("with address, allow-all and use-current", func(t *testing.T) {
+			profile := mock.NewProfile(t)
+
+			inputs := createInputs{Address: "0.0.0.0", AllowAll: true, UseCurrent: true}
+			err := inputs.Resolve(profile, nil)
+
+			assert.Equal(t, err, errors.New("must only provide one IP address or CIDR block at a time"))
+		})
+	})
```
makesitgo, line 119: we should add a few more test cases for `cmd.Handler`:

* a success test case that proves we print out the right messaging
* a failure test case that proves we error out if `realmClient.FindApps` fails
* a failure test case that proves we error out if `realmClient.AllowedIPCreate` fails

makesitgo (commented):
next batch of comments from me... mostly organizational/readability focused, and then just want to see a few more test cases on the actual `Handler` code paths

internal/commands/accesslist/create_test.go
```diff
@@ -0,0 +1,218 @@
...
+	allowedIPUseCurrent := false
+	allowedIPAllowAll := false
+	app := realm.App{
+		ID:          appID,
+		GroupID:     projectID,
+		ClientAppID: "eggcorn-abcde",
+		Name:        "eggcorn",
+	}
+
+	t.Run("should create an allowed IP", func(t *testing.T) {
```
makesitgo, line 30:
```suggestion
	t.Run("should create an allowed ip", func(t *testing.T) {
```

internal/commands/accesslist/create_test.go
```diff
@@ -0,0 +1,218 @@
...
+				App:     appID,
+			},
+			Address:    allowedIPAddress,
+			Comment:    allowedIPComment,
+			UseCurrent: allowedIPUseCurrent,
+			AllowAll:   allowedIPAllowAll,
+		}}
+
+		assert.Nil(t, cmd.Handler(nil, ui, cli.Clients{Realm: realmClient}))
+		assert.Equal(t, "Successfully created allowed IP, id: allowedIPID\n", out.String())
```
makesitgo, line 63: [test/style nit] I'm personally not the biggest fan of having simple `string`/`bool` vars declared in a test for re-usability (see `allowedIPID := "allowedIPID"` above), but if we are going to have them then we should use them in the test assert via `fmt.Sprintf(...)`

internal/commands/accesslist/create_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+				cmd := &CommandCreate{}
+
+				err := cmd.Handler(nil, nil, cli.Clients{Realm: realmClient})
+				assert.Equal(t, tc.expectedErr, err)
+			})
+		}
+	})
+}
+func TestCreateInputs(t *testing.T) {
```
makesitgo, line 117:
```suggestion
func TestAllowedIPCreateInputs(t *testing.T) {
```

makesitgo (approved):
final few things from me, but great job with this!

can we just update the commit string a bit?  something along the lines of `Create accesslist create command` or `Build out accesslist create command`?

🆒  beans

PR #180, opened 2021-07-09
REALMC-9541: Fix beta.7 template handling

UnicodeSnowman:
merging this into `support/v2.0.x` (i.e. `beta.7`)

### DIFF
--- a/internal/commands/app/new_app_inputs.go
+++ b/internal/commands/app/new_app_inputs.go
@@ -64,7 +64,7 @@ func (i *newAppInputs) resolveRemoteApp(ui terminal.UI, rc realm.Client) (realm.
 }
 
 func (i *newAppInputs) resolveTemplateID(ui terminal.UI, client realm.Client) error {
-	if i.Template == "" && ui.AutoConfirm() {
+	if i.Template == "" {
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


internal/commands/app/new_app_inputs.go
```diff
@@ -73,50 +73,16 @@ func (i *newAppInputs) resolveTemplateID(ui terminal.UI, client realm.Client) er
 		return err
```
andrewzlchen, line 73: opt/nit: since it seems like template starter apps is going to be disabled in beta.7 still, we probably don't need this client call.

We should probably also comment out the flag in [here](https://github.com/10gen/realm-cli/blob/90c58fda935f6a9dedfd24294dfc0a332f0a2759/internal/commands/app/create.go#L93-L101) so people don't run in to this unnecessarily

andrewzlchen (approved):
basically looks good to me, but I think we should first comment out the --template flag in the `create.go:Flags()` function.

I think we should just comment the flag out altogether so users don't file bug tickets for an incomplete feature.

internal/commands/app/new_app_inputs.go
makesitgo: just pointing out, if we comment out the flag, then the rest of this func after the first if check is effectively dead code (in addition to removing `ui.Auto.Confirm` above as well).  easy one line comment to avoid all of this (and could then probably just not change the rest)

UnicodeSnowman: hmm I thought the idea was that we would still allow users who want to opt in to using templates could specify the flag? happy to ditch this, but that was the assumption I was operating under.

andrewzlchen: hmm if we wanted to do that, I think we'd need to rebase this PR with the changes on the master branch to get the pull and create commands.

However, as they are kinda rough around the edges, I'd opt to remove for now and add templates support in 2.1.0

UnicodeSnowman:
@makesitgo @andrewzlchen ready for a final look, can do the GA release after this goes in

makesitgo (approved):
works for me

🆒  beans

PR #181, opened 2021-07-13
REALMC-9317: Build accesslist list command

asudhakal:
Built out the `accessList list` command for the CLI


internal/cloud/realm/client.go
```diff
@@ -77,6 +77,7 @@ type Client interface {
 	ClientTemplate(groupID, appID, templateID string) (*zip.Reader, bool, error)
 	CompatibleTemplates(groupID, appID string) ([]Template, error)
 
+	AllowedIPs(groupID, appID string) (AccessList, error)
```
makesitgo, line 80: feel obligated to ask if we can either a) rename this to `AccessList(...)` or b) adjust the return signature to provide `([]AllowedIP, string, error)` (where the `string` is just the current IP)

feel free to tell me this makes the most sense semantically and we'll keep it as such, but otherwise I think I lean towards a) over b)

internal/cloud/realm/access_list_test.go
```diff
@@ -29,6 +29,12 @@ func TestRealmIPAccess(t *testing.T) {
 		testApp, teardown := setupTestApp(t, client, groupID, "accesslist-test")
 		defer teardown()
 
+		t.Run("should have no allowed IPs upon app initialization", func(t *testing.T) {
+			accessList, err := client.AllowedIPs(groupID, testApp.ID)
+			assert.Nil(t, err)
+			assert.Equal(t, 0, len(accessList.AllowedIPs))
```
makesitgo, line 35: worth having an assert on the `current_ip` here as well?

internal/cloud/realm/access_list_test.go
```diff
@@ -38,7 +44,12 @@ func TestRealmIPAccess(t *testing.T) {
 			assert.Nil(t, err)
 			assert.Equal(t, address, allowedIP.Address)
 			assert.Equal(t, comment, allowedIP.Comment)
-		})
 
+			t.Run("and list all app allowed IPs", func(t *testing.T) {
+				accessList, err := client.AllowedIPs(groupID, testApp.ID)
+				assert.Nil(t, err)
+				assert.Equal(t, []realm.AllowedIP{allowedIP}, accessList.AllowedIPs)
```
makesitgo, line 51: likewise, is it worth asserting on `current_ip` here?

internal/cloud/realm/access_list.go
```diff
@@ -26,6 +32,29 @@ type allowedIPCreatePayload struct {
...
+		http.MethodGet,
+		fmt.Sprintf(allowedIPsPathPattern, groupID, appID),
+		api.RequestOptions{},
+	)
+	if resErr != nil {
+		return AccessList{}, resErr
+	}
+
+	if res.StatusCode != http.StatusOK {
+		return AccessList{}, api.ErrUnexpectedStatusCode{"get allowed ips", res.StatusCode}
```
makesitgo, line 46: depending on where we land with naming the client method, this should update accordingly

internal/commands/accesslist/list.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/10gen/realm-cli/internal/utils/flags"
+)
+
+// CommandMetaList is the command meta for the `accesslist list` command
+var CommandMetaList = cli.CommandMeta{
+	Use:         "list",
+	Aliases:     []string{"ls"},
+	Display:     "accesslist list",
+	Description: "List the allowed entries in the Access List of your Realm app",
+	HelpText:    `This will display the IP addresses/CIDR blocks in your Access List`,
```
makesitgo, line 19:
```suggestion
	HelpText:    `This will display the IP addresses and/or CIDR blocks in the Access List of your Realm app`,
```

internal/commands/accesslist/list.go
```diff
@@ -0,0 +1,87 @@
...
+		return appErr
+	}
+
+	accessList, accessListErr := clients.Realm.AllowedIPs(app.GroupID, app.ID)
+	if accessListErr != nil {
+		return accessListErr
+	}
+
+	if len(accessList.AllowedIPs) == 0 {
+		ui.Print(terminal.NewTextLog("No available allowed IPs to show"))
```
makesitgo, line 58: we should consistently refer to "ip addresses" (and/or "cidr blocks") in our CLI messaging.  you likely have a better idea than i how to go about that the right way, but across all help/usage strings and the command logs, let's make sure the language is consistent

internal/commands/accesslist/list.go
```diff
@@ -0,0 +1,87 @@
...
+		return nil
+	}
+
+	ui.Print(terminal.NewTableLog(
+		fmt.Sprintf("Found %d allowed IPs", len(accessList.AllowedIPs)),
+		tableHeaders,
+		tableRowsList(accessList.AllowedIPs)...,
```
makesitgo, line 65: [question] do we even care about the `current_ip` field when querying this?  if not, I think that's our answer to the "client name" question: keep it as `AllowedIPs` but only return (`[]AllowedIP, error)` from the method

internal/commands/accesslist/list.go
```diff
@@ -0,0 +1,87 @@
...
+		return accessListErr
+	}
+
+	if len(accessList.AllowedIPs) == 0 {
+		ui.Print(terminal.NewTextLog("No available allowed IPs to show"))
+		return nil
+	}
+
+	ui.Print(terminal.NewTableLog(
+		fmt.Sprintf("Found %d allowed IPs", len(accessList.AllowedIPs)),
```
makesitgo, line 63: same comment about using the same language when referring to ip addresses/cidr blocks

but another note here that you should probably include `(s)` to indicate the optional plurality (that or use the `len` value to decide if pluralization is needed or not, but that feels like overkill to me)

internal/commands/accesslist/output.go
```diff
@@ -0,0 +1,38 @@
...
+const (
+	headerAddress = "IP Address"
+	headerComment = "Comment"
+)
+
+var (
+	tableHeaders = []string{headerAddress, headerComment}
+)
+
+type allowedIPOutputs []allowedIPOutput
```
makesitgo, line 14: this looks like some copy/paste that ultimately was never used

fwiw, i don't think having this abstraction is necessary yet until we're printing allowed ips in tabular format in more than one place.  seems like you could move `tableHeaders` to the `list.go` file (maybe rename it to `listTableHeaders`) and then delete this whole file

internal/commands/accesslist/list.go
```diff
@@ -0,0 +1,87 @@
...
+		tableHeaders,
+		tableRowsList(accessList.AllowedIPs)...,
+	))
+	return nil
+}
+
+func tableRowsList(allowedIPs []realm.AllowedIP) []map[string]interface{} {
```
makesitgo, line 70: the helper func probably isnt necessary here either, could just initialize a `tableRows := make(...)` inside the body of `Handler` here

internal/commands/accesslist/list_test.go
```diff
@@ -0,0 +1,122 @@
...
+		"0.0.0.0",
+	}
+
+	for _, tc := range []struct {
+		description    string
+		accessList     realm.AccessList
+		expectedOutput string
+	}{
+		{
+			description:    "should list no allowed IPs with no app allowed IPs found",
```
makesitgo, line 38: reminder to keep capitalization and punctuation out of test descriptions

makesitgo (commented):
first thoughts here for you, nothing major, looks nice and easy here

internal/cloud/realm/access_list_test.go
asudhakal: Since we aren't returning the `current_ip` anymore, we can simply ignore it.

internal/cloud/realm/client.go
asudhakal: Going with `AllowedIPs(groupID, appID string) ([]AllowedIP, error)` as per the other comment below.

internal/cloud/realm/access_list.go
```diff
@@ -26,6 +32,29 @@ type allowedIPCreatePayload struct {
...
+		http.MethodGet,
+		fmt.Sprintf(allowedIPsPathPattern, groupID, appID),
+		api.RequestOptions{},
+	)
+	if resErr != nil {
+		return nil, resErr
+	}
+
+	if res.StatusCode != http.StatusOK {
+		return nil, api.ErrUnexpectedStatusCode{"get allowed ips and/or CIDR blocks", res.StatusCode}
```
makesitgo, line 46: [meganit] considering the context of this client call and that this error condition should never hit in the first place, lets just go with:
```suggestion
		return nil, api.ErrUnexpectedStatusCode{"get allowed ips", res.StatusCode}
```

internal/cloud/realm/access_list.go
```diff
@@ -12,6 +12,12 @@ const (
 	allowedIPsPathPattern = appPathPattern + "/security/allowed_ips"
 )
 
+// AccessList is a list of allowed IPs stored in a Realm app
+type AccessList struct {
+	AllowedIPs []AllowedIP `json:"allowed_ips"`
+	CurrentIP  string      `json:"current_ip"`
```
makesitgo, line 18: [nit] I'd recommend just removing this field until its actually needed (similar to `realm.App` and how it only partially defines the fields it needs)

internal/cloud/realm/access_list.go
```diff
@@ -42,7 +71,7 @@ func (c *client) AllowedIPCreate(groupID, appID, ipAddress, comment string, useC
 	}
 
 	if res.StatusCode != http.StatusCreated {
-		return AllowedIP{}, api.ErrUnexpectedStatusCode{"create allowed ip", res.StatusCode}
+		return AllowedIP{}, api.ErrUnexpectedStatusCode{"create allowed ip or CIDR block", res.StatusCode}
```
makesitgo, line 74: sorry it seems like i'm going back and forth on this, but these errors are not intended to be user facing, we can keep them concise and relevant to the specific context of the client call itself here

```suggestion
		return AllowedIP{}, api.ErrUnexpectedStatusCode{"create allowed ip", res.StatusCode}
```

makesitgo (approved):
just tiny things from me here... nice work!

🆒  beans

PR #184, opened 2021-07-14
REALMC-9315: Build accesslist update command

asudhakal:


internal/commands/accesslist/update.go
```diff
@@ -0,0 +1,151 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/cli/user"
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/10gen/realm-cli/internal/utils/flags"
+	"github.com/AlecAivazis/survey/v2"
```
makesitgo, line 12: newline here

internal/commands/accesslist/update.go
```diff
@@ -0,0 +1,151 @@
...
+	Comment    string
+}
+
+// CommandMetaUpdate is the command meta for the `accesslist update` command
+var CommandMetaUpdate = cli.CommandMeta{
+	Use:         "update",
+	Display:     "accesslist update",
+	Description: "Modify an IP address or CIDR block in the Access List of your Realm app",
+	HelpText: `Changes an existing entry from the Access List of your Realm app. You will be
+prompted to select an IP address or CIDR block if none is provided in the initial command.`,
```
makesitgo, line 28: maybe this last sentence should read as: "You will be prompted to select an IP address or CIDR block to update if neither is specified."

also, this second line is over the 80 char limit... depending on how you potentially reword it here, I'd go in and double check the line widths after that

internal/commands/accesslist/update.go
```diff
@@ -0,0 +1,151 @@
...
+					Description: "Specify the existing IP address or CIDR block that you would like to modify",
+				},
+			},
+		},
+		flags.StringFlag{
+			Value: &cmd.inputs.NewAddress,
+			Meta: flags.Meta{
+				Name: "new-ip",
+				Usage: flags.Usage{
+					Description: "Specify the new IP address or CIDR block that will replace the existing entry",
```
makesitgo, line 56: seems like this flag too is "optional" (in the same sense that `--comment` is).  if that is the case, i would opt to just remove that `Note` from each flag and leave the usages a little simpler/shorter

internal/commands/accesslist/update.go
```diff
@@ -0,0 +1,151 @@
...
+				},
+			},
+		},
+		flags.StringFlag{
+			Value: &cmd.inputs.Comment,
+			Meta: flags.Meta{
+				Name: "comment",
+				Usage: flags.Usage{
+					Description: "Add or edit a comment to the IP address or CIDR block that is being modified in the Access List",
```
makesitgo, line 65: can probably drop the "in the Access List" here

internal/commands/accesslist/update.go
```diff
@@ -0,0 +1,151 @@
...
+				return allowedIP, nil
+			}
+		}
+		return realm.AllowedIP{}, fmt.Errorf("unable to find allowed IP: %s", i.Address)
+	}
+
+	selectableAllowedIPs := map[string]realm.AllowedIP{}
+	selectableOptions := make([]string, len(allowedIPs))
+	for i, allowedIP := range allowedIPs {
+		option := allowedIP.ID
```
makesitgo, line 134: have you tested this locally?  seems to me we are just going to end up presenting the user with a list of ID values, probably not all that useful for them in this context

I think with secrets we do something like `<ID> - <Name>` ... take a look around and try to see if we've landed on a pattern for things like that.  i think another thing here that might be useful would be dynamically adding on ` (<Comment>)` if there's a non-blank comment present, just to really give the user an idea of what they're updating

internal/commands/accesslist/update.go
```diff
@@ -0,0 +1,151 @@
...
+		selectableOptions[i] = option
+		selectableAllowedIPs[option] = allowedIP
+	}
+
+	var selected string
+	if err := ui.AskOne(
+		&selected,
+		&survey.Select{
+			Message: "Which IP address or CIDR block would you like to update?",
```
makesitgo, line 143: i think more declarative messaging works a bit better here, something along the lines of:

`"Select an IP address or CIDR block to update"`

internal/commands/accesslist/update_test.go
```diff
@@ -0,0 +1,207 @@
...
+			testNewAddress: "2.2.2.2",
+			testComment:    "new comment",
+		},
+		{
+			description:    "should return a successful message for an update with only an address",
+			testAddress:    "192.1.1.1",
+			testNewAddress: "68.192.33.2",
+		},
+		{
+			description: "should return a successful message for an empty update",
```
makesitgo, line 47: should we be testing this scenario if its not technically possible based on the way inputs are resolved?

internal/commands/accesslist/update_test.go
```diff
@@ -0,0 +1,207 @@
...
+			inputs      updateInputs
+			clientSetup func() realm.Client
+			expectedErr error
+		}{
+			{
+				description: "if there is no app",
+				clientSetup: func() realm.Client {
+					return mock.RealmClient{
+						FindAppsFn: func(filter realm.AppFilter) ([]realm.App, error) {
+							return nil, errors.New("Something went wrong with the app")
```
makesitgo, line 91: [ubernit] i know its just test code, but errors shouldn't be starting with capital letters, might as well adjust that the few times it occurs here

internal/commands/accesslist/update_test.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func TestAllowedIPUpdateInputs(t *testing.T) {
+	allowedIPs := []realm.AllowedIP{
+		{ID: "address1", Address: "0.0.0.0", Comment: "comment"},
+		{ID: "address2", Address: "192.1.1.1"},
+		{ID: "address3", Address: "192.158.1.38", Comment: "cool comment"},
+	}
+
+	t.Run("should return the allowed ip if searching by address", func(t *testing.T) {
```
makesitgo, line 166: i would nest these tests inside a `t.Run("resolving the allowed ip", func(t *testing.T) { ... })`

and then also add tests for `inputs.Resolve` (mainly that we error if/when both comment and new address are not specified)

makesitgo (commented):
first pass here, just a few questions and nits from me, otherwise looks pretty straightforward and clean here

internal/commands/accesslist/update_test.go
asudhakal: noticed that secrets was testing it like this too, but yeah I agree we don't necessarily need to be testing this functionality since it isn't a possible scenario.

makesitgo: could probably drive-by remove that one as well then

internal/commands/secrets/update_test.go
```diff
@@ -52,10 +52,6 @@ func TestSecretUpdateHandler(t *testing.T) {
 			testSecret:  "secretID2",
 			testValue:   "newValue2",
 		},
-		{
```
christineschen, line 55: [q] is this just an unrelated driveby?

christineschen (approved):
looking good! just a couple questions on naming conventions

internal/commands/secrets/update_test.go
asudhakal: yes!

internal/cloud/realm/access_list.go
```diff
@@ -10,6 +10,7 @@ import (
 
 const (
 	allowedIPsPathPattern = appPathPattern + "/security/allowed_ips"
+	allowedIPPathPattern  = allowedIPsPathPattern + "/%s"
```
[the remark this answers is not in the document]
asudhakal: yes, will update the routes.

jandersongo (approved):

internal/cloud/realm/access_list.go
```diff
@@ -81,3 +82,24 @@ func (c *client) AllowedIPCreate(groupID, appID, ipAddress, comment string, useC
 	}
 	return allowedIP, nil
 }
+
+func (c *client) AllowedIPUpdate(groupID, appID, allowedIPID, newAddress, comment string) error {
```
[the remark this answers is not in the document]
makesitgo: just bumping this as it seems unchanged

internal/cloud/realm/client.go
```diff
@@ -78,7 +78,8 @@ type Client interface {
 	CompatibleTemplates(groupID, appID string) ([]Template, error)
 
 	AllowedIPs(groupID, appID string) ([]AllowedIP, error)
-	AllowedIPCreate(groupID, appID, ipAddress, comment string, useCurrent bool) (AllowedIP, error)
+	AllowedIPCreate(groupID, appID, address, comment string, useCurrent bool) (AllowedIP, error)
+	AllowedIPUpdate(groupID, appID, allowedIPID, newAddress, comment string) error
```
makesitgo, line 82: if we go with `newComment` in the implementation, lets do the same here

internal/commands/accesslist/update.go
```diff
@@ -0,0 +1,160 @@
...
+				Name: "ip",
+				Usage: flags.Usage{
+					Description: "Specify the existing IP address or CIDR block that you would like to modify",
+				},
+			},
+		},
+		flags.StringFlag{
+			Value: &cmd.inputs.NewAddress,
+			Meta: flags.Meta{
+				Name: "new-ip",
```
makesitgo, line 56: just a thought here (cc: @christineschen @sumedhamehta )

what do we think about adopting a naming strategy (it could be for flag names, future command names, etc) where instead of something like `"new-ip"`, we use `"ip-new"`?

i mention this b/c if/when auto-complete is ever supported, this "inverted" name would prove more useful

internal/cloud/realm/access_list.go
```diff
@@ -81,3 +82,24 @@ func (c *client) AllowedIPCreate(groupID, appID, ipAddress, comment string, useC
...
+	res, err := c.doJSON(
+		http.MethodPut,
```
makesitgo, line 88: [question/nit] does this maybe want to be a `PATCH`?  from its usage, it looks like we'll only conditionally update the IP Address with values that are non-blank

internal/commands/accesslist/update.go
```diff
@@ -0,0 +1,160 @@
...
+	return nil
+}
+
+func (i *updateInputs) Resolve(profile *user.Profile, ui terminal.UI) error {
+	if err := i.ProjectInputs.Resolve(ui, profile.WorkingDirectory, true); err != nil {
+		return err
+	}
+
+	if i.NewAddress == "" && i.Comment == "" {
+		return errors.New("must set either --new-ip or --comment when updating an allowed IP address or CIDR block")
```
makesitgo, line 116: probably worth extracting the flag names to consts and using `fmt.Errorf` here

check out `utils/flags/arg.go`, for string formatting here you could do something like:

```go
return fmt.Errorf(
  "must set either %s or %s when updating an allowed IP address or CIDR block",
  flags.Arg{Name: flagIPNew}, flags.Arg{Name: flagComment},
)
```

internal/commands/accesslist/update_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/cloud/realm"
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/10gen/realm-cli/internal/utils/test/mock"
+	"github.com/Netflix/go-expect"
```
makesitgo, line 11: newline here

internal/commands/accesslist/update_test.go
```diff
@@ -0,0 +1,222 @@
...
+		testComment    string
+	}{
+		{
+			description:    "should return a successful message for a successful update for address and comment",
+			testAddress:    "0.0.0.0",
+			testNewAddress: "2.2.2.2",
+			testComment:    "new comment",
+		},
+		{
+			description:    "should return a successful message for an update with only an address",
```
makesitgo, line 43: can we add one more test case for updating with only a comment?

with that, let's also add a few `var`s for what gets passed into the mocked client's `AllowedIPUpdate` function and have asserts on those as well

internal/commands/accesslist/update_test.go
```diff
@@ -0,0 +1,222 @@
...
+		t.Run("should return the allowed ip if searching by address", func(t *testing.T) {
+			inputs := updateInputs{
+				Address: "0.0.0.0",
+			}
+			result, err := inputs.resolveAllowedIP(nil, allowedIPs)
+			assert.Nil(t, err)
+			assert.Equal(t, allowedIPs[0], result)
+		})
+
+		t.Run("should return an error if we cannot find the allowed ip to update", func(t *testing.T) {
```
makesitgo, line 173: avoid words like "we"... something along the lines of:

`"should return an error if allowed ip is not found"`

internal/commands/accesslist/update_test.go
```diff
@@ -0,0 +1,222 @@
...
+			inputs := updateInputs{
+				Address: "1.1.1.1",
+			}
+
+			_, err := inputs.resolveAllowedIP(nil, allowedIPs)
+			assert.Equal(t, errors.New("unable to find allowed IP: 1.1.1.1"), err)
+		})
+
+		t.Run("should return an error if one of new ip or comment is not provided", func(t *testing.T) {
```
makesitgo, line 182:
```suggestion
		t.Run("should return an error if neither new ip nor comment is provided", func(t *testing.T) {
```

internal/commands/accesslist/update_test.go
```diff
@@ -0,0 +1,222 @@
...
+			profile := mock.NewProfile(t)
+			inputs := updateInputs{
+				Address: "1.1.1.1",
+			}
+
+			err := inputs.Resolve(profile, nil)
+			assert.Equal(t, errors.New("must set either --new-ip or --comment when updating an allowed IP address or CIDR block"), err)
+		})
+
+		t.Run("should prompt for address when none provided", func(t *testing.T) {
```
makesitgo, line 192:
```suggestion
		t.Run("should prompt for address when is is not provided", func(t *testing.T) {
```

internal/utils/test/mock/realm_client.go
```diff
@@ -503,24 +504,34 @@ func (rc RealmClient) CompatibleTemplates(groupID, appID string) ([]realm.Templa
...
 }
 
-// AllowedIPs calls the mocked AllowedIPs implementation if provided,
+// AllowedIPUpdate calls the mocked AllowedIPs implementation if provided,
 // otherwise the call falls back to the underlying realm.Client implementation.
 // NOTE: this may panic if the underlying realm.Client is left undefined
-func (rc RealmClient) AllowedIPs(groupID, appID string) ([]realm.AllowedIP, error) {
-	if rc.AllowedIPsFn != nil {
-		return rc.AllowedIPsFn(groupID, appID)
+func (rc RealmClient) AllowedIPUpdate(groupID, appID, allowedIPID, newAddress, comment string) error {
```
makesitgo, line 530: same thing with `newComment` if we go that route here

internal/utils/test/mock/realm_client.go
```diff
@@ -503,24 +504,34 @@ func (rc RealmClient) CompatibleTemplates(groupID, appID string) ([]realm.Templa
...
 	if rc.AllowedIPCreateFn != nil {
-		return rc.AllowedIPCreateFn(groupID, appID, ipAddress, comment, useCurrent)
+		return rc.AllowedIPCreateFn(groupID, appID, address, comment, useCurrent)
 	}
-	return rc.AllowedIPCreate(groupID, appID, ipAddress, comment, useCurrent)
+	return rc.AllowedIPCreate(groupID, appID, address, comment, useCurrent)
 }
 
-// AllowedIPs calls the mocked AllowedIPs implementation if provided,
+// AllowedIPUpdate calls the mocked AllowedIPs implementation if provided,
```
makesitgo, line 527:
```suggestion
// AllowedIPUpdate calls the mocked AllowedIPUpdate implementation if provided,
```

makesitgo (commented):
I think these are the last few things from me, should be good to go afterwards!

internal/cloud/realm/access_list.go
asudhakal: I think we can go with `PATCH` for this one, since we're updating an existing resource and not replacing it completely.

makesitgo: id just check/run it by the rest of the project team first... if so, then its probably best left as a `TODO(REALMC-XXXX)` here so the corresponding backend change can go with it

internal/commands/accesslist/update.go
makesitgo: should use `flagIPNew` as the `Name` here

internal/commands/accesslist/update.go
```diff
@@ -0,0 +1,167 @@
...
+				Name: "new-ip",
+				Usage: flags.Usage{
+					Description: "Specify the new IP address or CIDR block that will replace the existing entry",
+				},
+			},
+		},
+		flags.StringFlag{
+			Value: &cmd.inputs.NewComment,
+			Meta: flags.Meta{
+				Name: "comment",
```
makesitgo, line 69: should use `flagComment` here

internal/commands/accesslist/update_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+			realmClient.FindAppsFn = func(filter realm.AppFilter) ([]realm.App, error) {
+				return []realm.App{app}, nil
+			}
+			realmClient.AllowedIPsFn = func(groupID, appID string) ([]realm.AllowedIP, error) {
+				return allowedIPs, nil
+			}
+			realmClient.AllowedIPUpdateFn = func(groupID, appID, allowedIPID, newAddress, newComment string) error {
+				updateArgs = struct {
+					groupID, appID, allowedIPID string
```
makesitgo, line 69: ideally we're asserting on every value passed through here, not just the ids

makesitgo (approved):
very small things left for me here, nice work here!

PR #185, opened 2021-07-19
REALMC-9452: Write template clients to file system in sorted order

andrewzlchen:
- REALMC-9452: Write client templates to file system in sorted order

### DIFF
--- a/internal/commands/pull/command.go
+++ b/internal/commands/pull/command.go
@@ -111,7 +111,7 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		return err
 	}
 
-	var clientTemplates []pkg
+	var clientTemplates []clientTemplate
 	if app.TemplateID != "" {
 		clientTemplates, err = cmd.inputs.resolveClientTemplates(ui, clients.Realm, app.GroupID, app.ID)
 		if err != nil {
@@ -216,7 +216,7 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 
 	successfulTemplateWrites := make([]string, 0, len(clientTemplates))
 	for _, ct := range clientTemplates {
-		if err := local.WriteZip(filepath.Join(pathFrontend, ct.id), ct.client); err != nil {
+		if err := local.WriteZip(filepath.Join(pathFrontend, ct.id), ct.zipPkg); err != nil {
 			return fmt.Errorf("unable to save template '%s' to disk: %s", ct.id, err)
 		}
 		successfulTemplateWrites = append(successfulTemplateWrites, ct.id)
--- a/internal/commands/pull/command.go
+++ b/internal/commands/pull/command.go
@@ -111,8 +111,6 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		return err
 	}
 
-	// TODO(REALMC-9462): add template_id to /apps payload to remove FindApp within resolveRemoteApp in order to
-	//  determine whether the current app has been created with a template
 	var clientZipPkgs []clientTemplate
 	if app.TemplateID != "" {
 		clientZipPkgs, err = cmd.inputs.resolveClientTemplates(ui, clients.Realm, app.GroupID, app.ID)
--- a/internal/commands/pull/command.go
+++ b/internal/commands/pull/command.go
@@ -111,8 +111,9 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		return err
 	}
 
-	// TODO(REALMC-9462): maybe make this less hacky (pass app in to resolveClientTemplates perhaps?)
-	var clientZipPkgs map[string]*zip.Reader
+	// TODO(REALMC-9462): add template_id to /apps payload and pass into resolveClientTemplate to correctly report an
+	//  error when an app has not created with a template
+	var clientZipPkgs []clientTemplate
 	if app.TemplateID != "" {
 		clientZipPkgs, err = cmd.inputs.resolveClientTemplates(ui, clients.Realm, app.GroupID, app.ID)
 		if err != nil {
@@ -216,12 +217,11 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 	}
 
 	successfulTemplateWrites := make([]string, 0, len(clientZipPkgs))
-	// TODO(REALMC-9452) fix to iterate in a deterministic order
-	for templateID, templateZipPkg := range clientZipPkgs {
-		if err := local.WriteZip(filepath.Join(pathFrontend, templateID), templateZipPkg); err != nil {
-			return fmt.Errorf("unable to save template '%s' to disk: %s", templateID, err)
+	for _, namedClient := range clientZipPkgs {
+		if err := local.WriteZip(filepath.Join(pathFrontend, namedClient.id), namedClient.client); err != nil {
+			return fmt.Errorf("unable to save template '%s' to disk: %s", namedClient.id, err)
 		}
-		successfulTemplateWrites = append(successfulTemplateWrites, templateID)
+		successfulTemplateWrites = append(successfulTemplateWrites, namedClient.id)
 	}
 
 	ui.Print(terminal.NewTextLog("Successfully pulled app down: %s", pathRelative))
--- a/internal/commands/pull/command.go
+++ b/internal/commands/pull/command.go
@@ -111,8 +111,8 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		return err
 	}
 
-	// TODO(REALMC-9462): add template_id to /apps payload and pass into resolveClientTemplate to correctly report an
-	//  error when an app has not created with a template
+	// TODO(REALMC-9462): add template_id to /apps payload to remove FindApp within resolveRemoteApp in order to
+	//  determine whether the current app has been created with a template
 	var clientZipPkgs []clientTemplate
 	if app.TemplateID != "" {
 		clientZipPkgs, err = cmd.inputs.resolveClientTemplates(ui, clients.Realm, app.GroupID, app.ID)
--- a/internal/commands/pull/command.go
+++ b/internal/commands/pull/command.go
@@ -111,9 +111,9 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		return err
 	}
 
-	var clientZipPkgs []clientTemplate
+	var clientTemplates []pkg
 	if app.TemplateID != "" {
-		clientZipPkgs, err = cmd.inputs.resolveClientTemplates(ui, clients.Realm, app.GroupID, app.ID)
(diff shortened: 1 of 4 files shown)


internal/commands/pull/inputs.go
```diff
@@ -94,7 +95,18 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
...
 }
 
-func (i *inputs) resolveClientTemplates(ui terminal.UI, realmClient realm.Client, groupID, appID string) (map[string]*zip.Reader, error) {
+type clientTemplate struct {
+	id     string
+	client *zip.Reader
+}
+
+func (i *inputs) resolveClientTemplates(ui terminal.UI, realmClient realm.Client, groupID, appID string) ([]clientTemplate, error) {
+	if proceed, err := ui.Confirm("Would you like to export with a client template?"); err != nil {
```
andrewzlchen, line 104: moved this up so we can avoid making network call if not necessary

jo-tse-mongo (approved):
LGTM pending Unit Tests in Evergreen are fixed! 🎉

internal/commands/pull/inputs.go
makesitgo: this should be reverted.  I like the thought process of making things more efficient, but these commands aren't all that expensive to run in the first place.

in general, its nice to avoid prompts where we can (much like you try to minimize a user's button clicks in a UI).  if the user set `--template` in the original command, its safe to assume the answer to this prompt is always "yes".

also, the extra network call here serves as a quick check that we even know the options available to present to the user, if/when they _do_ select "yes" when wishing to continue

internal/commands/pull/inputs.go
```diff
@@ -147,8 +155,14 @@ func (i *inputs) resolveClientTemplates(ui terminal.UI, realmClient realm.Client
 		} else if !ok {
 			return nil, fmt.Errorf("template '%s' does not have a frontend to pull", templateID)
 		}
-		result[templateID] = templateZip
+		result = append(result, clientTemplate{
+			id:     templateID,
+			client: templateZip,
+		})
 	}
+	sort.SliceStable(result, func(i, j int) bool {
```
makesitgo, line 163: I don't think this is needed ... we already sort on the backend, unless i'm missing something everything should already be deterministic, right?

internal/commands/pull/inputs.go
```diff
@@ -94,7 +95,18 @@ func (i *inputs) resolveRemoteApp(ui terminal.UI, clients cli.Clients) (realm.Ap
 	return app, nil
 }
 
-func (i *inputs) resolveClientTemplates(ui terminal.UI, realmClient realm.Client, groupID, appID string) (map[string]*zip.Reader, error) {
+type clientTemplate struct {
+	id     string
+	client *zip.Reader
```
makesitgo, line 100: [naming nit] not in love with the `client` field of a `clientTemplate` struct... maybe we can call this `pkg` or `zipPkg` or something along those lines?  i think that ultimately reads better above with `namedClient.pkg` (in the for loop)

internal/commands/pull/command.go
```diff
@@ -111,8 +111,7 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		return err
 	}
 
-	// TODO(REALMC-9462): maybe make this less hacky (pass app in to resolveClientTemplates perhaps?)
-	var clientZipPkgs map[string]*zip.Reader
+	var clientZipPkgs []clientTemplate
```
makesitgo, line 114: [opt] `clientTemplates`?

internal/commands/pull/command.go
```diff
@@ -216,12 +215,11 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 	}
 
 	successfulTemplateWrites := make([]string, 0, len(clientZipPkgs))
-	// TODO(REALMC-9452) fix to iterate in a deterministic order
-	for templateID, templateZipPkg := range clientZipPkgs {
-		if err := local.WriteZip(filepath.Join(pathFrontend, templateID), templateZipPkg); err != nil {
-			return fmt.Errorf("unable to save template '%s' to disk: %s", templateID, err)
+	for _, namedClient := range clientZipPkgs {
```
makesitgo, line 218: [opt] `clientTemplate`/`ct`

internal/commands/pull/command_test.go
```diff
@@ -99,7 +99,8 @@ Contents would have been written to: app
 		t.Run("should not write any contents to the destination in a dry run with a template id passed in", func(t *testing.T) {
 			profile := mock.NewProfile(t)
 
-			out, ui := mock.NewUI()
+			out := new(bytes.Buffer)
+			ui := mock.NewUIWithOptions(mock.UIOptions{AutoConfirm: true}, out)
```
makesitgo, line 103: should revert this as well when we move the prompt back

internal/commands/pull/inputs_test.go
```diff
@@ -7,6 +7,8 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/Netflix/go-expect"
```
makesitgo, line 10: there's going to need to be a JIRA to fix the whole codebase before we start adopting this import pattern

ive no strong objections (but also no strong inclination) to do so.  feel free to file it, but otherwise lets keep things consistent for now and move this below the project imports

internal/commands/pull/inputs_test.go
```diff
@@ -179,16 +181,26 @@ func TestPullTemplatesResolve(t *testing.T) {
 	assert.Nil(t, err)
 	defer templateZipPkg2.Close()
 
+	respondToPrompt := func(console *expect.Console, reply string) {
```
makesitgo, line 184: this likely can go away too

makesitgo (changes requested):
we should revert the change to move the prompt up in the code/resolve path... but otherwise the ordering stuff looks good

internal/commands/pull/inputs.go
andrewzlchen: I think this is needed because `result` is populated by iterating through a map and getting the result of `realmClient.ClientTemplate` which returns a singular payload.

The sorted output on the backend is the output of the endpoint that returns (templateId, templateName) pairs that are used to display as options to choose a template from.

makesitgo: how is `result` populated by iterating through a `map`?

`nameOptions` is built by iterating through `CompatiableTemplates` (which is an array) and `selectedTemplateIdxs` is also an array.  i'm probably missing something?

andrewzlchen: ah right, that's true. I saw the for-loop and for some reason assumed it was a map. I'll remove

internal/commands/pull/command_test.go
```diff
@@ -441,14 +440,21 @@ Successfully pulled app down: app
 			profile, teardown := mock.NewProfileFromTmpDir(t, "profile_name")
 			defer teardown()
 
-			_, ui := mock.NewUI()
+			_, console, _, ui, err := mock.NewVT10XConsole()
+			assert.Nil(t, err)
+			defer console.Close()
+			go func() {
+				console.ExpectString("Would you like to export with a client template?")
```
makesitgo, line 447: is this still needed?  if so, we need to make sure we wait on it either through ` doneCh` or `waitgroup`, but I think we just revert this and we're ok right?

internal/commands/pull/command_test.go
```diff
@@ -441,14 +440,21 @@ Successfully pulled app down: app
...
+				console.ExpectEOF()
+			}()
 
 			cmd := &Command{inputs{Project: "some_project", LocalPath: "app", TemplateID: templateID}}
 			assert.Nil(t, cmd.Handler(profile, ui, cli.Clients{Realm: realmClient}))
 
 			destination := filepath.Join(profile.WorkingDirectory, "app", local.FrontendPath, templateID)
 
-			_, err := os.Stat(destination)
+			_, err = os.Stat(destination)
```
makesitgo, line 457: would revert this too if/when we can go back to `_, ui := mock.NewUI()`

internal/commands/pull/inputs.go
makesitgo: ok so I was thinking the other way, with renaming the field to `pkg` and the struct remaining `clientTemplate` ... I don't want to spend too much effort on this though so feel free to leave as-is, but with a name like `pkg` I think I would at least recommend adding a docstring here to help identify what it's doing

internal/commands/pull/inputs_test.go
```diff
@@ -188,7 +188,7 @@ func TestPullTemplatesResolve(t *testing.T) {
 
 		_, ui := mock.NewUI()
 		input := inputs{}
-		_, err := input.resolveClientTemplates(ui, realmClient, "some-group-id", "some-app-id")
+		_, err = input.resolveClientTemplates(ui, realmClient, "some-group-id", "some-app-id")
```
makesitgo, line 191: all of these changes seem to be able to be reverted (they're shadowing `err` from higher-in-scope tests at this point)

makesitgo (approved):
just some cleanup around needed still.... i think i'd go with the name swap too 

```go
type clientTemplate struct {
  id string
  zipPkg *zip.Reader
}
```

but you're call there

internal/commands/pull/inputs.go
andrewzlchen: oh man.. i really need some more coffee today... not sure why i misinterpreted your initial comment but I reverted the last change and updated the `client`  field name to `zipPkg`

internal/commands/pull/command.go
```diff
@@ -111,10 +111,9 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		return err
 	}
 
-	// TODO(REALMC-9462): maybe make this less hacky (pass app in to resolveClientTemplates perhaps?)
```
makesitgo, line 114: can we actually remove this yet?  tbh i forget where we ultimately landed on this one, but the ticket itself is still in backlog

internal/commands/pull/command_test.go
```diff
@@ -100,7 +100,6 @@ Contents would have been written to: app
 			profile := mock.NewProfile(t)
 
 			out, ui := mock.NewUI()
-
```
makesitgo, line 103: probably not a necessary diff here

internal/commands/pull/command.go
andrewzlchen: I removed it from here because I don't think there's anything to do in this location for REALMC-9462.

It seems like we made a temporary app service method and use it here: https://github.com/10gen/realm-cli/blob/1436051fe79d09dc3c6e986562a70ce41ce4b3da/internal/commands/pull/inputs.go#L89

I think once we tackle REALMC-9462 (by adding template id into the /apps payload), we would ideally remove the code in the `resolveRemoteApp` and that'd be it

internal/commands/pull/command_test.go
andrewzlchen: for some reason, adding this space caused the linter to fail: https://spruce.mongodb.com/task/realm_cli_rhel70_lint_patch_0a4628b5d4bc5ae8593504528accb8e1a22369d4_610858c51e2d171708d1876a_21_08_02_20_42_46/logs?execution=0

so i just removed it

PR #186, opened 2021-07-19
REALMC-9684: Add tests to verify each command is capable of compiling and running

makesitgo:
Adding a whole bunch of reviewers, feel free to remove yourself if you'd prefer.  I just want to get as many eyes as possible on the `guidelines.md` for feedback on anything missed or perhaps not worded very clearly


internal/commands/app/diff.go
```diff
@@ -73,7 +73,7 @@ func (cmd *CommandDiff) Flags() []flags.Flag {
 			Value: &cmd.inputs.IncludeHosting,
 			Meta: flags.Meta{
 				Name:      "include-hosting",
-				Shorthand: "h",
```
makesitgo, line 76: this (previous) change was the instigator for this PR... i very casually flipped the shorthand from `-s` to `-h` in [this commit](65101cd5674d741fb0dbc53bf71d97c2beac9e9c) without thinking too much.  of course, `-h` is already reserved as the shorthand for help text, so these commands panic'd when run.

i'm reverting these back (I shouldn't have changed them without expecting to do a major release in the first place 😬 ), and the test lays down some preliminary groundwork for what an "e2e test" might look like for these commands.

the `guidelines.md` is a result from the project retro, where it was mentioned documenting a lot of these command patterns would be beneficial for developers down the line

internal/commands/commands.go
```diff
@@ -191,6 +191,10 @@ var (
 				Command:     &accesslist.CommandCreate{},
 				CommandMeta: accesslist.CommandMetaCreate,
 			},
+			{
```
makesitgo, line 194: cc: @asudhakal ... this was missed in the [previous commit](https://github.com/10gen/realm-cli/commit/e7b26ce2292ced23a1bd7677771dea9df16e65e4) ... feel free to drive-by fix it in your current PR and i'll revert it from here, otherwise I've got it covered

asudhakal: I got it for the current PR!

internal/commands/app/diff.go
UnicodeSnowman: [honest q] do we actually need a shorthand for this (and maybe some other commands)? `-s` isn't really a super intuitive shorthand imo (unless I'm missing something), would it be terrible if we just forced users to provide the full `--include-hosting` command... or just `--hosting`, for example?

makesitgo: im with you, and my answer would be "no" (and likely just avoid shorthands all together until they're explicitly called/requested for).

i just felt like preserving the initially released behavior with `2.0.0` was more preferable than making this a "breaking change" (though i doubt anyone actually uses it this way, currently)

UnicodeSnowman: cool I feel ya. yeah, agreed, let's not force a 3.0 version bump just because we changed a shorthand flag 😆

nathantfrank (approved):
N🧊 !

e2e/cli_test.go
```diff
@@ -0,0 +1,206 @@
...
+			args:        []string{"accesslist", "create"},
+			firstLine:   "Create an IP address or CIDR block in the Access List for your Realm app",
+		},
+		{
+			description: "the accesslist list command",
+			args:        []string{"accesslist", "list"},
+			firstLine:   "List the allowed entries in the Access List of your Realm app",
+		},
+		{
+			description: "the accesslist list command",
```
jandersongo, line 165: should this be "the accesslist **update** command"?

jandersongo (approved):
LGTM with 1 change

e2e/cli_test.go
makesitgo: it most certainly should, thanks for catching!

kpatel71716 (approved):
LGTM! 🥇

PR #187, opened 2021-07-20
REALMC-9696: Force mongodb-atlas as template data source name

andrewzlchen:
- Register new --template-data-source flag
- Use --template-data-source to pick template data source for app creation payload
- Add tests

### DIFF
--- a/internal/commands/app/create.go
+++ b/internal/commands/app/create.go
@@ -99,6 +99,15 @@ func (cmd *CommandCreate) Flags() []flags.Flag {
 				},
 			},
 		},
+		flags.StringFlag{
+			Value: &cmd.inputs.TemplateDataSource,
+			Meta: flags.Meta{
+				Name: flagTemplateDataSource,
+				Usage: flags.Usage{
+					Description: "Specify the data source that you want to initialize your template app with",
+				},
+			},
+		},
 		flags.BoolFlag{
 			Value: &cmd.inputs.DryRun,
 			Meta: flags.Meta{
--- a/internal/commands/app/create.go
+++ b/internal/commands/app/create.go
@@ -1,8 +1,10 @@
 package app
 
 import (
+	"errors"
 	"fmt"
 	"path"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -99,15 +101,6 @@ func (cmd *CommandCreate) Flags() []flags.Flag {
 				},
 			},
 		},
-		flags.StringFlag{
-			Value: &cmd.inputs.TemplateDataSource,
-			Meta: flags.Meta{
-				Name: flagTemplateDataSource,
-				Usage: flags.Usage{
-					Description: "Specify the data source that you want to initialize your template app with",
-				},
-			},
-		},
 		flags.BoolFlag{
 			Value: &cmd.inputs.DryRun,
 			Meta: flags.Meta{
@@ -166,17 +159,13 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 		return err
 	}
 
-	if err := cmd.inputs.resolveTemplateID(ui, clients.Realm); err != nil {
+	if err = cmd.inputs.resolveTemplateID(ui, clients.Realm); err != nil {
 		return err
 	}
 
-	var dsClusters []dataSourceCluster
-	var dsClustersMissing []string
-	if len(cmd.inputs.Clusters) > 0 {
-		dsClusters, dsClustersMissing, err = cmd.inputs.resolveClusters(ui, clients.Atlas, groupID)
-		if err != nil {
-			return err
-		}
+	dsClusters, dsClustersMissing, err := cmd.inputs.resolveClusters(ui, clients.Atlas, groupID)
+	if err != nil {
+		return err
 	}
 
 	var dsDatalakes []dataSourceDatalake
@@ -207,25 +196,29 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 		}
 	}
 
-	// If using a template, we nest backendDir under rootDir and export the
-	// backend code there alongside a sibling directory containing the frontend
-	// code. Otherwise, all code is exported in rootDir
-	backendDir := rootDir
 	if cmd.inputs.Template != "" {
-		backendDir = path.Join(rootDir, local.BackendPath)
+		return cmd.handleCreateTemplateApp(profile, ui, clients, groupID, rootDir, dsClusters, dsDatalakes)
 	}
+	return cmd.handleCreateApp(profile, ui, clients, groupID, rootDir, appRemote, dsClusters, dsDatalakes)
+}
 
+func (cmd CommandCreate) handleCreateApp(
+	profile *user.Profile,
+	ui terminal.UI,
+	clients cli.Clients,
+	groupID string,
+	rootDir string,
+	appRemote realm.App,
+	dsClusters []dataSourceCluster,
+	dsDatalakes []dataSourceDatalake,
+) error {
 	if cmd.inputs.DryRun {
 		logs := make([]terminal.Log, 0, 4)
 		var appCreatedText string
 		if appRemote.GroupID == "" && appRemote.ID == "" {
-			appCreatedText = fmt.Sprintf("A minimal Realm app would be created at %s", backendDir)
+			appCreatedText = fmt.Sprintf("A minimal Realm app would be created at %s", rootDir)
 		} else {
-			appCreatedText = fmt.Sprintf("A Realm app based on the Realm app '%s' would be created at %s", cmd.inputs.RemoteApp, backendDir)
-		}
-
-		if cmd.inputs.Template != "" {
-			appCreatedText = fmt.Sprintf("%s using the '%s' template", appCreatedText, cmd.inputs.Template)
+			appCreatedText = fmt.Sprintf("A Realm app based on the Realm app '%s' would be created at %s", cmd.inputs.RemoteApp, rootDir)
 		}
 
 		logs = append(logs, terminal.NewTextLog(appCreatedText))
@@ -247,16 +240,6 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 		Environment:     cmd.inputs.Environment,
 	}
 
-	// choose a data source to import template app schema data onto
-	if cmd.inputs.Template != "" {
-		initialDataSource, err := cmd.inputs.resolveTemplateDataSource(ui, dsDatalakes, dsClusters)
-		if err != nil {
-			return err
-		}
-		createAppMetadata.Template = cmd.inputs.Template
-		createAppMetadata.DataSource = initialDataSource
-	}
-
 	appRealm, err := clients.Realm.CreateApp(
 		groupID,
 		cmd.inputs.Name,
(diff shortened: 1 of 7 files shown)


UnicodeSnowman (approved):

internal/commands/app/create.go
```diff
@@ -157,17 +159,13 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
 	}
 
-	var dsClusters []dataSourceCluster
-	var dsClustersMissing []string
-	if len(cmd.inputs.Clusters) > 0 {
-		dsClusters, dsClustersMissing, err = cmd.inputs.resolveClusters(ui, clients.Atlas, groupID)
-		if err != nil {
-			return err
-		}
+	dsClusters, dsClustersMissing, err := cmd.inputs.resolveClusters(ui, clients.Atlas, groupID)
```
andrewzlchen, line 166: moved the `if len(cmd.inputs.Clusters) > 0` to inside of `resolveClusters`

it will be applied if `cmd.inputs.Template` is an empty string (aka default app creation)

internal/commands/app/create_inputs.go
```diff
@@ -153,55 +154,110 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
 }
 
 func (i *createInputs) resolveClusters(ui terminal.UI, client atlas.Client, groupID string) ([]dataSourceCluster, []string, error) {
-	clusters, err := client.Clusters(groupID)
-	if err != nil {
-		return nil, nil, err
-	}
+	if i.Template != "" {
```
andrewzlchen, line 157: decided to do some similar "if creating template app vs default app" logic splitting here as well

internal/commands/app/create_test.go
```diff
@@ -605,32 +604,17 @@ Check out your app: cd ./remote-app && realm-cli app describe
...
 		backendFileInfo, err := ioutil.ReadDir(filepath.Join(profile.WorkingDirectory, cmd.inputs.Name, local.BackendPath))
 		assert.Nil(t, err)
-		assert.Equal(t, len(backendFileInfo), 9)
+		assert.Equal(t, len(backendFileInfo), 6)
 
 		frontendFileInfo, err := ioutil.ReadDir(filepath.Join(profile.WorkingDirectory, cmd.inputs.Name, local.FrontendPath, templateID))
 		assert.Nil(t, err)
 		assert.Equal(t, len(frontendFileInfo), 1)
 		assert.Equal(t, frontendFileInfo[0].Name(), "react-native")
-
```
andrewzlchen, line 619: I think this is now somewhat flaky because I added a spinner for template app creation.

I removed these assertions because they are unnecessary for testing whether "should create a new app from a template" is true.

internal/commands/app/new_app_inputs.go
```diff
@@ -94,42 +93,3 @@ func (i *newAppInputs) resolveTemplateID(ui terminal.UI, client realm.Client) er
 
 	return nil
 }
-
-func (i createInputs) resolveInitialTemplateDataSource(ui terminal.UI, dataLakes []dataSourceDatalake, clusters []dataSourceCluster) (interface{}, error) {
```
andrewzlchen, line 98: removed this because we only allow 1 cluster to be created with a template app. Since this is the case, we no longer need a `resolve*` function that chooses a template data source because there is only going to be one choice when it gets to this point.

andrewzlchen:
Ready for another look! 

I re-requested a review because the code has changed quite a bit and it's probably worth re-reviewing these changes

internal/commands/app/create.go
```diff
@@ -354,11 +342,102 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
+	profile *user.Profile,
+	ui terminal.UI,
+	clients cli.Clients,
+	groupID string,
+	rootDir string,
+	dsClusters []dataSourceCluster,
+	dsDatalakes []dataSourceDatalake,
+) error {
+	if cmd.inputs.DryRun {
+		logs := make([]terminal.Log, 0, len(dsClusters)+len(dsDatalakes)+2)
```
UnicodeSnowman, line 355: The `+2` here might deserve a comment and/or explanatorily named variable

internal/commands/app/create_inputs.go
```diff
@@ -153,55 +154,110 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
+		}
+
+		// If creating template app, only allow a 0 or 1 passed-in cluster names
+		switch len(i.Clusters) {
+		case 0, 1:
+		default:
+			return nil, nil, errors.New("template apps can only be created with one data source")
+		}
+
+		if len(clusters) == 0 {
```
UnicodeSnowman, line 170: Do all template apps require a cluster?

UnicodeSnowman (approved):
LGTM

internal/commands/app/create_inputs.go
andrewzlchen: Yes they do. In the future, that may be a different story if there were to be a template that only had functions, for example.

but until that happens, it's just a lot easier to assume that template apps require a cluster

internal/commands/app/create.go
```diff
@@ -354,11 +342,102 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
+		for _, datalake := range dsDatalakes {
+			logs = append(logs, terminal.NewTextLog("The data lake '%s' would be linked as data source '%s'", datalake.Config.DatalakeName, datalake.Name))
+		}
+		logs = append(logs, terminal.NewFollowupLog("To create this app run", cmd.display(true)))
+		ui.Print(logs...)
+		return nil
+	}
+
+	if len(dsClusters) != 1 {
+		return errors.New("could not send app creation request due to unexpected number of clusters")
```
makesitgo, line 374: the error message here will ultimately be prefixed with `[command.display] failed: `... can we change the error message here to `"must specify a cluster when creating app from template"` or something along those lines?

internal/commands/app/create.go
```diff
@@ -354,11 +342,102 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
+		if err != nil {
+			return realm.App{}, err
+		}
+		return appRealm, nil
+	}
+	appRealm, err := createTemplateApp()
+	if err != nil {
+		return err
+	}
+
```
makesitgo, line 404: when showing a spinner, it's good to have a `Log` printed right afterwards to prevent the effect of shifting text. just use the past tense of the spinner text, so its `"Created template app"` here

internal/commands/app/create.go
```diff
@@ -354,11 +342,102 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
+	output := newAppOutputs{
+		AppID:    appRealm.ClientAppID,
+		Filepath: rootDir,
+		Backend:  filepath.Join(rootDir, local.BackendPath),
+		URL:      fmt.Sprintf("%s/groups/%s/apps/%s/dashboard", profile.RealmBaseURL(), appRealm.GroupID, appRealm.ID),
+	}
+
+	frontendFolder := filepath.Join(rootDir, local.FrontendPath)
```
makesitgo, line 417: [opt/ocd] rename to `frontendDir`

internal/commands/app/create.go
```diff
@@ -354,11 +342,102 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
...
+	_, err = os.Stat(frontendFolder)
+	if err != nil && os.IsNotExist(err) {
+		output.Frontends = frontendFolder
+	}
+
+	for _, dsCluster := range dsClusters {
+		output.Clusters = append(output.Clusters, dataSourceOutputs{dsCluster.Name})
```
makesitgo, line 424: instead of ranging over `dsClusters`, let's just append `dataSourceOutputs{output.DataSource.Name}` for now (since we've asserted there's only one cluster already)

internal/commands/app/create_inputs.go
```diff
@@ -153,55 +154,110 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
+		clusters, err := client.Clusters(groupID)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		// If creating template app, only allow a 0 or 1 passed-in cluster names
+		switch len(i.Clusters) {
+		case 0, 1:
+		default:
+			return nil, nil, errors.New("template apps can only be created with one data source")
```
makesitgo, line 167:
```suggestion
			return nil, nil, errors.New("template apps can only be created with one cluster")
```

internal/commands/app/create_inputs.go
```diff
@@ -153,55 +154,110 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
+		}
+
+		if len(clusters) == 0 {
+			// TODO(REALMC-9713): Enable the ability to create a new cluster upon template app creation
+			return nil, nil, errors.New("please create an atlas cluster before creating a template app")
```
makesitgo, line 172:
```suggestion
			return nil, nil, errors.New("please create an Atlas cluster before creating a template app")
```

internal/commands/app/create_inputs.go
```diff
@@ -153,55 +154,110 @@ func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, erro
...
+				Message: "Select a cluster to link to your Realm application:",
+				Options: clusterOptions,
+			}); err != nil {
+				return nil, nil, err
+			}
+		}
+
+		i.Clusters = []string{clusterName}
+		return []dataSourceCluster{{
+			Name: "mongodb-atlas",
```
makesitgo, line 194: might be worth adding this as a `const` to `cloud/realm/services.go`

internal/commands/app/create_inputs_test.go
```diff
@@ -709,6 +709,92 @@ func TestAppCreateInputsResolveCluster(t *testing.T) {
...
+			}, nil
+		}
+		inputs := createInputs{
+			newAppInputs: newAppInputs{
+				Template: "ios.template.todo",
+			},
+			Clusters: clusterNames,
+		}
+		_, _, err := inputs.resolveClusters(ui, ac, "123")
+		assert.Equal(t, "template apps can only be created with one data source", err.Error())
```
makesitgo, line 730: [opt/nit] for here and the other spots if you go the route
```suggestion
		assert.Equal(t, errors.New("template apps can only be created with one data source"), err)
```

internal/commands/app/create_inputs_test.go
```diff
@@ -709,6 +709,92 @@ func TestAppCreateInputsResolveCluster(t *testing.T) {
...
+			return []atlas.Cluster{
+				{ID: "456", Name: clusterNames[0]},
+			}, nil
+		}
+		inputs := createInputs{
+			newAppInputs: newAppInputs{
+				Template: "ios.template.todo",
+			},
+			Clusters:            clusterNames,
+			ClusterServiceNames: []string{"overridden_name"},
```
makesitgo, line 791: just thinking about this now, it would be nice to print a warning message about `"overridden_name"` being ignored when this happens

internal/commands/app/create_test.go
makesitgo: i'd be curious to see if adding that `Log` post-spinner de-flakes these asserts, but otherwise 👍

makesitgo (approved):
nothing blocking from my end, nice work!

🆒  beans

internal/commands/app/create.go
```diff
@@ -157,17 +160,13 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 		return err
 	}
 
-	if err := cmd.inputs.resolveTemplateID(ui, clients.Realm); err != nil {
+	if err = cmd.inputs.resolveTemplateID(ui, clients.Realm); err != nil {
```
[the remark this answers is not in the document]
andrewzlchen: yes we do need it because we would have to fill in the --template flag if it's not already passed in by the user

I have a separate ticket for not automatically prompting for a template.

https://jira.mongodb.org/browse/REALMC-9460

PR #188, opened 2021-07-20
REALMC-9316: Build accesslist delete command

asudhakal:
Implement `accesslist delete` command, which allows the deletion of an allowed IP by specifying the `address`(es).


internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,195 @@
...
+	for _, selection := range selections {
+		allowedIPs = append(allowedIPs, allowedIPsByOption[selection])
+	}
+	return allowedIPs, nil
+}
+
+func tableRowDelete(output deleteAllowedIPOutput, row map[string]interface{}) {
+	deleted := false
+	if output.err != nil {
+		row[headerDetails] = output.err.Error()
```
makesitgo, line 188: related to this line, I'd like you to add something to `terminal/table.go`, in the `parseValue` function:

```go
case error:
  parsed = v.Error()
```

and then here you should be able to avoid using `.Error()`

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,195 @@
...
+		row := map[string]interface{}{
+			headerAddress: output.allowedIP.Address,
+			headerComment: output.allowedIP.Comment,
+		}
+		tableRows = append(tableRows, row)
+		tableRowDelete(output, row)
+	}
+
+	ui.Print(terminal.NewTableLog(
+		fmt.Sprintf("Deleted %d IP address(es) and CIDR block(s)", len(outputs)),
```
makesitgo, line 118:
```suggestion
		fmt.Sprintf("Deleted %d IP address(es) and/or CIDR block(s)", len(outputs)),
```

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,195 @@
...
+		return err
+	}
+
+	selected, err := cmd.inputs.resolveAllowedIP(ui, allowedIPs)
+	if err != nil {
+		return err
+	}
+
+	if len(selected) == 0 {
+		ui.Print(terminal.NewTextLog("No IP addresses or CIDR blocks to delete"))
```
makesitgo, line 93: I would check this one by @kpham1231 first...
```suggestion
		ui.Print(terminal.NewTextLog("No IP addresses nor CIDR blocks to delete"))
```

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,195 @@
...
+		return nil, nil
+	}
+
+	if len(i.Addresses) > 0 {
+		allowedIPsByAddress := make(map[string]realm.AllowedIP, len(allowedIPs))
+		for _, allowedIP := range allowedIPs {
+			allowedIPsByAddress[allowedIP.Address] = allowedIP
+		}
+
+		allowedIPs := make([]realm.AllowedIP, 0, len(i.Addresses))
```
makesitgo, line 144: do we want to account for warning the user for any addresses they may have specified that could not be found?  would probably need to check with the project team if that's worth just a WARNING or true ERROR

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,195 @@
...
+		for _, identifier := range i.Addresses {
+			if allowedIP, ok := allowedIPsByAddress[identifier]; ok {
+				allowedIPs = append(allowedIPs, allowedIP)
+			}
+		}
+
+		if len(allowedIPs) == 0 {
+			return nil, errors.New("unable to find allowed IPs")
```
makesitgo, line 152: and then depending on how you handle the cases where specified addresses aren't found, this check could seemingly go away (and then be handled similarly above in your `Handler` code where you call:

`ui.Print(terminal.NewTextLog("No IP addresses or CIDR blocks to delete"))`

internal/commands/accesslist/delete.go
makesitgo: also, this seems to me better suited as:

```go
return errors.New("No IP addresses nor CIDR blocks to delete")
```

makesitgo (changes requested):
first batch of comments/thoughts/questions for you... i held off on the test file for now

internal/commands/accesslist/delete.go
asudhakal: Would this necessarily be an error though? For example, if the user decided not to "select" any addresses, this would be output as a text log rather than an error. Fine either way though, just want to keep it consistent with how we're handling `secrets` as well (outputting it as a text log).

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	outputs := make(deleteAllowedIPOutputs, len(selected))
+	for i, allowedIP := range selected {
+		err := clients.Realm.AllowedIPDelete(app.GroupID, app.ID, allowedIP.ID)
+		outputs[i] = deleteAllowedIPOutput{allowedIP, err}
+	}
+
+	sort.SliceStable(outputs, func(i, j int) bool {
+		return outputs[i].err != nil && outputs[j].err == nil
```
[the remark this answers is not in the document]
asudhakal: This is how `secrets` are being handled, but I think we can order them based on selection instead. Makes the most sense imo.

internal/commands/accesslist/delete.go
asudhakal: @christineschen

internal/commands/accesslist/delete.go
makesitgo: intuitively, yes, i would think an error (and it was more than likely just oversight, rather than intentional, back in the secrets days).  at the end of the day it's messaging to the user, and I don't think prefixing `"no allowed IPs to delete" with `"accesslist delete failed: "` is the worst thing (for one, why run it if you want nothing to be deleted?  and also, that leaves us with a consistent/single CLI output for the case when there are no secrets to delete, either if none were selected or none were returned from the api, etc)

internal/commands/accesslist/delete.go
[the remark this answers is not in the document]
makesitgo: if its easy enough, making the similar change on `secrets` side would be great too

internal/commands/accesslist/delete.go
kpham1231: > I would check this one by @kpham1231 first...

I would go with the first one ("No IP Addresses or CIDR blocks to delete")—while both are grammatically correct, the second one sounds a bit weird to me for some reason

internal/commands/accesslist/delete.go
asudhakal: talked to Christine and Sumedha, we're going to go with error-ing out if they provide us with an address that doesn't exist in their access list, whether they provide a single address or multiple

makesitgo: sounds good to me... just want to call out that the error should still probably indicate _all_ of the addresses that aren't found, rather than simply error out on the first one

internal/commands/accesslist/delete.go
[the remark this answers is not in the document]
asudhakal: went with the alphanumeric option..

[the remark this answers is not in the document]
makesitgo: out of curiosity what drove that decision?  if its too complex to implement im fine with punting on it for now, but otherwise the more i think about it the more i'm convinced showing the results in the order they were flagged/selected feels the most intuitive here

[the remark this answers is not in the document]
asudhakal: hmm, so actually I think the default is ordering by selection actually, but I haven't looked into how the `&survey.MultiSelect` is inserting them into `selections`.  I can definitely take a deeper look into it. For now I actually removed any sort of sorting/re-ordering the list.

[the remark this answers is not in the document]
makesitgo: cool, lets take that deeper look now and:

* if it defaults to this natural/intuitive sort order, then great, no action needed but let's file a ticket/TODO for `secrets` to operate the same way
* if it doesn't default this way, then still ok, no action needed but let's file a ticket to get both `accesslist` and `secrets` to operate this way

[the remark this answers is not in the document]
asudhakal: https://jira.mongodb.org/browse/REALMC-9795

So the default order for `accesslist` is based on the original list of options (i.e whatever order we gave the user to select from would be maintained after selection as well). I have left the `secrets` selection as it was.

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,190 @@
...
+		return err
+	}
+
+	selectedAllowedIPs, err := cmd.inputs.resolveAllowedIP(ui, allowedIPs)
+	if err != nil {
+		return err
+	}
+
+	if len(selectedAllowedIPs) == 0 {
+		return errors.New("No IP addresses or CIDR blocks to delete")
```
makesitgo, line 93: generally keep capitalization out of error messages in go (unless its a proper noun)
```suggestion
		return errors.New("no IP addresses or CIDR blocks to delete")
```

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,190 @@
...
+		outputs[i] = deleteAllowedIPOutput{allowedIP, err}
+	}
+
+	tableRows := make([]map[string]interface{}, 0, len(outputs))
+	for _, output := range outputs {
+		row := map[string]interface{}{
+			headerAddress: output.allowedIP.Address,
+			headerComment: output.allowedIP.Comment,
+		}
+		deleted := false
```
makesitgo, line 108: [opt style] could perhaps make it slightly more readable here:

```go
var deleted bool
var err error
if output.err != nil {
  err = output.err
} else {
  deleted = true
}

row[headerDeleted] = deleted
row[headerDetails] = err
tableRows = append(tableRows, row)
```

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	if len(i.Addresses) > 0 {
+		allowedIPsByAddress := make(map[string]realm.AllowedIP, len(allowedIPs))
+		for _, allowedIP := range allowedIPs {
+			allowedIPsByAddress[allowedIP.Address] = allowedIP
+		}
+
+		allowedIPs := make([]realm.AllowedIP, 0, len(i.Addresses))
+		notFound := false
```
makesitgo, line 146: i don't think we really need this `bool`, we could just check `len(notFound)`

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,190 @@
...
+		var notFoundList []string
```
makesitgo, line 147: [opt/nit] rename to `notFoundAdresses`/`addressesNotFound`/something along those lines

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,190 @@
...
+		for _, identifier := range i.Addresses {
```
makesitgo, line 148: [opt/nit] rename to `address`/`addr`/something along those lines

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,190 @@
...
+				allowedIPs = append(allowedIPs, allowedIP)
+			} else {
+				notFound = true
+				notFoundList = append(notFoundList, identifier)
+			}
+		}
+
+		if notFound {
+			err := "unable to find IP address(es) and/or CIDR block(s): " + strings.Join(notFoundList, ", ")
+			return nil, errors.New(err)
```
makesitgo, line 159: I'd just inline all of this, if you're worried about line length then something like this would be fine:

```go
return nil, errors.New(
  "unable to find IP address(es) and/or CIDR block(s): " +
    strings.Join(notFoundAddresses, ", "),
)
```

internal/commands/accesslist/delete_test.go
```diff
@@ -0,0 +1,243 @@
...
+				cli.ProjectInputs{projectID, appID, nil},
+				tc.testInput,
+			}}
+
+			out, ui := mock.NewUI()
+
+			assert.Nil(t, cmd.Handler(nil, ui, cli.Clients{Realm: realmClient}))
+			assert.Equal(t, tc.expectedOutput, out.String())
+			assert.Equal(t, "projectID", deleteArgs.groupID)
+			assert.Equal(t, "appID", deleteArgs.appID)
```
makesitgo, line 114: can we assert on the `allowedIPID` thats passed in somehow?

internal/terminal/table.go
```diff
@@ -134,6 +134,8 @@ func parseValue(value interface{}) string {
 		parsed = v
 	case fmt.Stringer:
 		parsed = v.String()
+	case error:
```
makesitgo, line 137: i think there's a few test cases here we could add `error` data to as well

makesitgo (commented):
really just minor nits/questions from me at this point.  should be go after this round, nice work!

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,190 @@
...
+const (
+	headerDeleted = "Deleted"
+	headerDetails = "Details"
+)
+
+var (
+	deleteTableHeaders = []string{headerAddress, headerComment, headerDeleted, headerDetails}
+)
+
+type deleteAllowedIPOutputs []deleteAllowedIPOutput
```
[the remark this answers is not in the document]
asudhakal: we do not!

internal/commands/accesslist/delete_test.go
asudhakal: since we're passing in multiple different `allowedIPID`'s, I wasn't too sure how we could test both with just one assert. An option would be to change the `testInput` to have only one `address`, and have it be the same for both test cases.

makesitgo: I actually ask b/c I want us to explicitly test the fact we pass in multiple ids properly here...

two thoughts:

* each test case could add an `expectedAlowedIPIDs` field to be asserted on
* just break each test out into its own func (it _is_ just a table for two at this point ;) )

asudhakal: great idea! should be good to go now.

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+
+		if len(notFoundAddresses) > 0 {
+			return nil, errors.New(
+				"unable to find IP address(es) and/or CIDR block(s): " +
+					strings.Join(notFoundAddresses, ", "),
+			)
+		}
+
+		return allowedIPs, nil
```
christineschen, line 163: I think this would be more clear if we also named this `selectedAllowedIPs`.  Also I just don't like the feeling of overwriting `allowedIPs` with a new data structure.

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	addressOptions := make([]string, 0, len(allowedIPs))
```
christineschen, line 166: can you add a comment here just to say that this is the case where the user does not include any IPs to delete?

internal/commands/accesslist/delete.go
```diff
@@ -0,0 +1,191 @@
...
+		&selections,
+		&survey.MultiSelect{
+			Message: "Which IP Address(es) or CIDR block(s) would you like to delete?",
+			Options: addressOptions,
+		},
+	); err != nil {
+		return nil, err
+	}
+
+	allowedIPs = make([]realm.AllowedIP, 0, len(selections))
```
christineschen, line 186: same thing about naming here

christineschen (approved):
had a couple suggestions for clarity but otherwise it looks pretty good!

makesitgo (approved):
last few thoughts, but this looks good to go!

🆒  beans

PR #189, opened 2021-07-23
REALMC-9749: Remove -h shorthand flag conflict

andrewzlchen:
This PR removes this error:

`panic: unable to redefine 'h' shorthand in "pull" flagset: it's already used for "include-hosting" flag`

This is caused by having the --include-hosting flag use the `-h` shorthand flag which conflicts with the -h for the `--help` command


tsedgwick (commented):
@makesitgo has this in his pr here: https://github.com/10gen/realm-cli/pull/186#discussion_r672680224

let's work offline to see if we should keep this PR or not

tsedgwick (approved):

PR #191, opened 2021-07-28
REALMC-9460: Prevent template app creation prompt from occurring automatically

andrewzlchen:
## Description

In this PR, I've reworked how the template app creation works within the realm-cli.

We decided offline that we will start out without making a prompt for template ids and instead add the available template ids   to the help menu. We will update the help menu as we add more templates and rethink how we want to prompt for template ids once we find a good solution to REALMC-9512.

~First off, I've updated the --template flag to have a default value when it is declared but is not supplied any arguments.~

~This allows us to distinguish between the following commands:~

~1. `realm-cli app create --template=<template-id>`~

~2. `realm-cli app create --template`~

~3. `realm-cli app create`~

~1 does not prompt the user for a template because it's already supplied~
~2 prompts the user for a template~
~3 does not prompt the user for template because we assume that users do not always want to create a template app.~

~## Updated Help menu~

image

~`<prompt-templates>` is a special value used to distinguish whether we should prompt the user. This is probably going to be the point of contention in this PR since we don't really have this behavior anywhere else in the CLI. I'm happy to try and work this out and iron out the details if this is not acceptable.~

### DIFF
--- a/internal/cloud/realm/templates.go
+++ b/internal/cloud/realm/templates.go
@@ -11,6 +11,19 @@ import (
 	"github.com/10gen/realm-cli/internal/utils/api"
 )
 
+// AllowedTemplates is an array of templates that we know to work. We need to update this in the event we add more
+// templates in the future.
+// This array is used to display the allowed values in the --help menu of the templates flag
+var AllowedTemplates = []string{
+	"web.graphql.todo",
+	"web.mql.todo",
+	"triggers",
+	"ios.swift.todo",
+	"android.kotlin.todo",
+	"react-native.todo",
+	"xamarin.todo",
+}
+
 // Template represents an available Realm app template
 type Template struct {
 	ID   string `json:"id"`
--- a/internal/cloud/realm/templates.go
+++ b/internal/cloud/realm/templates.go
@@ -11,8 +11,7 @@ import (
 	"github.com/10gen/realm-cli/internal/utils/api"
 )
 
-// AllowedTemplates is an array of templates that we know to work. We need to update this in the event we add more
-// templates in the future.
+// AllowedTemplates is an array of templates that we support. We should update this slice whenever we add new templates.
 // This array is used to display the allowed values in the --help menu of the templates flag
 var AllowedTemplates = []string{
 	"web.graphql.todo",
--- a/internal/cloud/realm/templates.go
+++ b/internal/cloud/realm/templates.go
@@ -17,6 +17,7 @@ type Template struct {
 	Name string `json:"name"`
 }
 
+// Templates is a slice of Template structs
 type Templates []Template
 
 // MapByID converts an array of templates into a map whose keys are the template ids and values are the template
--- a/internal/cloud/realm/templates.go
+++ b/internal/cloud/realm/templates.go
@@ -11,7 +11,7 @@ import (
 	"github.com/10gen/realm-cli/internal/utils/api"
 )
 
-// AllowedTemplates is an array of templates that we support. We should update this slice whenever we add new templates.
+// AllowedTemplates is an array of supported templates.
 // This array is used to display the allowed values in the --help menu of the templates flag
 var AllowedTemplates = []string{
 	"web.graphql.todo",
--- a/internal/cloud/realm/templates.go
+++ b/internal/cloud/realm/templates.go
@@ -17,13 +17,24 @@ type Template struct {
 	Name string `json:"name"`
 }
 
+type Templates []Template
+
+// MapByID converts an array of templates into a map whose keys are the template ids and values are the template
+func (templates Templates) MapByID() map[string]Template {
+	out := make(map[string]Template, len(templates))
+	for _, t := range templates {
+		out[t.ID] = t
+	}
+	return out
+}
+
 const (
 	allTemplatesPath               = adminAPI + "/templates"
 	clientTemplatePathPattern      = appPathPattern + "/templates/%s/client"
 	compatibleTemplatesPathPattern = appPathPattern + "/templates"
 )
 
-func (c *client) AllTemplates() ([]Template, error) {
+func (c *client) AllTemplates() (Templates, error) {
 	res, resErr := c.do(
 		http.MethodGet,
 		allTemplatesPath,
@@ -37,7 +48,7 @@ func (c *client) AllTemplates() ([]Template, error) {
 	}
 	defer res.Body.Close()
 
-	var templates []Template
+	var templates Templates
 	if err := json.NewDecoder(res.Body).Decode(&templates); err != nil {
 		return nil, err
 	}
@@ -71,7 +82,7 @@ func (c *client) ClientTemplate(groupID, appID, templateID string) (*zip.Reader,
 	return zipPkg, true, nil
 }
 
-func (c *client) CompatibleTemplates(groupID, appID string) ([]Template, error) {
+func (c *client) CompatibleTemplates(groupID, appID string) (Templates, error) {
 	res, resErr := c.do(http.MethodGet, fmt.Sprintf(compatibleTemplatesPathPattern, groupID, appID), api.RequestOptions{})
 	if resErr != nil {
 		return nil, resErr
@@ -85,7 +96,7 @@ func (c *client) CompatibleTemplates(groupID, appID string) ([]Template, error) 
 	}
 	defer res.Body.Close()
 
-	var templates []Template
+	var templates Templates
 	if err := json.NewDecoder(res.Body).Decode(&templates); err != nil {
 		return nil, err
 	}
--- a/internal/cloud/realm/templates.go
+++ b/internal/cloud/realm/templates.go
@@ -12,7 +12,6 @@ import (
 )
 
 // AllowedTemplates is an array of supported templates.
-// This array is used to display the allowed values in the --help menu of the templates flag
 var AllowedTemplates = []string{
 	"web.graphql.todo",
 	"web.mql.todo",
--- a/internal/commands/app/create.go
+++ b/internal/commands/app/create.go
@@ -20,6 +20,10 @@ import (
 	"github.com/briandowns/spinner"
 )
 
+var (
+	noArgsDefaultValueTemplate = "<prompt-templates>"
+)
+
 // CommandMetaCreate is the command meta for the `app create` command
 var CommandMetaCreate = cli.CommandMeta{
 	Use:         "create",
@@ -97,7 +101,8 @@ func (cmd *CommandCreate) Flags() []flags.Flag {
 			},
 		},
 		flags.StringFlag{
-			Value: &cmd.inputs.Template,
+			Value:              &cmd.inputs.Template,
+			NoArgsDefaultValue: &noArgsDefaultValueTemplate,
 			Meta: flags.Meta{
 				Name: flagTemplate,
 				Usage: flags.Usage{
@@ -201,10 +206,10 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 		}
 	}
 
-	if cmd.inputs.Template != "" {
-		return cmd.handleCreateTemplateApp(profile, ui, clients, groupID, rootDir, dsClusters, dsDatalakes)
+	if cmd.inputs.Template == "" {
+		return cmd.handleCreateApp(profile, ui, clients, groupID, rootDir, appRemote, dsClusters, dsDatalakes)
 	}
-	return cmd.handleCreateApp(profile, ui, clients, groupID, rootDir, appRemote, dsClusters, dsDatalakes)
+	return cmd.handleCreateTemplateApp(profile, ui, clients, groupID, rootDir, dsClusters, dsDatalakes)
 }
 
 func (cmd CommandCreate) handleCreateApp(
--- a/internal/commands/app/create.go
+++ b/internal/commands/app/create.go
@@ -20,10 +20,6 @@ import (
 	"github.com/briandowns/spinner"
 )
 
-var (
-	noArgsDefaultValueTemplate = "<prompt-templates>"
-)
-
 // CommandMetaCreate is the command meta for the `app create` command
 var CommandMetaCreate = cli.CommandMeta{
 	Use:         "create",
@@ -101,13 +97,12 @@ func (cmd *CommandCreate) Flags() []flags.Flag {
 			},
 		},
 		flags.StringFlag{
-			Value:              &cmd.inputs.Template,
-			NoArgsDefaultValue: &noArgsDefaultValueTemplate,
+			Value: &cmd.inputs.Template,
 			Meta: flags.Meta{
 				Name: flagTemplate,
 				Usage: flags.Usage{
 					Description: "Create your Realm app from an available template",
-					Note:        "Manually specifying a template id must be done with a '='. ie: --template=web.graphql.todo",
+					Note:        "Valid templates: web.graphql.todo, web.mql.todo, triggers, ios.swift.todo, android.kotlin.todo, react-native.todo, xamarin.todo",
 				},
 			},
 		},
@@ -169,7 +164,7 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 		return err
 	}
 
-	if err := cmd.inputs.resolveTemplateID(ui, clients.Realm); err != nil {
+	if err := cmd.inputs.resolveTemplateID(clients.Realm); err != nil {
 		return err
 	}
 
--- a/internal/commands/app/create.go
+++ b/internal/commands/app/create.go
@@ -164,7 +164,7 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 		return err
 	}
 
-	if err = cmd.inputs.resolveTemplateID(ui, clients.Realm); err != nil {
+	if err := cmd.inputs.resolveTemplateID(ui, clients.Realm); err != nil {
 		return err
 	}
 
--- a/internal/commands/app/create.go
+++ b/internal/commands/app/create.go
@@ -63,6 +63,7 @@ func (cmd *CommandCreate) Flags() []flags.Flag {
 				Name: flagCluster,
 				Usage: flags.Usage{
 					Description: "Link Atlas cluster(s) to your Realm app",
+					Note:        "Only one cluster can be linked during app creation if creating a template app",
 				},
 			},
 		},
@@ -72,6 +73,7 @@ func (cmd *CommandCreate) Flags() []flags.Flag {
 				Name: flagClusterServiceName,
 				Usage: flags.Usage{
 					Description: "Specify the Realm app Service name to reference your Atlas cluster",
+					Note:        "Service names will be overwritten when creating a template app",
 				},
 			},
 		},
@@ -81,6 +83,7 @@ func (cmd *CommandCreate) Flags() []flags.Flag {
 				Name: flagDatalake,
 				Usage: flags.Usage{
 					Description: "Link Atlas data lake(s) to your Realm app",
+					Note:        "Data lakes cannot be used to create template apps",
 				},
 			},
 		},
@@ -99,6 +102,7 @@ func (cmd *CommandCreate) Flags() []flags.Flag {
 				Name: flagTemplate,
 				Usage: flags.Usage{
 					Description: "Create your Realm app from an available template",
+					Note:        "A cluster must be supplied in order to create a template app",
 				},
 			},
 		},
--- a/internal/commands/app/create.go
+++ b/internal/commands/app/create.go
@@ -101,8 +101,8 @@ func (cmd *CommandCreate) Flags() []flags.Flag {
 			Meta: flags.Meta{
 				Name: flagTemplate,
 				Usage: flags.Usage{
-					Description: "Create your Realm app from an available template",
-					Note:        "Valid templates: web.graphql.todo, web.mql.todo, triggers, ios.swift.todo, android.kotlin.todo, react-native.todo, xamarin.todo",
+					Description:   "Create your Realm app from an available template",
+					AllowedValues: realm.AllowedTemplates,
 				},
 			},
 		},
--- a/internal/commands/app/create.go
+++ b/internal/commands/app/create.go
@@ -107,7 +107,7 @@ func (cmd *CommandCreate) Flags() []flags.Flag {
 				Name: flagTemplate,
 				Usage: flags.Usage{
 					Description: "Create your Realm app from an available template",
-					Note:        "A cluster must be supplied in order to create a template app",
+					Note:        "Manually specifying a template id must be done with a '='. ie: --template=web.graphql.todo",
 				},
 			},
 		},
--- a/internal/commands/app/create_inputs.go
+++ b/internal/commands/app/create_inputs.go
@@ -99,7 +99,7 @@ func (i *createInputs) resolveName(ui terminal.UI, client realm.Client, groupID,
 }
 
 func (i *createInputs) resolveLocalPath(ui terminal.UI, wd string) (string, error) {
-	//check if we are in an app directory already
+	// Check if current working directory is the app directory
 	_, appOK, err := local.FindApp(wd)
 	if err != nil {
 		return "", err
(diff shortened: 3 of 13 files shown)


internal/commands/app/create.go
```diff
@@ -20,6 +20,10 @@ import (
 	"github.com/briandowns/spinner"
 )
 
+var (
+	noArgsDefaultValueTemplate = "<prompt-templates>"
```
andrewzlchen, line 24: this is the default value of the --template flag when a user declares it without any args

internal/commands/app/create.go
```diff
@@ -94,11 +101,13 @@ func (cmd *CommandCreate) Flags() []flags.Flag {
...
 		},
 		flags.StringFlag{
-			Value: &cmd.inputs.Template,
+			Value:              &cmd.inputs.Template,
+			NoArgsDefaultValue: &noArgsDefaultValueTemplate,
 			Meta: flags.Meta{
 				Name: flagTemplate,
 				Usage: flags.Usage{
 					Description: "Create your Realm app from an available template",
+					Note:        "Manually specifying a template id must be done with a '='. ie: --template=web.graphql.todo",
```
andrewzlchen, line 110: this is the limitation of using `NoArgsDefaultValue`

https://github.com/spf13/pflag/issues/134

In practice, I think a relatively few amount of people would encounter this, but I'm hoping that this note is enough to cover us for the portion of users that know the template id and would like to specify them manually

internal/commands/app/create.go
```diff
@@ -160,7 +169,7 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 		return err
 	}
 
-	if err = cmd.inputs.resolveTemplateID(ui, clients.Realm); err != nil {
+	if err := cmd.inputs.resolveTemplateID(ui, clients.Realm); err != nil {
```
andrewzlchen, line 172: driveby to control variable scoping to avoid weird bugs

internal/commands/app/create_inputs.go
```diff
@@ -322,7 +329,11 @@ func (i createInputs) args(omitDryRun bool) []flags.Arg {
 		args = append(args, flags.Arg{flagLocalPathCreate, i.LocalPath})
 	}
 	if i.Template != "" {
-		args = append(args, flags.Arg{flagTemplate, i.Template})
+		if i.Template == noArgsDefaultValueTemplate {
+			args = append(args, flags.Arg{Name: flagTemplate})
```
andrewzlchen, line 333: remove `<prompt-template>` when we display the command if a user declared --template with no args

internal/commands/app/create.go
UnicodeSnowman: ahh this is in fact a little awkward, but I do agree that users are not as likely to specify a template explicitly. what happens if they provide `--template web.graphql.todo`? do they get prompted to select a template? or does it complain about not knowing what `web.graphql.todo` means?

andrewzlchen: The current behavior is that it ignores `web.graphql.todo` and prompts for a template id.

makesitgo: im confused by this as well, why does it ignore `web.graphql.todo`?  by ignore im guessing that means `inputs.Template` is blank still?

andrewzlchen: no the value of `inputs.Template` will be the default no-args value which is `<prompt-templates>`

in its current state,

`realm-cli app create --template web.graphql.todo`

is the same as

`realm-cli app create web.graphql.todo --template`

andrewzlchen: it seems like we don't process any keywords after `create` for the `realm-cli app` command so `web.graphql.todo` doesn't really do anything if you do `realm-cli app create --template web.graphql.todo`

makesitgo: alright managed to get to the bottom of this offline, I missed what `NoArgsDefaultValue` was entailing at first.  I think we can ultimately avoid this awkwardness while still achieving the same functionality by tackling [REALMC-9512](https://jira.mongodb.org/browse/REALMC-9512) first.

spoke offline with andrew and he said he's happy to take on that ticket first

UnicodeSnowman: oh wow, interesting... that is indeed a bit hacky/funky, but it ends up working out okay.

andrewzlchen:
blocked on REALMC-9512 for now. we're trying to figure out what our best course of action for providing a way to create a flag with optional arguments

internal/commands/app/create.go
```diff
@@ -72,6 +73,7 @@ func (cmd *CommandCreate) Flags() []flags.Flag {
 				Name: flagClusterServiceName,
 				Usage: flags.Usage{
 					Description: "Specify the Realm app Service name to reference your Atlas cluster",
+					Note:        "Service names will be overwritten when creating a template app",
```
UnicodeSnowman, line 76: [q] I didn't catch/think about this until now, but should we stop referring to services here? should this just be "Data Source names..." and so on?

UnicodeSnowman: I feel like we can just change references to "service" here, I wouldn't consider that a breaking change... however, we probably _do_ need to leave the `cluster-service-name` flag/arg for now (to avoid a breaking change) and just add an additional `data-source` arg.

cc @makesitgo

UnicodeSnowman (approved):

internal/commands/app/create.go
makesitgo: Agreed that changing the "note" text should be fine.  Do you mind elaborating on what the extra `data-source` arg would be?  Or are you suggesting replacing `cluster-service-name` with `cluster-data-source-name`?

Whatever we do with this one, there's an analogous `datalake-service-name` flag we should consider as well.  And if we're in the mood for breaking changes, maybe we remove the support for "multiple data sources at once", which would then eliminate the `*-service-name` flags all together (we could align with templates and go back to forcing `mongodb-atlas` as the name).

andrewzlchen: hmm I could be wrong, but maybe Chris is suggesting making 2 flags do the same thing: `--cluster-service-name` and `--cluster-data-source-name`. I'm hoping there's a way to alias flags so we don't need to manually account for that in our code if we end up going this route.

That said, I think this is a good opportunity to audit the CLI, Frontend (and maybe backend) to find wherever we still refer to data sources or HTTP endpoints as services and change them accordingly.

I can make a new ticket for this if we want to do this

andrewzlchen: https://jira.mongodb.org/browse/REALMC-9950

internal/commands/app/create_inputs.go
```diff
@@ -174,7 +174,16 @@ func (i *createInputs) resolveClusters(ui terminal.UI, client atlas.Client, grou
...
 		var clusterName string
 		if len(i.Clusters) == 1 {
-			clusterName = i.Clusters[0]
+			for _, c := range clusters {
+				if c.Name == i.Clusters[0] {
+					clusterName = i.Clusters[0]
+				}
+			}
+			if clusterName == "" {
+				return nil, nil, fmt.Errorf("could not find atlas cluster '%s'", i.Clusters[0])
```
makesitgo, line 183:
```suggestion
				return nil, nil, fmt.Errorf("could not find Atlas cluster '%s'", i.Clusters[0])
```

internal/commands/app/create_inputs.go
```diff
@@ -174,7 +174,16 @@ func (i *createInputs) resolveClusters(ui terminal.UI, client atlas.Client, grou
...
+			}
+		} else if len(clusters) == 1 {
```
makesitgo, line 185: do we have a test case for this already?  can we add one if not?

internal/commands/app/create_test.go
```diff
@@ -335,75 +330,59 @@ Check out your app: cd ./test-app && realm-cli app describe
...
-			return &clientZipPkg.Reader, true, nil
-		}
-		ac := mock.AtlasClient{}
-		ac.GroupsFn = func() ([]atlas.Group, error) {
-			return []atlas.Group{{ID: "123"}}, nil
-		}
-		ac.ClustersFn = func(groupID string) ([]atlas.Cluster, error) {
-			return []atlas.Cluster{{Name: "test-cluster"}}, nil
+
+		client := mock.RealmClient{
```
makesitgo, line 343: [nit] can we rename this to `realmClient` and define `atlasClient` similarly alongside here (rather than inline to the `cmd.Handler` call below)?

internal/commands/app/create_test.go
```diff
@@ -335,75 +330,59 @@ Check out your app: cd ./test-app && realm-cli app describe
...
 				DeploymentModel: realm.DeploymentModelGlobal,
-				ConfigVersion:   realm.DefaultAppConfigVersion,
 			},
 			Clusters: []string{"test-cluster"},
 		}}
 
-		assert.Nil(t, cmd.Handler(profile, ui, cli.Clients{Realm: rc, Atlas: ac}))
-
-		console.Tty().Close() // flush the writers
-		<-doneCh              // wait for procedure to complete
```
makesitgo, line 400: is this test no longer interactive (i.e. is `procedure` needed at all)?

if its no longer interactive, we should switch the `ui` to just be `mock.NewUI` (rather than the `V10XConsole` variant), as well as this bit:

```go
		doneCh := make(chan (struct{}))
		go func() {
			defer close(doneCh)
			procedure(console)
		}()
```

if it _is_ interactive still, then we definitely should still keep the `<-doneCh` and psuedo-terminal `Close` calls

internal/commands/app/create_test.go
```diff
@@ -335,75 +330,59 @@ Check out your app: cd ./test-app && realm-cli app describe
...
 
-		path := filepath.Join(profile.WorkingDirectory, cmd.inputs.Name, local.BackendPath)
-		appLocal, err := local.LoadApp(path)
+		_, err = local.LoadApp(filepath.Join(profile.WorkingDirectory, cmd.inputs.Name))
 		assert.Nil(t, err)
 
-		assert.Equal(t, appLocal.RootDir, path)
+		backendFileInfo, err := ioutil.ReadDir(filepath.Join(profile.WorkingDirectory, cmd.inputs.Name))
+		assert.Nil(t, err)
+		assert.Equal(t, len(backendFileInfo), 10)
```
makesitgo, line 385: was this an existing assert?  do you mind at least providing some clarity into where the `10` comes from?  (and/or is there a better/more specific assert we can use instead of this?)

internal/commands/app/new_app_inputs.go
```diff
@@ -38,22 +36,21 @@ func (i *newAppInputs) resolveRemoteApp(ui terminal.UI, rc realm.Client) (realm.
...
 
-func (i *newAppInputs) resolveTemplateID(ui terminal.UI, client realm.Client) error {
-	if i.Template == "" && ui.AutoConfirm() {
+// resolveTemplateID is responsible for resolving a template id from a user cli request.
+// a user can use the --template flag in 1 of the following ways:
+// 1. --template=<template-id>: This populates the value of i.Template to be <template-id> and attempts to create a
+//    template with id = <template-id>
+// 2. no flag defined: This populates the value of i.Template to be "". This tells us to skip the template-app-related
+//    parts of app creation and create a default app
+func (i *newAppInputs) resolveTemplateID(client realm.Client) error {
```
makesitgo, line 45: I might simplify this docstring a bit (I don't think we need much explanation here about the way flags work/set values)

```go
// resolveTemplateID is responsible for resolving a template id from a user cli request.
// If the --template flag is not set, the CLI will create a default app without any template,
// otherwise the CLI will attempt to create an app based on the specified template
func (i *newAppInputs) resolveTemplateID(client realm.Client) error {
```

internal/commands/pull/command.go
```diff
@@ -84,13 +84,14 @@ func (cmd *Command) Flags() []flags.Flag {
...
-			Value: &cmd.inputs.TemplateID,
+		flags.StringSliceFlag{
+			Value: &cmd.inputs.TemplateIDs,
 			Meta: flags.Meta{
 				Name:      "template",
 				Shorthand: "t",
 				Usage: flags.Usage{
-					Description: "Specify the Template ID that is used for this Realm app",
+					Description: "Specify the IDs of the template frontend to export. Specified templates must be compatible with the remote app",
+					Note:        "Options include: web.graphql.todo, web.mql.todo, triggers, ios.swift.todo, android.kotlin.todo, react-native.todo, xamarin.todo",
```
makesitgo, line 94: similar not here about using `AllowedValues` instead of `Note` for this data

internal/commands/pull/command.go
```diff
@@ -84,13 +84,14 @@ func (cmd *Command) Flags() []flags.Flag {
...
-		flags.StringFlag{
-			Value: &cmd.inputs.TemplateID,
+		flags.StringSliceFlag{
+			Value: &cmd.inputs.TemplateIDs,
 			Meta: flags.Meta{
 				Name:      "template",
 				Shorthand: "t",
 				Usage: flags.Usage{
-					Description: "Specify the Template ID that is used for this Realm app",
+					Description: "Specify the IDs of the template frontend to export. Specified templates must be compatible with the remote app",
```
makesitgo, line 93:
```suggestion
					Description: "Specify the frontend Template ID(s) to export",
					Note: "Specified templates must be compatible with the remote app",
```

internal/commands/pull/command_test.go
```diff
@@ -478,18 +478,6 @@ Successfully pulled app down: app
 			assert.Nil(t, err)
 			defer console.Close()
 
-			doneCh := make(chan struct{})
```
makesitgo, line 481: if this test is no longer interactive, then let's move the `ui` back to `mock.NewUI` (rather than the `V10X` variant)

internal/commands/pull/inputs.go
```diff
@@ -27,7 +26,7 @@ type inputs struct {
 	IncludeDependencies bool
 	IncludeHosting      bool
 	DryRun              bool
-	TemplateID          string
+	TemplateIDs         []string
```
makesitgo, line 29: [question] did Templates want to be an array input for `app create` similar to how it is here?

internal/commands/pull/inputs_test.go
```diff
@@ -179,16 +179,24 @@ func TestPullTemplatesResolve(t *testing.T) {
 	assert.Nil(t, err)
 	defer templateZipPkg2.Close()
 
+	t.Run("should not do anything if no template id is passed in", func(t *testing.T) {
+		var realmClient mock.RealmClient
+
+		input := inputs{}
+		out, err := input.resolveClientTemplates(realmClient, "some-group-id", "some-app-id")
+		assert.Equal(t, nil, err)
```
makesitgo, line 187: use `assert.Nil` instead

makesitgo (commented):
nothing blocking from me, but just some nits/minor things

internal/commands/app/create.go
andrewzlchen: Hmm, I'm not sure if this is a valid concern, but I think that depending on the solution we end up with in REALMC-9950, we may want to leave this flag alone.

I'd rather just keep the descriptions aligned with their flag name until we know for sure that the change is needed.

internal/commands/app/create.go
```diff
@@ -99,6 +102,7 @@ func (cmd *CommandCreate) Flags() []flags.Flag {
 				Name: flagTemplate,
 				Usage: flags.Usage{
 					Description: "Create your Realm app from an available template",
+					Note:        "Valid templates: web.graphql.todo, web.mql.todo, triggers, ios.swift.todo, android.kotlin.todo, react-native.todo, xamarin.todo",
```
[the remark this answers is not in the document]
andrewzlchen: ooo that's a good idea. I'll do that instead!

internal/commands/app/create.go
makesitgo: no strong opinions on my end, I'd maybe just ask sumedha/drew if they want this quick update or not

internal/commands/app/create.go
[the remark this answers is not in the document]
makesitgo: btw I'm all ears if you have any thoughts on how to make these other fields more discoverable... slash how to make the `Meta`/`Usage` more fool-proof to use

internal/commands/app/create_inputs.go
andrewzlchen: I've grouped these template app creation tests and added two new test cases:

1. For this comment: for whether a cluster will be automatically selected if user does not supply cluster and there is only one option returned from the atlas clusters endpoint
2. For whether a prompt appears if user does not supply cluster and there are multiple options returned from the atlas clusters endpoint

internal/commands/app/create_test.go
andrewzlchen: ah yeah, that was an existing change that I updated. This 10 comes from the mock backend zip package which is the contents of `testdata/project.zip`.

This test serves to ensure that when a user does not supply --template, a default app will be created.

the assertion that the resulting directory has len = 10 is just to ensure that the result is the same as the result of the mocked export function.

maybe we can unzip this `testdata/project.zip` first and use this as the assertion value

ie:
```go
backendZipPkg, err := zip.OpenReader("testdata/project.zip")
assert.Nil(t, err)

// add this
expectedBackendZipLen := len(backendZipPkg.File)

realmClient := mock.RealmClient{
    ExportFn: func(groupID, appID string, req realm.ExportRequest) (string, *zip.Reader, error) {
        return "", &backendZipPkg.Reader, nil
    },
    // ...
}

// ...

// and use the new expected length here
assert.Equal(t, len(backendFileInfo), expectedBackendZipLen)
```

internal/commands/app/create_test.go
andrewzlchen: i suppose it's because the structure of the test cases are similar, but the diff is trying to compare a deleted test case with a different one altogether

internal/commands/app/create_test.go
andrewzlchen: edit nevermind, we cant do this. the zip file also lists out nested files so the number of files in the zip file != the number of `backendFileInfo` files.

I think it's probably sufficient to assert that:

1. we do not create a `backend` and `frontend` directory in the top level
2. and that realm_config.json exists in the top level to denote that this directory is a realm app folder

I'll do this for now, but I'm happy to change this if you have any opinions here

internal/commands/pull/inputs.go
andrewzlchen: User Story:

When you create a template, you choose a single template, but when you pull, you should have an option of what frontend templates you can create. I realize that this may be a bit confusing. Would calling this flag --template-client be better?

Explanation:

I think it was an oversight that this was just a string before, so I fixed it here. 

I didn't realize that this was a string until I tried to pull multiple templates without the prompt.

The prompt from before allowed users to pull multiple templates which I think we should maintain feature parity for.

internal/commands/app/create_inputs_test.go
```diff
@@ -710,90 +710,166 @@ func TestAppCreateInputsResolveCluster(t *testing.T) {
 		assert.Equal(t, "123", expectedGroupID)
 	})
 
-	t.Run("should error when more than one cluster passed in when creating template app", func(t *testing.T) {
-		_, ui := mock.NewUI()
-		clusterNames := []string{"Cluster0", "Cluster1"}
+	t.Run("when creating template apps", func(t *testing.T) {
```
andrewzlchen, line 713: i decided to group all of the template app creation tests together to tidy them up a bit

andrewzlchen: hiding whitespace changes will make reading through this significantly easier

internal/commands/app/create_test.go
```diff
@@ -335,75 +330,78 @@ Check out your app: cd ./test-app && realm-cli app describe
 			procedure(console)
 		}()
 
-		var createdApp realm.App
-		rc := mock.RealmClient{}
-
-		zipPkg, err := zip.OpenReader("testdata/project.zip")
-		assert.Nil(t, err)
-		defer zipPkg.Close()
```
makesitgo, line 343: looks like we dropped the `defer Close()` on `defaultBackendZipPkg` (the same I think goes for `frontendZipPkg` in this file too?)

internal/commands/app/create_test.go
```diff
@@ -335,75 +330,78 @@ Check out your app: cd ./test-app && realm-cli app describe
...
-		assert.Equal(t, appLocal.RootDir, path)
+		// we expect for the command to have created a default app. we will assert that realm_config.json exists and that
+		// backend/ and frontend/ do not exist in the directory.
+		var containsRealmConfig bool
+		var containsTemplateDirChild bool
+		for _, fi := range backendFileInfo {
+			if strings.Contains(fi.Name(), local.NameRealmConfig) {
+				containsRealmConfig = true
+			}
+			if fi.Name() == local.BackendPath || fi.Name() == local.FrontendPath {
```
makesitgo, line 395: is it worth just using the same style of `strings.Contains` checks for these?

internal/commands/pull/command.go
```diff
@@ -84,13 +84,14 @@ func (cmd *Command) Flags() []flags.Flag {
...
-		flags.StringFlag{
-			Value: &cmd.inputs.TemplateID,
+		flags.StringSliceFlag{
+			Value: &cmd.inputs.TemplateIDs,
 			Meta: flags.Meta{
 				Name:      "template",
 				Shorthand: "t",
 				Usage: flags.Usage{
-					Description: "Specify the Template ID that is used for this Realm app",
+					Description:   "Specify the frontend Template ID(s) to export. Specified templates must be compatible with the remote app",
```
makesitgo, line 93: what are your thoughts on making the second sentence here a `Note`?  there's not a _ton_ of description/flag usages with periods/multiple sentences yet

makesitgo (approved):
last few bits here, but looks good.  nice work!

🆒  beans

internal/cloud/realm/templates.go
```diff
@@ -11,19 +11,43 @@ import (
 	"github.com/10gen/realm-cli/internal/utils/api"
 )
 
+// AllowedTemplates is an array of templates that we support. We should update this slice whenever we add new templates.
```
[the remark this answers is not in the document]
andrewzlchen: just to confirm, is the 2nd line okay?

`// This array is used to display the allowed values in the --help menu of the templates flag`

internal/commands/app/create_test.go
andrewzlchen: Hmm, I think I'd rather make the realm_config.json be an exact match than keeping the `strings.Contains` just so the assertion is as strict as possible.

I'll do that now but if you feel differently, I can change it

internal/cloud/realm/templates.go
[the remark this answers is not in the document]
makesitgo: i'd remove that too, again not super important I just don't think it's all that necessary

internal/commands/pull/command.go
andrewzlchen: I'm okay with that but the only this is that the `Note` is not on it's own line.

It seems to be displayed after the `AllowedValues` so it may potentially be lost since the allowed values list is kinda long and not super human readable


image

internal/commands/app/create_test.go
makesitgo: the exact match on `realm_config.json` is fine by me since ultimately the assert here is that `false` went to `true` (b/c this condition hit)

what worries me is the opposite style of assert for making sure we don't have `backend` or `frontend` directories.  if for w/e reason the `fi.Name()` started having a trailing slash or any other characters in it, this test wouldn't catch that (b/c `false` would have remained `false`).  that's why I'm wondering if it makes sense to at least use the more permissive `strings.Contains` check for this frontend/backend logic.

[question] instead of using `ioutil.ReadDir` on the whole project directory and iterating over its files, why don't we just explicitly look for `realm_config.json` (& verify it exists) and `backend`/`frontend` directories (& verify they don't exist)?

internal/commands/pull/command.go
kpham1231: Having the Note before the Defaults makes sense to me! Given the defaults can get pretty long, getting the Note across first would be good

internal/commands/app/create_test.go
andrewzlchen: oh i see what you mean. That's true.

I think your suggestion here is good. I'll go ahead and do that

PR #192, opened 2021-07-30
REALMC 9512

andrewzlchen:
## Description

So this is still in the WIP / POC stage

The issue at the moment is that we need to figure out if it's possible to not call `flag.Set()` if the user did not supply the flag that we're concerned with.

I've tried to find references of `flag.Set` within our codebase and it seems like it's only being called from within the pflag lib. This probably means that we either need to maintain our own fork of `pflag` (highly yikes) or we stick to `NoOptDefVal`

As a manual test, I've implemented the `app create --template` flag with this new `OptionalStringFlag` and I'm finding that it behaves similarly as if we used the `StringFlag` with a `DefaultValue: "<prompt-templates>"`.

in other words:

when user calls `realm-cli app create` => `cmd.input.Template.String() = <prompt-template>`

and when user calls `realm-cli app create --template` => error: 2021/07/30 16:28:33 flag needs an argument: --template

image

### DIFF
--- a/internal/commands/app/create.go
+++ b/internal/commands/app/create.go
@@ -40,6 +40,17 @@ type CommandCreate struct {
 	inputs createInputs
 }
 
+// NewCommandCreate returns a default uninitialized CommandCreate
+func NewCommandCreate() *CommandCreate {
+	return &CommandCreate{
+		inputs: createInputs{
+			newAppInputs: newAppInputs{
+				Template: flags.OptionalString{DefaultValue: "<prompt-templates>"},
+			},
+		},
+	}
+}
+
 // Flags is the command flags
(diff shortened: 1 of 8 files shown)


makesitgo:
the `flag needs an argument` error is really an unfortunate one... although im sure it covers some sort of edge-case I can't think of.

i think we need to find a different approach than using `NoOptDefVal` though.  having flags in the CLI behave differently than others (i.e. needing the use of `=` or not) is a UX nightmare.

happy to hop on a call with product to discuss what's going on here too.  i'm officially back to leaning towards us having dynamically created flags (i.e. `--template` remains the string flag, a user could specify `--use-template` or some other programmatically determined name for the equivalent bool flag)

PR #195, opened 2021-08-04
REALMC-9689: change UPDATE to PATCH in Access List

asudhakal:
Corresponding backend change: https://github.com/10gen/baas/pull/4658

### DIFF
--- a/internal/cloud/realm/access_list.go
+++ b/internal/cloud/realm/access_list.go
@@ -83,10 +83,9 @@ func (c *client) AllowedIPCreate(groupID, appID, address, comment string, useCur
 	return allowedIP, nil
 }
 
-// TODO(REALMC-9689): Update to PATCH
 func (c *client) AllowedIPUpdate(groupID, appID, allowedIPID, newAddress, newComment string) error {
 	res, err := c.doJSON(
-		http.MethodPut,
+		http.MethodPatch,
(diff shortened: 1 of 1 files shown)


internal/cloud/realm/access_list.go
```diff
@@ -83,10 +83,9 @@ func (c *client) AllowedIPCreate(groupID, appID, address, comment string, useCur
 	return allowedIP, nil
 }
 
-// TODO(REALMC-9689): Update to PATCH
 func (c *client) AllowedIPUpdate(groupID, appID, allowedIPID, newAddress, newComment string) error {
 	res, err := c.doJSON(
-		http.MethodPut,
+		http.MethodPatch,
```
makesitgo, line 88: is there no corresponding test case that would need to be updated for this change?

makesitgo (approved):
one question about testing, im guessing they're just not implemented yet?  otherwise...

🆒  beans

internal/cloud/realm/access_list.go
asudhakal: they are implemented but are skipped for now!

makesitgo: ah thats totally my bad, I thought we were in `baas` 🙄 ... carry on

jandersongo (approved):

PR #196, opened 2021-08-05
REALMC-9934: Include error message in Segment event properties

makesitgo:


internal/telemetry/event.go
```diff
@@ -34,16 +34,23 @@ const (
...
+	eventDataKeyVersion      = "v"
+	eventDataKeyError        = "err"
+	eventDataKeyErrorMessage = "err_msg"
 
 	// EventDataKeyTemplate used to tracked if templates were used to create an app
 	EventDataKeyTemplate = "templateId"
 )
 
+func EventDataError(err error) []EventData {
+	return []EventData{
```
tsedgwick, line 48: should we check `if err != nil`? This will panic otherwise

tsedgwick (approved):
one question - otherwise LGTM

also, the PR title should be REALMC-9934 (missing the C)

internal/telemetry/event.go
makesitgo: the usages of this function are already wrapped in `!= nil`, but the better-safe-than-sorry approach seems fine here

makesitgo:
was able to confirm with sumedha this is working as expected now, it's good to go in whatever next release we have

PR #207, opened 2021-09-16
REALMC-10273: Add logforwarder to AppDescription

LijieZhang1998:
Implementation includes:
1. Add logforwardersummary to app_description.go
2. Add unit test for it

Question:
Integration is failing cause I think we don't have baas side change merged yet. I ran the test on my local and it works fine. Should we not merge this change until baas change is merged?

### DIFF
--- a/internal/cloud/realm/app_description.go
+++ b/internal/cloud/realm/app_description.go
@@ -88,6 +88,7 @@ type EventSubscriptionSummary struct {
 	Enabled bool   `json:"enabled"`
 }
 
+// LogForwarderSummary is a short summary for a log forwarder model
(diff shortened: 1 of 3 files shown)


nathantfrank:
> Implementation includes:
> 
> 1. Add logforwardersummary to app_description.go
> 2. Add unit test for it
> 
> Question:
> Integration is failing cause I think we don't have baas side change merged yet. I ran the test on my local and it works fine. Should we not merge this change until baas change is merged?

Yes, let's do the baas change first, and then quickly follow up with this one.

nathantfrank (approved):
LGTM, let's make sure tests are passing!

PR #209, opened 2021-09-17
REALMC-10247: Discard orphaned drafts after push fails to complete or deploy

makesitgo:


internal/commands/push/command_test.go
```diff
@@ -437,14 +581,19 @@ An error occurred while uploading hosting assets: failed to remove /deleteme.htm
 
 				err := cmd.Handler(profile, ui, cli.Clients{Realm: realmClient})
 				assert.Equal(t, errors.New("2 error(s) occurred while importing hosting assets"), err)
-				assert.Equal(t, `Determining changes
```
makesitgo, line 440: ⚠️  drive-by here: the `404.html` and `index.html` lines output were non-deterministic (we run those actions asynchronously)... so loosened this assertion a bit to avoid the flakes

internal/commands/push/command.go
```diff
@@ -568,3 +567,9 @@ func deployDraftAndWait(ui terminal.UI, realmClient realm.Client, remote appRemo
 	ui.Print(terminal.NewTextLog("Deployment complete"))
 	return nil
 }
+
+func discardDraft(ui terminal.UI, realmClient realm.Client, remote appRemote, draftID string) {
+	if err := realmClient.DiscardDraft(remote.GroupID, remote.AppID, draftID); err != nil {
+		ui.Print(terminal.NewWarningLog("Failed to discard the draft created for your deployment"))
+	}
```
tsedgwick, line 574: I'd rather have this inline instead of a void func so there's less 🪄

internal/commands/push/command.go
```diff
@@ -546,9 +548,6 @@ func deployDraftAndWait(ui terminal.UI, realmClient realm.Client, remote appRemo
 
 			deployment, err = realmClient.Deployment(remote.GroupID, remote.AppID, deployment.ID)
 			if err != nil {
-				if e := realmClient.DiscardDraft(remote.GroupID, remote.AppID, draftID); e != nil {
```
tsedgwick, line 549: why isn't this needed any longer?

tsedgwick (approved):

internal/commands/push/command.go
makesitgo: 🪄 ... the `err` returned from `deployDraftAndWait` brings us to the newly-created, your-not-so-favorite, `discardDraft` helper function

internal/commands/push/command.go
makesitgo: you've beaten me down pretty good on avoiding simple helpers, but I tend to go back to those ways when user-facing messaging is in play (as consistent messaging should be considered a key part of the CLI UX)

fwiw, i envisioned this method potentially needing to scale further with retry logic if this still was an issue for users at some point.  but as a compromise, I can inline the calls for now with hard-coded `var` for making sure the `terminal.NewWarningLog` remains consistent

internal/commands/push/command.go
makesitgo: helper's gone, discarding draft happens in line, but this still remains true: the `err` (from the line above here) is returned from the anonymous `waitForDeployment` func defined within `deployDraftAndWait`, which is then promptly returned and will trigger the `Handler` to discard the draft at that point (which allows us to account for completed-but-failed deployments)

tsedgwick: :shipit:

PR #211, opened 2021-09-22
REALMC-10294: Fix regression to allow uploading dependencies from node_modules directory

makesitgo:
This re-introduces the necessary code to support uploading a `node_modules` directory

We lost this ability in the PR: #178 

As an aside, we were previously just grabbing the first match returned from `filepath.Globs` when matching on `node_modules*` (which is how we initially started supporting directory uploads in the first place).  Part of this change involves a more deterministic approach to finding that `node_modules` archive, by sequentially checking the archive extensions in order of preferred format.  If none match, we fall back to the directory approach


internal/local/archive.go
```diff
@@ -19,143 +16,77 @@ const (
...
-	if err != nil {
-		return nil, err
-	}
-
-	zr, err := zip.NewReader(f, fi.Size())
-	if err != nil {
-		return nil, err
+var (
+	preferredArchiveMatchers = []archiveMatcher{
+		isZip,
```
makesitgo, line 35: open to suggestions here, but I check with andy and he thought we tend to read `.zip` files the fastest

internal/commands/push/command.go
```diff
@@ -305,7 +311,13 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 				if err != nil {
 					return err
 				}
-				s.SetMessage(fmt.Sprintf("Installing dependencies: %s...", status.Message))
+
+				message := status.Message
+				if message == "" {
+					message = "finishing up"
```
makesitgo, line 317: this is a drive-by fix, I noticed at the tail-end of uploading dependencies, we ended up with a spinner message saying `Installing dependencies: ...` I guess `status.Message` gets returned as blank towards the end of our polling there

arahmanan (approved):
LGTM! Just some optional comments from my end.

internal/local/archive.go
```diff
@@ -19,143 +16,77 @@ const (
...
-func newTarReader(r io.Reader) ArchiveReader {
-	return &tarReader{Reader: tar.NewReader(r)}
-}
-
-func (r *tarReader) Next() (FileHeader, error) {
-	h, err := r.Reader.Next()
-	if err != nil {
-		return FileHeader{}, err
+	for _, path := range archives {
+		ext, ok := extCache[path]
```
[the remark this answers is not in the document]
makesitgo: the purpose for it was to avoid calling `strings.ToLower(filepath.Ext(path))` twice on each path... probably an unnecessary optimization in the grand scheme of cli-things but here we are /shrug

internal/commands/push/command.go
makesitgo: we already have that success message, I like your idea to just `break` early

internal/local/dependencies.go
```diff
@@ -61,5 +70,90 @@ func FindPackageJSON(path string) (Dependencies, error) {
 		return Dependencies{}, err
 	}
 
-	return Dependencies{rootDir, packageJSONPath}, nil
+	return Dependencies{rootDir, packageJSONPath, false}, nil
+}
+
+// PrepareUpload will prepare the dependencies for upload
+func (d Dependencies) PrepareUpload() (string, func(), error) {
```
nathantfrank, line 77: [q] Would it be beneficial to future devs to leave a short doc string/comment that the second argument is a cleanup function and should be called?

nathantfrank (approved):
LGTM! Just had a non blocking Q

internal/local/dependencies.go
makesitgo: yeah, i like that callout here

PR #220, opened 2021-10-28
REALMC-10671 add endpoints to app structure and description

kpatel71716:


kpatel71716:
I left things in v2 considering the decision to go to keeping endpoints in the `http_endpoints` folder. Open to any feedback here

internal/cloud/realm/app_description.go
```diff
@@ -25,12 +25,60 @@ type IncomingWebhookSummary struct {
...
+	HTTPServiceSummary
+	EndpointSummary
+}
+
+// HTTPEndpoints contains summaries for an http service or endpoint
+type HTTPEndpoints struct {
+	Summaries []interface{}
+}
+
+// MarshalJSON marshals the inner HTTPEndpoints data to JSON
```
makesitgo, line 50:
```suggestion
// MarshalJSON marshals the HTTPEndpoints data to JSON

// or MarshalJSON marshals the inner summaries data to JSON
```

internal/local/structure_v2.go
```diff
@@ -241,6 +253,24 @@ func parseFunctionsV2(rootDir string) (FunctionsStructure, error) {
 	return FunctionsStructure{configs, sources}, nil
 }
 
+func parseEndpointsV2(rootDir string) (EndpointStructure, error) {
```
makesitgo, line 256: [question] where did we land on endpoints being for all config versions?  I still think its an additive change here, but I see in the backend `/import` supports it but `/export` does not

makesitgo (commented):
one last question about config versions, but otherwise this is lookin good to go

internal/cloud/realm/app_description.go
```diff
@@ -31,6 +31,12 @@ type HTTPEndpointSummary struct {
 	IncomingWebhooks []IncomingWebhookSummary `json:"webhooks"`
 }
 
+// HTTPSEndpointSummary is a short summary for a https endpoint model
+type HTTPSEndpointSummary struct {
+	Route      string `json:"route"`
+	HTTPMethod string `json:"http_method"`
```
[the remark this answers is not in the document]
kpatel71716: Forgot to reply to this, but with endpoints and http services in the same config, I am fine with being a bit more verbose here, especially if users end up  using the ANY(*) type

internal/local/structure_v2.go
kpatel71716: I agree, it would be additive. When I last spoke with product it was not a requirement to add support for other config versions. Though I am open to discussing this change in the next check in since the disconnect between import/export support could be confusing

makesitgo: yeah my main thing is, 6 months from now, it's going to be _more_ confusing than not if we don't just have it fully coded out.  it just ends up being another wrinkle to the overall differences b/w config versions that's going to need to be unwound once we tackle the refactor.

please file ticket regardless (and mark the `TODO` here) ... we can address how it to triage it in the check-in

kpatel71716: done! created https://jira.mongodb.org/browse/REALMC-10879, will add detail to the ticket

PR #222, opened 2021-11-02
REALMC-10590: Detect unknown '--local' directories during 'realm-cli push'

jandersongo:


internal/commands/push/inputs.go
```diff
@@ -57,6 +59,16 @@ func (i *inputs) Resolve(profile *user.Profile, ui terminal.UI) error {
 			return errProjectNotFound{}
 		}
 		i.LocalPath = app.RootDir
+	} else {
+		_, err := os.Stat(path.Join(profile.WorkingDirectory, i.LocalPath))
```
makesitgo, line 63: [thought] I think we can tweak this approach a bit (and change a bit of the existing implementation).  we could check this (on `searchPath`) before we attempt to `local.LoadAppConfig`, and return the proper error if the directory doesn't exist.

then we `local.LoadAppConfig` as normal.

then we can move the `if app.RootDir == ""` outside of its enclosing `if i.LocalPath == ""` check to be right after `local.LoadAppConfig`, and return the other error type if `RootDir` is empty.

internal/commands/push/errors.go
```diff
@@ -8,3 +10,19 @@ func (err errProjectNotFound) Error() string {
 }
 
 func (err errProjectNotFound) DisableUsage() struct{} { return struct{}{} }
+
+type errNonExistantDirectory struct {
```
makesitgo, line 14: so I think we can do all of this with just refactoring the existing `errProjectNotFound` type (which may want a bit of a rename)

we could define it as:

```go
type errProjectInvalid struct {
  path string // effectively searchPath from the 'inputs.go` file
  exists bool // indicating the path existed, but just wasn't a Realm app directory
}

func (err errProjectInvalid) Error() string {
  // logic based on `err.exists` for which error message to return
}
```

makesitgo (commented):
just a small thought on re-working the approach here, but you have all the proper checks coded out

internal/commands/push/errors.go
```diff
@@ -1,10 +1,19 @@
...
+
+type errProjectInvalid struct {
+	Path   string // app project path
+	Exists bool   // indicates if the path exists
 }
 
-func (err errProjectNotFound) Error() string {
-	return "must specify --local or run command from inside a Realm app directory"
+func (err errProjectInvalid) Error() string {
+	if err.Exists { // if path exists but is not a realm app directory
```
makesitgo, line 13: you could probably omit the comment and make this the `if !err.Exists` condition, and everything would still read fairly clearly (but carry on as you see fit here)

internal/commands/push/errors.go
```diff
@@ -1,10 +1,19 @@
 package push
 
-type errProjectNotFound struct {
+import (
+	"fmt"
+)
+
+type errProjectInvalid struct {
+	Path   string // app project path
```
makesitgo, line 8: do these fields actually need to be `Exported`?

i'd also say we can remove this particular comment (considering its possible its _just a path_, and not anything app-related).  the comment on `Exists` is fine, or the field could be named `pathExists`

internal/commands/push/errors_test.go
```diff
@@ -9,7 +9,7 @@ import (
 
 func TestPushErrors(t *testing.T) {
 	t.Run("err project not found should disable usage", func(t *testing.T) {
-		var err error = errProjectNotFound{}
+		var err error = errProjectInvalid{}
 
 		_, ok := err.(cli.DisableUsage)
 		assert.True(t, ok, "expected project not found error to disable usage")
```
makesitgo, line 15:
```suggestion
		assert.True(t, ok, "expected project invalid error to disable usage")
```

internal/commands/push/inputs.go
```diff
@@ -47,17 +49,20 @@ func (i *inputs) Resolve(profile *user.Profile, ui terminal.UI) error {
 		searchPath = profile.WorkingDirectory
 	}
 
+	_, err := os.Stat(path.Join(profile.WorkingDirectory, i.LocalPath))
+	if os.IsNotExist(err) {
+		return errProjectInvalid{Path: i.LocalPath, Exists: false}
```
makesitgo, line 54: [nit] can omit the `Exists: false` part here (that's just the zero value anyways)
```suggestion
		return errProjectInvalid{Path: searchPath, Exists: false}
```

internal/commands/push/inputs.go
```diff
@@ -47,17 +49,20 @@ func (i *inputs) Resolve(profile *user.Profile, ui terminal.UI) error {
...
 
-	if i.LocalPath == "" {
-		if app.RootDir == "" {
-			return errProjectNotFound{}
-		}
-		i.LocalPath = app.RootDir
+	if app.RootDir == "" {
+		return errProjectInvalid{Path: i.LocalPath, Exists: true}
 	}
+	i.LocalPath = app.RootDir
```
makesitgo, line 65: we can still leave this line wrapped in `if i.LocalPath == "" { ... }`, just so we preserve the user's actual input (if provided)

internal/commands/push/inputs.go
```diff
@@ -47,17 +49,20 @@ func (i *inputs) Resolve(profile *user.Profile, ui terminal.UI) error {
...
 		return err
 	}
 
-	if i.LocalPath == "" {
-		if app.RootDir == "" {
-			return errProjectNotFound{}
-		}
-		i.LocalPath = app.RootDir
+	if app.RootDir == "" {
+		return errProjectInvalid{Path: i.LocalPath, Exists: true}
```
makesitgo, line 63:
```suggestion
		return errProjectInvalid{Path: searchPath, Exists: true}
```

internal/commands/push/inputs.go
```diff
@@ -47,17 +49,20 @@ func (i *inputs) Resolve(profile *user.Profile, ui terminal.UI) error {
 		searchPath = profile.WorkingDirectory
 	}
 
+	_, err := os.Stat(path.Join(profile.WorkingDirectory, i.LocalPath))
```
makesitgo, line 52:
```suggestion
	_, err := os.Stat(searchPath)
```

internal/commands/push/inputs_test.go
```diff
@@ -19,7 +19,7 @@ func TestPushInputsResolve(t *testing.T) {
 		defer teardown()
 
 		var i inputs
-		assert.Equal(t, errProjectNotFound{}, i.Resolve(profile, nil))
+		assert.Equal(t, errProjectInvalid{Exists: true}, i.Resolve(profile, nil))
```
makesitgo, line 22: this test should update to reflect the path as well

internal/commands/push/inputs_test.go
```diff
@@ -34,6 +34,32 @@ func TestPushInputsResolve(t *testing.T) {
...
+		profile, teardown := mock.NewProfileFromTmpDir(t, "app_init_input_test")
+		defer teardown()
+
+		localPath := "fakePath"
+
+		i := inputs{LocalPath: localPath}
+		assert.Equal(t, errProjectInvalid{Path: localPath, Exists: false}, i.Resolve(profile, nil))
+	})
+
+	t.Run("should return an error when specified local path is not a realm app project", func(t *testing.T) {
```
makesitgo, line 47: this is _effectively_ the same test as above, but i'm fine with keeping it (to prove out the difference between `LocalPath` not being an app vs `WorkingDirectory` not being an app)

internal/commands/push/inputs_test.go
```diff
@@ -34,6 +34,32 @@ func TestPushInputsResolve(t *testing.T) {
...
 	})
 
+	t.Run("should return an error when specified local path does not exist", func(t *testing.T) {
+		profile, teardown := mock.NewProfileFromTmpDir(t, "app_init_input_test")
+		defer teardown()
+
+		localPath := "fakePath"
+
+		i := inputs{LocalPath: localPath}
+		assert.Equal(t, errProjectInvalid{Path: localPath, Exists: false}, i.Resolve(profile, nil))
```
makesitgo, line 44: [nit] can omit the `Exists: false` part here for brevity

internal/commands/push/inputs_test.go
```diff
@@ -34,6 +34,32 @@ func TestPushInputsResolve(t *testing.T) {
...
+
+	t.Run("should return an error when specified local path is not a realm app project", func(t *testing.T) {
+		profile, teardown := mock.NewProfileFromTmpDir(t, "app_init_input_test")
+		defer teardown()
+
+		localPath := "pathWithNoApp"
+
+		assert.Nil(t, ioutil.WriteFile(
+			filepath.Join(profile.WorkingDirectory, localPath),
+			[]byte(`{"app_id": "eggcorn-abcde", "name":"eggcorn"}`),
```
makesitgo, line 55: you probably don't need all of `WriteFile` here, if you just `os.Mkdir` to make sure the directory exists, the test should fail the same

makesitgo (commented):
one small tweak to the logic of which `path` we're displaying to the user in these errors, and then otherwise various cleanup items.  this is good to go after that though, nicely done!

internal/commands/push/inputs.go
jandersongo: this might have to stay as-is since the `os.Stat` fails to find the path if searchPath is just the "local path"

makesitgo: ah yeah I think I see what I was missing before...

so my question to you is: if/when the `LocalPath` does not exist, does `FindApp` return an error?  or an empty `App`?

if it returns an error, maybe we need to look for that specific error in the `err != nil` case of `local.LoadAppConfig`.  otherwise, I think instead of just my suggestion here, you may need to call `filepath.Abs(searchPath)` first, before doing `os.Stat`

jandersongo: `FindApp` does not return an error when `LocalPath` does not exist. It just returns an empty `App`.

I added the filepath.abs check for `path.Join(profile.WorkingDirectory, i.LocalPath)` instead of just `searchPath`
otherwise it would fail to find the path as such: 
```
=== RUN   TestPushInputsResolve/should_return_an_error_when_specified_local_path_is_not_a_realm_app_project
    inputs_test.go:57: 
        failed to assert equals ( actual, expected )
                push.errProjectInvalid{directory 'pathWithNoApp' does not exist}
                push.errProjectInvalid{directory 'pathWithNoApp' is not a supported Realm app project}
```

makesitgo: the issue with relying on `path.Join(profile.WorkingDirectory, i.LocalPath)` is that when `--local` is the fully qualified path to an app directory, this will fail (because it will prepend `WorkingDirectory` to the fully qualified path).

ultimately, what we're looking for is:

* when `--local` is not set, we assume we're looking in the current working directory
* when `--local` is set to a relative directory (basically when it doesn't begin with either `/` or `~`), we assume we're looking in the current working directory PLUS that local path
* when `--local` is set to a fully qualified path, we assume we're just looking there

thankfully, using `filepath.Abs` on what we define as `searchPath` handles most of this complexity (that's why the `local.FindApp` function works the way it does)

I think after you determine `searchPath`, instead of calling `filepath.Abs` on the joined paths like you have, you instead want:

```go
// we'll replace searchPath with its "absolute representation"
searchPath, err = filepath.Abs(searchPath)
if err != nil {
	return err
}

// you can os.Stat(searchPath) now and see if it even exists first

// its this searchPath that we should ultimately pass to our error for messaging
```

internal/commands/push/inputs.go
```diff
@@ -47,15 +49,25 @@ func (i *inputs) Resolve(profile *user.Profile, ui terminal.UI) error {
...
+		return errProjectInvalid{path: searchPath}
+	}
+
 	app, err := local.LoadAppConfig(searchPath)
 	if err != nil {
 		return err
 	}
 
+	if app.RootDir == "" {
+		return errProjectInvalid{path: searchPath, pathExists: true}
```
makesitgo, line 67: [style/go]  no change needed here, just figured i'd take this opportunity to point out some go things:

in case you weren't aware, this error could also be created via
```go
return errProjectInvalid{searchPath, true}
```

basically if/when you are instantiating a struct with all of its fields populated, you can specify the values without the respective `fieldName: ` in front of it.

the only reason for it would be making things potentially read easier.  however, there is a drawback of it to be aware of: there's a chance this syntax can inadvertently introduce a bug.  if a field is added or removed, the go compiler will catch that here (b/c it has the incorrect number of fields).  if the fields are re-arranged, the go compiler would _also_ catch that for this instance b/c the fields are different types.  but if you have all `string` (or same typed) fields, then using this syntax can be dangerous as you'll silently change which fields you are assigning the values towards

anyways, ignore me if you knew all of this already.  just thought i'd take a second to highlight some neat "go things" and the gotchas associated with them

internal/commands/push/inputs_test.go
```diff
@@ -19,7 +19,7 @@ func TestPushInputsResolve(t *testing.T) {
 		defer teardown()
 
 		var i inputs
-		assert.Equal(t, errProjectNotFound{}, i.Resolve(profile, nil))
+		assert.Equal(t, errProjectInvalid{path: profile.WorkingDirectory, pathExists: true}, i.Resolve(profile, nil))
```
makesitgo, line 22: the shorthand (read: no need for `fieldName: ` declarations) syntax works well in tests, b/c presumably even if we inadvertently switch the ordering of the fields, the test will still fail

PR #228, opened 2021-11-30
REALMC-11032: Enable package.json to be uploaded from app root dir

Amaan95:
- I tested these changes locally with a `package.json` in the root dir as well as in the functions dir. The functionality works as expected.
- [drive-by] add additional info on which dependencies path failed to upload

### DIFF
--- a/internal/local/dependencies.go
+++ b/internal/local/dependencies.go
@@ -59,18 +59,27 @@ func FindPackageJSON(path string) (Dependencies, error) {
 		return Dependencies{}, nil
 	}
 
-	rootDir := filepath.Join(app.RootDir, NameFunctions)
+	functionsDir := filepath.Join(app.RootDir, NameFunctions)
+	packageJSONFunctionsPath := filepath.Join(functionsDir, NamePackageJSON)
+	packageJSONRootPath := filepath.Join(app.RootDir, NamePackageJSON)
 
-	packageJSONPath := filepath.Join(rootDir, NamePackageJSON)
+	// Take the package.json in the functions directory as the source of truth
+	if _, err := os.Stat(packageJSONFunctionsPath); err != nil {
+		if !os.IsNotExist(err) {
+			return Dependencies{}, err
+		}
+	} else {
+		return Dependencies{functionsDir, packageJSONFunctionsPath, false}, nil
+	}
 
-	if _, err := os.Stat(packageJSONPath); err != nil {
+	// If the package.json is not found in the functions directory, look in the root dir
+	if _, err := os.Stat(packageJSONRootPath); err != nil {
 		if os.IsNotExist(err) {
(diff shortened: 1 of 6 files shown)


internal/local/dependencies.go
```diff
@@ -59,18 +59,27 @@ func FindPackageJSON(path string) (Dependencies, error) {
 		return Dependencies{}, nil
 	}
 
-	rootDir := filepath.Join(app.RootDir, NameFunctions)
-
-	packageJSONPath := filepath.Join(rootDir, NamePackageJSON)
+	functionsDir := filepath.Join(app.RootDir, NameFunctions)
+	packageJSONFunctionsPath := filepath.Join(functionsDir, NamePackageJSON)
+	packageJSONRootPath := filepath.Join(app.RootDir, NamePackageJSON)
```
arahmanan, line 64: [opt] thoughts on moving this variable down right before we use it?

internal/local/dependencies.go
```diff
@@ -59,18 +59,27 @@ func FindPackageJSON(path string) (Dependencies, error) {
...
+
+	// Take the package.json in the functions directory as the source of truth
+	if _, err := os.Stat(packageJSONFunctionsPath); err != nil {
```
arahmanan, line 67: [opt] to reduce nesting / increase readability you could do:
```
err := os.Stat(packageJSONFunctionsPath)
if err == nil {
  return Dependencies{functionsDir, packageJSONFunctionsPath, false}, nil
}
if !os.IsNotExist(err) {
  return Dependencies{}, err
}
...
```

internal/local/dependencies.go
```diff
@@ -59,18 +59,27 @@ func FindPackageJSON(path string) (Dependencies, error) {
...
+		if !os.IsNotExist(err) {
+			return Dependencies{}, err
+		}
+	} else {
+		return Dependencies{functionsDir, packageJSONFunctionsPath, false}, nil
+	}
 
-	if _, err := os.Stat(packageJSONPath); err != nil {
+	// If the package.json is not found in the functions directory, look in the root dir
+	if _, err := os.Stat(packageJSONRootPath); err != nil {
```
arahmanan, line 76: [opt] not really part of your change, but you could simplify this as well with:
```
err = os.Stat(packageJSONRootPath);
if os.IsNotExist(err) {
  return Dependencies{}, fmt.Errorf("package.json not found at %s or %s", functionsDir, app.RootDir)
}
if err != nil {
  return Dependencies{}, err
}
return Dependencies{app.RootDir, packageJSONRootPath, false}, nil
```

arahmanan (approved):
LGTM! just optional stuff from my end.

Amaan95:
Per the discussion in [this thread](https://mongodb.slack.com/archives/C01T87L1QE5/p1638379698012600), these changes are no longer needed so I'll be closing this PR out.

PR #232, opened 2021-12-10
REALMC-10548: Improve CLI documentation in README and for releases

makesitgo:
_pretty sure_ this works... I have a followup bug fix to be able to actually test this out with


.evg.yml
```diff
@@ -320,6 +322,104 @@ tasks:
...
+            touch release.true
+      - command: shell.exec
+        params:
+          shell: "bash"
+          script: |
+            set -e
+            if [ ! -f realm-cli/release.true ]; then
+              # not a release build
+              exit 0
+            fi
```
tsedgwick, line 366: we should be able to simplify this and remove it completely. 
since we don't need to add the npm creds _yet_, we can remove L357-366

.evg.yml
```diff
@@ -320,6 +322,104 @@ tasks:
...
+            echo "updating manifest"
+            ./realm-cli/contrib/manifest.js ${build_id} ${cli_version} manifest.json
+      - command: s3.put
+        params:
+          aws_key: ${artifacts_aws_key}
+          aws_secret: ${artifacts_aws_secret}
+          local_file: manifest.json
+          remote_file: versions/cloud-prod/CURRENT
```
tsedgwick, line 374: so I'm understanding, we _always_ push to s3 on every master build, but then we push again on release?

.evg.yml
```diff
@@ -320,6 +322,104 @@ tasks:
...
+      - command: shell.exec
+        params:
+          shell: "bash"
+          script: |
+            set -e
+            cd realm-cli
+            if [ ! -f release.true ]; then
+              # not a release build
+              exit 0
+            fi
```
tsedgwick, line 387: we probably can add this above instead of splitting across commands (and thus remove the release.true)

.evg.yml
```diff
@@ -339,10 +439,11 @@ buildvariants:
     realm_server_url: "http://localhost:9090"
     yarn_url: "https://s3.amazonaws.com/stitch-artifacts/yarn/latest.tar.gz"
     node_version: 12.16.2
-    version: 2.3.0
+    cli_version: 2.3.0
```
tsedgwick, line 442: what was this renamed for?

tsedgwick (commented):
mostly looks good! couple comments from me.

is `testInstall.js` needed?

makesitgo:
`testInstall.js` is a manual sanity check at this point.  I think there were historical plans (@chickenbug might recall) to potentially include this as an evergreen task at some point but we never went down that route.  it _was_ helpful for me when I was updating `install.js` itself

.evg.yml
makesitgo: "grep-ability" ... basically this bit:

```cmd
cat .evg.yml | grep cli_version: | awk '{ print $2 }' | tail -n 1
```

the `tail -n 1` worked for now in grabbing the `version: ` field I was looking for, but this felt a bit more future-proof

.evg.yml
makesitgo: yeah, so our s3 interactions currently:

1) on _every_ merge to `master`, we will push CLI builds to S3 (keyed by evergreen's build id ... which is effectively the commit hash)

2) on _every_ merge to `master`, we will fetch AND push the CLI's version manifest (the `CURRENT` doc)

3) on merge to `master` when the `.evg.yml` file has a `cli_version` field which does not yet have a corresponding tag on the remote repo, we will also update the CLI's version manifest to include the new version info (before pushing it back to `s3`)

.evg.yml
makesitgo: I think I'm fine with moving the "tag stuff" above, so that this runs

```yaml
echo "tagging release: v${cli_version}"
PREVIOUS_VERSION=$(git describe --tags $(git rev-list --tags --max-count=1))
BODY=$(git log --no-merges $PREVIOUS_VERSION..HEAD --pretty="format:%s (%h); %an" | sed '$d')
git tag -a -m "Changelog since $PREVIOUS_VERSION:" -m "$BODY" "v${cli_version}"
git push origin master --tags
```

before pushing back to s3 with the potentially updated manifest.

however, the `release.true` could still be helpful by keeping the `npm` stuff separate/after the push to s3.  i'm assuming the `npm publish` part is prone to failure, so trying to keep that at the tail end of this task so that should it fail, we're still in a good place for someone to just locally run `npm publish` without having to do anything else (i.e. `s3` is in a good place and github has its expected tags)

tsedgwick (approved):
can we update the title of this PR since it is automating our release primarily, and secondarily updating the documentation?

PR #235, opened 2021-12-14
REALMC-11049: Use remote app id for push command success message

makesitgo:


internal/commands/push/command.go
```diff
@@ -372,7 +373,7 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		}
 	}
 
-	ui.Print(terminal.NewTextLog("Successfully pushed app up: %s", app.ID()))
+	ui.Print(terminal.NewTextLog("Successfully pushed app up: %s", appRemote.ClientAppID))
```
makesitgo, line 376: ⚠️  ... this used to print the `app_id` found in the app root's `realm_config.json` (or similar file). however, as we have started to see users use a single local app directory to then push to various remote Realm apps, printing the appRemote id makes more sense here.

just going to say it out loud as food for thought in our import/export refactor, but we probably shouldn't be dealing with app_id in any app configuration file (seems more and more likely that we end up remembering/managing multiple sets of remote app+group ids per "local app directory")

jandersongo (approved):
LGTM

tsedgwick (approved):

PR #236, opened 2021-12-15
REALMC-11198: Bump version to 2.3.1

makesitgo:


.evg.yml
```diff
@@ -419,7 +419,7 @@ buildvariants:
     realm_server_url: "http://localhost:9090"
     yarn_url: "https://s3.amazonaws.com/stitch-artifacts/yarn/latest.tar.gz"
     node_version: 12.16.2
-    cli_version: 2.3.0
+    cli_version: 2.3.1
```
tsedgwick, line 422: couldn't we pull this from the package.json instead of having to update in two spots?

tsedgwick (approved):
let's get that money 🤞🏻

.evg.yml
makesitgo: we probably could... a quick refactor of my `cat`/`grep` command pretty quickly grabs `"2.3.1",` (and trimming the quotes and comma shouldn't be too hard).

it _is_ slightly irksome we're left with modifying a version AND running a version bump script here.  however, it also feels odd to make `package.json` as the CLI version source of truth... it just seems like there's more to the CLI than just it being an `npm` package (to mike's point last week, we could start exploring this as a `homebrew` thing or encourage folks to just `cURL` it)

tsedgwick: we use `jq` in other .evg.yml release tasks, maybe we should utilize it here as well to stay consistent?

I agree package.json may not be the best long term, but it would be a good short-mid term until we figure out how we want to handle the install going forward

makesitgo: sounds good to me... I filed [REALMC-11205](https://jira.mongodb.org/browse/REALMC-11205) to automate the commit message in the version bump script, so I can tackle that there as well

PR #242, opened 2021-12-30
REALMC-9026: display the default value for input prompt when selected

jandersongo:
Updated the survey pkg: https://github.com/alphayan/survey to version [v2.2.3](https://github.com/AlecAivazis/survey/releases/tag/v2.2.3)
Now the default value is shown when selected.


makesitgo (approved):
nice find on this :)

🆒  beans

PR #244, opened 2022-01-12
REALMC-11391: Support local-userpass auth provider type for login

makesitgo:
Mentioned this in the JIRA, but this functionality was technically missed when going from `v1` to `v2`, see: https://github.com/10gen/realm-cli/blob/support/1.3.x/auth/auth.go

Not sure how far we should go to support these use cases of "self-contained Realm instances" when testing in CI/CD environments, but Nikola reached out asking me about this so it's safe to say we have at least one real use-case for all of this

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,3 +138,90 @@ c, err := expect.NewConsole(expect.WithStdout(out))
 assert.Nil(t, err)
 defer c.Close()
 ```
+
+## Undocumented Features and Flags
+
+### Global Flags
+
+#### Using `--realm-url` and `--atlas-url`
+
+At some point, you may wish to configure which Realm and Atlas servers the CLI is connecting to.  By default, these URLs point to the Production instances of Realm and Atlas, which should be sufficient for most use cases.
+
+However, you can set `--realm-url` and `--atlas-url` to any URLs you know a respective server instance is running at.  If you are configuring one, chances are you'll want to configure both of these.
+
+### Pseudo-Global Flags
+
+#### Using `--project`
+
+At some point, you may wish to target a specific "project id" (a.k.a. "group id") when performing any sort of app lookup.  By default, the project id is derived from the list of projects available to the logged in user, which should be sufficient for most use cases (but may force a prompt for a project selection, thus breaking typical CI/CD setups).  The impact of this setting effects only the prompts presented by the CLI (and the data found within them) to allow the user to select a project (and ultimately an app) to work with.
+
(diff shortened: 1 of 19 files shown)


README.md
```diff
@@ -138,3 +138,90 @@ c, err := expect.NewConsole(expect.WithStdout(out))
...
+
+## Undocumented Features and Flags
+
+### Global Flags
+
+#### Using `--realm-url` and `--atlas-url`
+
+At some point, you may wish to configure which Realm and Atlas servers the CLI is connecting to.  By default, these URLs point to the Production instances of Realm and Atlas, which should be sufficient for most use cases.
+
+However, you can set `--realm-url` and `--atlas-url` to any URLs you know a respective server instance is running at.  If you are configuring one, chances are you'll want to configure both of these.
```
tsedgwick, line 150: wouldn't you _always_ need to configure both? not sure how realm dev would work with cloud qa for example

tsedgwick (approved):

README.md
makesitgo: the one a-typical scenario thats come up in practice is local+dev, but the technicality still remains that you'd be setting both of those at once.

also, if you didn't heed this advice, then you'd read it and say "oh yeah, those chances were right!  but now I only need to set one" :)

makesitgo: I wonder if you could ever rig up a local Realm + prod Atlas env...

jandersongo (approved):

makesitgo (commented):
got burned by "pending review" instead of just submitting comments... submitting them now

internal/commands/login/inputs_test.go
```diff
@@ -89,6 +94,83 @@ func TestLoginInputs(t *testing.T) {
...
+		{
+			description: "Should not prompt for inputs when flags provide the data for local type",
+			inputs: inputs{
+				AuthType: authTypeLocal,
+				Username: "username",
+				Password: "password",
+			},
+			prepareProfile: func(p *user.Profile) {},
+			procedure: func(c *expect.Console) {
+				// c.ExpectEOF()
```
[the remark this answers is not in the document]
makesitgo: probably not needed (so yes, should be commented out or really just removed) as the tests are passing... I admittedly just blindly copy/pated the initial test cases here and dupe'd them for the next auth provider type.

going to remove this here and in the likely original source spot as well

makesitgo:
test failure is currently related to https://github.com/10gen/baas/pull/5563/files.  proceeding with the merge

PR #251, opened 2022-02-07
REALMC-11700: Refactor errors to allow suggestions and reference links to display

makesitgo:
this was (mostly) previously broken (only `DisableUsage` was working as intended here, but for the wrong reasons)

for an "invalid session" error returned by Realm, the output was previously:

```cmd
$ realm-cli --profile local apps list
apps list failed: invalid session
$ ...
```

with these changes, we now get:

```cmd
$ realm-cli --profile local apps list
apps list failed: invalid session
Try instead: realm-cli login --profile local
$ ...
```


internal/utils/cli/error.go
```diff
@@ -0,0 +1,138 @@
...
+		suggestions:    d.Suggestions,
+		referenceLinks: d.ReferenceLinks,
+	}
+}
+
+// WrapErr wraps the provided error with the provided message
+// The extra details are applied on top of any existing, wrapped details
+// Note: it is assumed the message contains a single `fmt` verb to include the cause/error
+func WrapErr(msg string, cause error, details ...ErrDetail) error {
+	err, ok := cause.(cliErr)
```
andrewzlchen, line 46: q: Is there a strong reason why we need `cliError` and `ErrDetails`? Is `ErrDetails` a struct that should only be used as the arguments to the `Apply` interface method?

In the end, I'm fine with keeping both, but wanted to see if we can DRY things up a bit here.

internal/utils/cli/error.go
```diff
@@ -0,0 +1,138 @@
...
+		cause:          cause,
+		hideUsage:      hideUsage,
+		suggestions:    d.Suggestions,
+		referenceLinks: d.ReferenceLinks,
+	}
+}
+
+// WrapErr wraps the provided error with the provided message
+// The extra details are applied on top of any existing, wrapped details
+// Note: it is assumed the message contains a single `fmt` verb to include the cause/error
```
andrewzlchen, line 44: > // Note: it is assumed the message contains a single `fmt` verb to include the cause/error

q: can you clarify this a bit more? I'm not really sure what this means

internal/cli/command_factory.go
```diff
@@ -160,11 +161,15 @@ func (factory *CommandFactory) Run(cmd *cobra.Command) int {
 		log.Print(err)
 	} else {
 		logs := []terminal.Log{terminal.NewErrorLog(err)}
-		if e, ok := err.(Suggester); ok {
-			logs = append(logs, terminal.NewFollowupLog(terminal.MsgSuggestions, e.Suggestions()))
+
+		var suggester cli.ErrSuggester
+		if errors.As(err, &suggester) {
+			logs = append(logs, terminal.NewFollowupLog(terminal.MsgSuggestions, suggester.Suggestions()...))
```
andrewzlchen, line 167: note: I attached a screenshot to my review comment regarding this, but I think we need to add a guard clause here to see if there any suggestions/links before printing the log message out.

ie:

```go
		if errors.As(err, &suggester) && len(suggester.Suggestions) > 0 {
			logs = append(logs, terminal.NewFollowupLog(terminal.MsgSuggestions, suggester.Suggestions()...))
...

```

andrewzlchen (changes requested):
🔥 Looks great so far! I really like the errors framework that you built out here. 

I think it's super easy for consumers of that `cli` error to define custom links, provide suggestions, and hide usage details, if necessary.

That said, I had a few comments regarding the implementation of that framework. Most of the comments are regarding making the code "easier to read" so they're mostly optional/subjective in nature.

I ran this locally and got this though

image

I think we may need to add guard clauses to the terminal printer to check if any reference links exist before printing the "For more information" string.

The PR looks good otherwise!

internal/utils/cli/error.go
andrewzlchen: EDIT: Ah i see now. Maybe we can rename the `msg` argument to `msgFormat`?

This is what `fmt.Sprintf` looks like:

image

internal/cloud/realm/error.go
```diff
@@ -22,19 +23,13 @@ var (
...
 // ErrInvalidSession is an invalid session error
-type ErrInvalidSession struct {
-	profileName string
-}
-
-func (err ErrInvalidSession) Error() string { return "invalid session" }
-
-// Suggestions returns a list of commands to run to remedy an invalid session error
-func (err ErrInvalidSession) Suggestions() []interface{} {
+func ErrInvalidSession(profileName string) error {
```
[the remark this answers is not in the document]
makesitgo: aside from the question of naming things "go-like" or not, i'm going with this name to reduce the "diff friction" for this PR (keep it as existing).

for my own style/preference though, I like "things" to be named similar, so whether the error is a `var`, `func`, or `type` I'd like to be referencing it via `err...`.  the compiler pretty quickly lets you know if youre using a struct as a func or any other wackiness

internal/commands/app/errors.go
```diff
@@ -1,15 +1,15 @@
...
-	path string
-}
+import (
+	"errors"
+
+	"github.com/10gen/realm-cli/internal/utils/cli"
+)
 
-func (err errProjectExists) Error() string {
+func errProjectExists(path string) error {
```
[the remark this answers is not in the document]
makesitgo: going to leave as-is primarily for ease-of-diff purposes.  I don't think the notion that `err*`/`Err*` prefixes should be reserved for types, as that both limits our ability to reliably prefix downstream from that (e.g. `errCode*`) and contradicts the error values we create with the same prefix (i.e. `errAudClaim`)

internal/commands/app/errors_test.go
```diff
@@ -3,15 +3,16 @@ package app
...
-	t.Run("err project exists should disable usage", func(t *testing.T) {
-		var err error = errProjectExists{}
+	t.Run("err project exists should hide usage", func(t *testing.T) {
+		var err error = errProjectExists("")
 
-		_, ok := err.(cli.DisableUsage)
-		assert.True(t, ok, "expected project exists error to disable usage")
+		usageHider, ok := err.(cli.ErrUsageHider)
+		assert.True(t, ok, "expected project exists error to hide usage")
+		assert.True(t, usageHider.HideUsage(), "should hide usage")
```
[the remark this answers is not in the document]
makesitgo: that's tested separately from here in `app/create_inputs_test`/`app/init_inputs_test`

internal/utils/cli/error.go
```diff
@@ -0,0 +1,138 @@
...
+	"fmt"
+)
+
+// ErrUsageHider may hide the command's usage when an error occurs
+type ErrUsageHider interface {
+	HideUsage() bool
+}
+
+// ErrSuggester provides a list of suggestions that will display to the user when an error occurs
+type ErrSuggester interface {
```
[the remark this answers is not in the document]
makesitgo: you are right that these things don't necessarily _need_ to be specific to errors, but that's currently the only way they can and are used for the CLI currently.

a big part of the command pattern in this CLI is that the command should be self-contained.  if there's output that is relevant to the success path of the command that ran, then the `Handler` func should own that.  it prevents us from having to traverse many code paths in determining how/why a command outputs the things it does.

to further enforce this pattern, the `Handler` interface itself was designed to only return `error` (and no other value)

when it came to the `cobra` framework and how they handled errors, it became clear there were many paths an error could surface and one place to make sure we handled _all_ of them: the return value of `cmd.Execute`.  because of that, we were forced to somehow try and be clever with the error handling.  as you can see from the previous versions of this code, that fell short.

internal/utils/cli/error.go
```diff
@@ -0,0 +1,138 @@
...
+// Note: it is assumed the message contains a single `fmt` verb to include the cause/error
+func WrapErr(msg string, cause error, details ...ErrDetail) error {
+	err, ok := cause.(cliErr)
+	if !ok {
+		return NewErr(fmt.Errorf(msg, cause), details...)
+	}
+
+	var d ErrDetails
+	for _, detail := range details {
+		detail.Apply(&d)
```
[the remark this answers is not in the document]
makesitgo: yeah, I was playing around with a new pattern here (trying to ease the load around dealing with an `*bool` type)

I fail to see too much complexity here though, its a really simple for-loop and setter at the end of the day (it's actually fairly similar to how the mongo driver presents `Options`)

from the consuming code perspective, `ErrDetail` and `ErrDetails` are more-or-less invisible, and folks are only concerned with using `cli.ErrNoUsage`/`cli.ErrSuggestion`/`cli.ErrReferenceLink`

if i could make `ErrDetail`/`ErrDetails`/`Apply` all unexported, I would.  and to be fair, I don't think the compiler is what's stopping me there, but the resulting function signatures would be a bit awkward with that

internal/utils/cli/error.go
makesitgo: `cliError` is the internal representation these helpers are relying on as an "error type" (read: implement `Error() string`)

`ErrDetails` is the "input struct" presenting calling code with the various options present to inject into the error.  it's fields are more options (`*bool` vs `bool`) and doesn't implement the `error` interface

internal/utils/cli/error.go
makesitgo: to be fair, I don't think many people would like the name "format" for what this is either... pattern or template maybe, but those are also already first-class things in the language.  i'll change the arg name here, but my hope is people look to the code for what this is doing over the arg name

internal/cli/command_factory.go
makesitgo: nice catch here

internal/utils/cli/error.go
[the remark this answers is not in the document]
jandersongo: it is an interesting approach! but I have to agree with Andrew, it was a bit confusing to follow
[opt] just a suggestion for replacing the mutation ( `Apply` method)
```

// NewErr returns a new CLI error
func NewErr(cause error, details ...ErrDetail) error {
	var d ErrDetails
	var hideUsage bool

	for _, detail := range details {
		switch detail.(type) {
		case ErrSuggestion:
			d.Suggestions = append(d.Suggestions, detail.Value())
		case ErrReferenceLink:
			d.ReferenceLinks = append(d.ReferenceLinks, detail.Value())
		case ErrNoUsage:
			hideUsage = true
		}
	}

	return cliErr{
		cause:          cause,
		hideUsage:      hideUsage,
		suggestions:    d.Suggestions,
		referenceLinks: d.ReferenceLinks,
	}
}

```

```
// ErrDetail represents a single detail of a CLI error
type ErrDetail interface {
	Value() interface{}
}
```
with ex: 
```
func (err ErrSuggestion) Value() interface{} {
	return err.Suggestion
}
```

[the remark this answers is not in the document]
makesitgo: worked off-thread on this together, and we're going to stick with the current design + a few name tweaks

internal/utils/cli/error_test.go
```diff
@@ -0,0 +1,119 @@
...
+				hideUsage:      true,
+				suggestions:    []interface{}{"suggestion"},
+				referenceLinks: []interface{}{"ref.link"},
+			},
+			err,
+		)
+
+		t.Run("and be recognized as a usage hider", func(t *testing.T) {
+			var usageHider ErrUsageHider
+			assert.True(t, errors.As(err, &usageHider), fmt.Sprintf("%T is not a usage hider", err))
```
jandersongo, line 54: should this assertion message be `"%T is a usage hider"` instead of "not"?

internal/utils/cli/error_test.go
```diff
@@ -0,0 +1,119 @@
...
+			assert.True(t, usageHider.HideUsage(), "should hide usage")
+		})
+
+		t.Run("and be recognized as a suggester", func(t *testing.T) {
+			var suggester ErrSuggester
+			assert.True(t, errors.As(err, &suggester), fmt.Sprintf("%T is not a suggester", err))
```
jandersongo, line 60: similarly should this assertion message be `"%T is a suggester"` instead of "not"?

internal/utils/cli/error_test.go
```diff
@@ -0,0 +1,119 @@
...
+			assert.Equal(t, suggester.Suggestions(), []interface{}{"suggestion"})
+		})
+
+		t.Run("and be recognized as a link referrer", func(t *testing.T) {
+			var linkReferrer ErrLinkReferrer
+			assert.True(t, errors.As(err, &linkReferrer), fmt.Sprintf("%T is not a link referrer", err))
```
jandersongo, line 66: same here ... `"%T is a link referrer"` instead of "not"?

internal/utils/cli/error_test.go
makesitgo: the message passed into `assert.True` is what prints back to the user...so this is displaying the reason the assert failed.

to your point it probably should read more like `"expected error to be ErrUsageHider, but got %T instead"`.  I had the same double-take on these error messages (much of this was existing code) thinking they were backwards

jandersongo (approved):

andrewzlchen (approved):
🔥 

```
    /-------\
    | lgtm! |
    \-------/
          \
           \
            \

                                _._
                              _/:|:
                             /||||||.
                             ||||||||.
                            /|||||||||:
                           /|||||||||||
                          .|||||||||||||
                          | ||||||||||||:
                        _/| |||||||||||||:_=---.._
                        | | |||||:'''':||  '~-._  '-.
                      _/| | ||'         '-._   _:    ;
                      | | | '               '~~     _;
                      | '                _.=._    _-~
                   _.~                  {     '-_'
           _.--=.-~       _.._          {_       }
       _.-~   O-,        {    '-._     _. '~==+  |
      ('          }       \_      \_.=~       |  |
      `,,,,,,,'  /_         ~-_    )         <_oo_>
        `-----~~/ /'===...===' +   /
               <_oo_>         /  //
                             /  //
                            <_oo_>
```

internal/utils/cli/error.go
```diff
@@ -0,0 +1,138 @@
+package cli
```
tsedgwick, line 1: idiomatic go strongly discourages util packages - and this is specific to errors not the cli.

what're your thoughts of making this its own package outside of internal maybe? I'd prefer not errors, but something along those lines.

tsedgwick (commented):
Is the thought we have to remember to wrap _all_ or _most_ of our errors with this new error type?

internal/utils/cli/error.go
makesitgo: haha `errors` was a close second throughout all this...and it'd kill my ocd to have this bugfix be the sole thing that creeped outside of `internal/` 😬 

as a compromise, two other thoughts:

existing `terminal` package: this is basically the lowest-level package of imports, and can be used freely through the codebase.  my hesitation here is that it will ultimately mean the `cloud` packages (`atlas` and `realm`) have an import to the `terminal`/"UI"package which feels like an unnecessary coupling

new `cli/feedback` package: as a compromise (similar to what we've already done with the `cli/user` package so that it can be imported anywhere

refactor `cli/command_factory` to some nested `cli/runner` package, and keep all of these low-level concepts in the root `cli` package.  I like this most, and actually started down this path at one point during the reviews, but it is probably the biggest life.

any thoughts from your end hearing all of this?

makesitgo: haha `errors` was a close second throughout all this...and it'd kill my ocd to have this bugfix be the sole thing that creeped outside of `internal/` 😬 

as a compromise, two other thoughts:

existing `terminal` package: this is basically the lowest-level package of imports, and can be used freely through the codebase.  my hesitation here is that it will ultimately mean the `cloud` packages (`atlas` and `realm`) have an import to the `terminal`/"UI"package which feels like an unnecessary coupling

new `cli/feedback` package: as a compromise (similar to what we've already done with the `cli/user` package so that it can be imported anywhere

refactor `cli/command_factory` to some nested `cli/runner` package, and keep all of these low-level concepts in the root `cli` package.  I like this most, and actually started down this path at one point during the reviews, but it is probably the biggest life.

any thoughts from your end hearing all of this?

makesitgo:
@tsedgwick the thought here (and main pattern throughout this codebase) is that developers should mostly not think about errors.  no wrapping, no `fmt.Errorf("...", err)` needed, just simply `return err` in the `Handler` and inputs resolution.

this is an opt-in pattern where you _could_ use it if you want to return any extra information, but should mostly not have to think about it

internal/utils/cli/error.go
tsedgwick: Since this is opt in, I think cli/feedback makes the most sense to me

tsedgwick (approved):

PR #252, opened 2022-02-08
REALMC-10890: fix search for app root config file

jandersongo:
Failed test is resolved in https://github.com/10gen/realm-cli/pull/256

This fix assumes that a root config file contains the config version. I figured this would be the easiest way to distinguish between a root config file and other configs like the one sometimes found in /graphql. 
[for this reason, I had to add the `config_version` param to some other already existing tests which would otherwise fail]

side note ...
Found another bug: https://jira.mongodb.org/browse/REALMC-11800


internal/local/app_test.go
```diff
@@ -111,47 +130,33 @@ func TestFindApp(t *testing.T) {
...
 					assert.True(t, insideProject, "should be inside project")
 
-					app, err := LoadApp(projectRoot)
+					app, err := LoadApp(path)
 					assert.Nil(t, err)
 					assert.Equal(t, tcInner.appData, app.AppData)
 				})
 			}
 
-			t.Run("And a working directory at the project root return the correct data", func(t *testing.T) {
```
jandersongo, line 133: this was a duplicate of the test below, so I removed it

internal/local/app_test.go
```diff
@@ -673,87 +678,39 @@ exports = function({ query }) {
...
-	CustomUserDataConfig: map[string]interface{}{
-		"enabled":            true,
-		"mongo_service_name": "mongodb-atlas",
-		"database_name":      "test",
-		"collection_name":    "coll3",
-		"user_id_field":      "xref",
-	},
-	Sync: map[string]interface{}{"development_mode_enabled": false},
-}}
+func createAppDataV1(version realm.AppConfigVersion, name string) AppDataV1 {
```
jandersongo, line 681: there would have been 6 very similar vars for each AppDataV1 used for testing 
[appData20200603Remote, appData20200603Local, appData20200603Nested
appData20180301Remote, appData20180301Local, appData20180301Nested]

so I replaced them with this `createAppDataV1` that takes in the various versions

internal/local/app.go
```diff
@@ -236,6 +238,33 @@ func (a *App) LoadConfig() error {
 	return nil
 }
 
+// ConfigContainsVersion returns true if the app root config file contains a version
+func (a *App) ConfigContainsVersion() bool {
```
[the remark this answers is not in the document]
makesitgo: what if we tried to limit ourselves to only reading that file once?

initially, the intended difference b/w `FindApp` and `LoadApp` was supposed to be that "find" would locate the app but not have to read any files/data.  obviously that distinction was a bit short-sighted, as this current bug indicated.  "FindApp" is still going to have to read data in order to truly find itself

in that sense, there's probably not much distinction needed b/w `LoadAppConfig` and `FindApp`

so instead of `FindApp` using the scanner to check for a matching config version, we should just skip right to reading the data and parsing json, to see that we have a non-zero value of `config_version`

internal/local/app.go
```diff
@@ -236,6 +238,33 @@ func (a *App) LoadConfig() error {
...
+	path := filepath.Join(a.RootDir, a.Config.String())
+	file, openErr := os.Open(path)
+	if openErr != nil {
+		return false
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		if strings.Contains(scanner.Text(), a.AppData.ConfigVersion().String()) {
```
makesitgo, line 261: I'm actually surprised this check is doing what we're expecting (do we have tests that prove this works as well as catches failures?)

basically, when you call `a.AppData.ConfigVersion()`, it's going to pull the data we've loaded into `AppData`, which doesn't seem to be anything _but_ the empty pointers youve assigned above in the switch/case.

what does `a.AppData.ConfigVersion().String()` look in each of these iterations?  we might just be checking something like: does the line contain a `"0"`

makesitgo (commented):
I have a few thoughts/questions on the approach here before I get down to the full review

internal/local/app.go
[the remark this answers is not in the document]
jandersongo: this is an interesting point! I did a bit of digging & I think its safe to moved the call to `LoadApp()` from `LoadAppConfig` to within `FindApp`

internal/local/app.go
jandersongo: Added some test cases that catch 'invalid' configs.

internal/local/app_test.go
```diff
@@ -111,46 +134,48 @@ func TestFindApp(t *testing.T) {
...
+				path := filepath.Join(testRoot, config.invalidAppData.Name())
+				_, foundApp, err := FindApp(path)
 				assert.Nil(t, err)
-				assert.True(t, insideProject, "should be inside project")
+				assert.False(t, foundApp, "should not be found")
 
-				app, err := LoadApp(projectRoot)
-				assert.Equal(t, errFailedToParseAppConfig(filepath.Join(projectRoot, config.file.String())), err)
+				app, err := LoadApp(path)
+				assert.Nil(t, err)
```
makesitgo, line 178: [question] what are you thoughts on this being an error scenario instead?  it seems fairly similar to the "empty" case, I'm trying to wrap my head around why they behave differently here

internal/local/app_test.go
```diff
@@ -771,3 +748,19 @@ var appData20210101Remote = AppDataV2{AppStructureV2{
...
+	DeploymentModel:       realm.DeploymentModelGlobal,
+	AllowedRequestOrigins: []string{"http://localhost:8080"},
+	GraphQL: GraphQLStructure{
+		Config: map[string]interface{}{
+			"use_natural_pluralization": true,
+		},
+	},
+}}
+
+var appData20210101Invalid = AppDataV2{AppStructureV2{}}
```
makesitgo, line 766: [ocd/nit] this doesn't seem to have a similar structure as the other "invalid" cases, is that intentional or just convenience?  might be worth either building this out to match, or adding a `if name == "invalid"` short-circuit near the top of the `createAppDataV1` helper

internal/local/app_test.go
```diff
@@ -673,87 +698,39 @@ exports = function({ query }) {
...
+			"mongo_service_name": "mongodb-atlas",
+			"database_name":      "test",
+			"collection_name":    "coll3",
+			"user_id_field":      "xref",
+		},
+		Sync: map[string]interface{}{"development_mode_enabled": false},
+	}}
+
+	switch name {
+	case "remote":
```
makesitgo, line 723: I don't _love_ having logic contained in this helper that requires knowledge of the various `name` inputs it could be passed.  (i also won't die on any hills here as it is just a test helper... 😬 )

one thought is to push this type of logic downstream of whatever's calling `createAppDataV1`... that has the unfortunate side effect for us of forcing a change in our test case fields.  basically the `AppData` fields need to become functional (e.g. `func() AppData` (not ideal!))

so let's consider the opposite approach: embracing the 6 different (but very similar) vars (with a few extra common `var`s, it becomes somewhat less crazy):

```go
var allowedRequestOrigins = []interface{}{...}

var appSecurity = map[string]interface{}{"allowed_request_origins": allowedRequestOrigins}

// other vars for the various re-used components

var appData20180301Local = AppDataV1{AppStructureV1{
  ConfigVersion:   realm.AppConfigVersion20180301,
  Name: "20180301-local",
  Location:        realm.LocationVirginia,
  DeploymentModel: realm.DeploymentModelGlobal,
  Security: appSecurity,
  Hosting: appHosting,
  CustomUserDataConfig: appCustomUserDataConfig,
  Sync: appSync,
}}
```

internal/local/app_test.go
```diff
@@ -673,87 +698,39 @@ exports = function({ query }) {
...
+		app.AppStructureV1.ID = version.String() + "-" + name + "-" + "abcde"
```
makesitgo, line 724: [meganit] this is probably a moot point based on my previous suggestion, this could just be:

`app.AppStructureV1.ID = app.AppStructureV1.Name + "-" + "abcde"`

makesitgo (commented):
just a few last questions/thoughts/suggestions from me at this point.  nice turnaround on this!

internal/local/app_test.go
jandersongo: we could actually remove the `appData****Invalid` vars since the apps are not expected to be found (because config file doesn't contain a valid version)
Since the apps aren't found, there is no appData to compare theses vars with.

internal/local/app_test.go
jandersongo: difference between the invalid case and the empty case:
--> the empty case will encounter an error when `LoadConfig` tries to read/parse the file [here](https://github.com/10gen/realm-cli/blob/29820cfb35f206efdf45983381e4acf607d686d8/internal/local/app.go#L230).
--> the invalid case will have its config file parsed without error but will not be recognized as an app path since it doesn't contain a real config 
```
{
  "garbage": 1
}
```

internal/local/app.go
```diff
@@ -288,7 +284,17 @@ func FindApp(path string) (App, bool, error) {
 				}
 				return App{}, false, err
 			}
-			return App{RootDir: wd, Config: config}, true, nil
+
+			app := App{RootDir: wd, Config: config}
+			if err := app.LoadConfig(); err != nil {
```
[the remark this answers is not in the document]
jandersongo: it gets called in Load() [here](https://github.com/10gen/realm-cli/blob/29820cfb35f206efdf45983381e4acf607d686d8/internal/local/app.go#L205).

internal/commands/app/create_test.go
```diff
@@ -727,6 +727,7 @@ Check out your app: cd ./remote-app && realm-cli app describe
 						Project:         "123",
 						Location:        realm.LocationVirginia,
 						DeploymentModel: realm.DeploymentModelGlobal,
+						ConfigVersion:   realm.AppConfigVersion20210101,
```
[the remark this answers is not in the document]
jandersongo: It was needed in order to have FindApp detect the testApp.
Changed to default value.

internal/commands/app/create_inputs_test.go
```diff
@@ -248,6 +248,12 @@ func TestAppCreateInputsResolveDirectory(t *testing.T) {
 		assert.Nil(t, cmd.Handler(profile, ui, cli.Clients{Realm: rc}))
 
 		existingDir := filepath.Join(profile.WorkingDirectory, existingApp.Name)
+		assert.Nil(t, ioutil.WriteFile(
```
makesitgo, line 251: these are odd existing tests b/c it seems like we're using the command `Handler` in order to setup the test env for the comamnd's `inputs.resolveLocalPath` to fail

so I guess my question is: what broke here?  I would guess that the `CommandCreate.Handler` would still be creating the config version like you're writing it here?

internal/commands/push/inputs_test.go
```diff
@@ -65,13 +66,13 @@ func TestPushInputsResolve(t *testing.T) {
...
 
-	t.Run("Should set the app data if no flags are set but is run from inside a project directory", func(t *testing.T) {
+	t.Run("should set the app data if no flags are set but is run from inside a project directory", func(t *testing.T) {
 		profile, teardown := mock.NewProfileFromTmpDir(t, "app_init_input_test")
 		defer teardown()
 
 		assert.Nil(t, ioutil.WriteFile(
 			filepath.Join(profile.WorkingDirectory, local.FileConfig.String()),
-			[]byte(`{"app_id": "eggcorn-abcde", "name":"eggcorn"}`),
+			[]byte(fmt.Sprintf(`{"config_version":%d,"app_id": "eggcorn-abcde", "name":"eggcorn"}`, realm.DefaultAppConfigVersion)),
```
makesitgo, line 75: we can be civilized here still :)
```suggestion
			[]byte(fmt.Sprintf(`{"config_version": %d, "app_id": "eggcorn-abcde", "name":"eggcorn"}`, realm.DefaultAppConfigVersion)),
```

internal/local/app.go
```diff
@@ -221,10 +225,11 @@ func (a *App) LoadConfig() error {
 		a.AppData = &AppConfigJSON{}
 	case FileStitch:
 		a.AppData = &AppStitchJSON{}
+	default:
+		return errInvalidConfigFile(a.Config.String())
```
makesitgo, line 229: considering the invalid config file error is only used again in a test, I think we can avoid the helper/abstraction for it and simplify things a bit in the process:

```go
default:
  return fmt.Errorf("invalid config file: %s", a.Config)
```

and then simply in the test assert things equal `errors.New("invalid config file: bogus.json")`

internal/local/app.go
```diff
@@ -259,10 +264,6 @@ func LoadAppConfig(path string) (App, error) {
 	if !appOK {
 		return App{}, nil
 	}
-
-	if err := app.LoadConfig(); err != nil {
-		return App{}, err
-	}
 	return app, nil
```
makesitgo, line 267: what do you think about removing the `LoadAppConfig` function entirely?  It's not doing much besides handling the `!ok` case of `FindApp`.  considering there's only a handful of usages at this point, it seems like a good chance to reduce the exported surface area of our `local` package

internal/local/app_test.go
```diff
@@ -83,6 +83,36 @@ func TestLoadApp(t *testing.T) {
...
+		{realm.AppConfigVersion20180301, FileStitch},
+		{realm.AppConfigVersion20200603, FileConfig},
+		{realm.AppConfigVersion20210101, FileRealmConfig},
+	} {
+		t.Run(fmt.Sprintf("should successfully load config file with version %d", config.version), func(t *testing.T) {
+			projectRoot := filepath.Join(wd, "testdata", config.version.String(), "local")
+			testApp := App{RootDir: projectRoot, Config: config.file}
+
+			err := testApp.LoadConfig()
+			assert.Nil(t, err)
```
makesitgo, line 103: [style/meganit] keeps a tighter scope if you inline these types of assertions
```suggestion
			assert.Nil(t, testApp.LoadConfig())
```

it honestly _shouldnt_ matter when we write perfect code...

but ive been burned enough times (read: once) by the following:

1) re-assigning to an `err` variable in a test
2) failing to assert on it
3) the compiler never notices the "un-used" var (b/c it was re-assigned with `=`, not newly assigned with `:=`)
4) i'm left wondering why the heck my test isn't catching the error I expect to be returned 😬 

that's why i now tend to push for the tightly scoped vars whenever possible

internal/local/app_test.go
makesitgo: i may just be tripped up on names here... correct me if i'm wrong here, but the "empty case" fails b/c we're trying to parse invalid json syntax, right?  and the "invalid case" doesn't error but fails to find app b/c the JSON that was parsed simply didn't have "config_version" as a field?

if that's so, maybe we can just tweak the naming in the test here: where "empty" is actually "invalid json" and "invalid" is actually "missing config version"?  curious as to your thoughts on that one

makesitgo (commented):
last few bits from me here... should be good to go afterwards

internal/commands/app/create_inputs_test.go
jandersongo: you're right. the command `Handler` was creating the config file. The issue was there was no config version being written to it so `local.FindApp` couldn't recognize it. 
Opted for simply adding the ConfigVersion to the createInputs instead of writing the config file.

internal/local/app_test.go
jandersongo: agreed! I'll change it

internal/local/app.go
jandersongo: these methods are fairly intertwined: `Load()`, `LoadConfig()`, `LoadApp()`, `LoadAppConfig()`
I think we could survive with just: `LoadApp()` and `LoadConfig()` [called by `LoadApp()` and `FindApp()`]

pushed a commit with the changes, lmk what you think!

makesitgo (commented):
just one larger comment left that I'd like to get your thoughts on first but otherwise this is 🔥

internal/commands/app/create.go
```diff
@@ -316,7 +316,7 @@ func (cmd CommandCreate) handleCreateApp(
 		return err
 	}
 
-	if err := appLocal.Load(); err != nil {
+	if err := appLocal.AppData.LoadData(rootDir); err != nil {
```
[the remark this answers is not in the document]
jandersongo: considering `FindApp` will return` !ok` if the app doesn't exist at the path specified by `LoadApp(path string)`, I think it would be appropriate to return an error

internal/commands/app/create.go
```diff
@@ -316,7 +316,7 @@ func (cmd CommandCreate) handleCreateApp(
 		return err
 	}
 
-	if err := appLocal.Load(); err != nil {
+	if err := appLocal.LoadData(rootDir); err != nil {
```
makesitgo, line 319: [nit] I think it's the same result, but could we use `appLocal.LoadData(appLocal.RootDir)` here?

internal/commands/app/create_test.go
```diff
@@ -379,8 +379,9 @@ Check out your app: cd ./test-app && realm-cli app describe
 
 		assert.Nil(t, cmd.Handler(profile, ui, cli.Clients{Realm: realmClient, Atlas: atlasClient}))
 
-		_, err = local.LoadApp(filepath.Join(profile.WorkingDirectory, cmd.inputs.Name))
-		assert.Nil(t, err)
+		appPath := filepath.Join(profile.WorkingDirectory, cmd.inputs.Name)
+		_, err = local.LoadApp(appPath)
+		assert.Equal(t, err, local.ErrFailedToFindApp(appPath))
```
makesitgo, line 384: rather than have to export the error, i'd just assert equal to `errors.New(...)` (with the expected string)

internal/commands/app/diff_test.go
```diff
@@ -105,7 +106,7 @@ func TestAppDiffHandler(t *testing.T) {
 		_, ui := mock.NewUI()
 
 		cmd := &CommandDiff{diffInputs{LocalPath: "./some/path"}}
-		assert.Equal(t, errors.New("no app directory found at ./some/path"), cmd.Handler(nil, ui, cli.Clients{}))
+		assert.Equal(t, local.ErrFailedToFindApp("./some/path"), cmd.Handler(nil, ui, cli.Clients{}))
```
makesitgo, line 109: same here

internal/local/app_test.go
```diff
@@ -68,19 +69,57 @@ func TestNewApp(t *testing.T) {
...
+	}{
+		{realm.AppConfigVersion20180301, FileStitch},
+		{realm.AppConfigVersion20200603, FileConfig},
+		{realm.AppConfigVersion20210101, FileRealmConfig},
+	} {
+		t.Run(fmt.Sprintf("should successfully load config file with version %d", config.version), func(t *testing.T) {
+			projectRoot := filepath.Join(wd, "testdata", config.version.String(), "local")
+			testApp := App{RootDir: projectRoot, Config: config.file}
+
+			assert.Nil(t, testApp.LoadConfig())
```
makesitgo, line 113: are there tests/asserts on `testApp` we could include here just for completeness?

makesitgo (approved):
last few things from me but nothing blocking... great work on this!

🆒  beans

PR #255, opened 2022-02-11
Revert "REALMC-11543: Use nil map in test data source's schema and relationships"

jandersongo:
Reverts 10gen/realm-cli#245


makesitgo (approved):
🆒  beans

PR #267, opened 2022-02-22
REALMC-10362: Update CLI to import schemas separate from rules

malexandert:
JIRA: https://jira.mongodb.org/browse/REALMC-10362

Essentially doing a similar thing to [this PR](https://github.com/10gen/baas/pull/5848). After this PR, pushing, pulling, and diffing an app should work in cases where:
* a collection folder (`data_sources/<data_source>/<database>/<collection>`) contains solely a `rules.json` file
* a collection folder contains solely a pair of `schema.json` and `relationships.json` files
This change was necessary because, as part of Schema Separation, an app is allowed to have a rule without an associated schema and vice versa

### DIFF
--- a/internal/local/structure_v2.go
+++ b/internal/local/structure_v2.go
@@ -34,6 +34,7 @@ type AppStructureV2 struct {
 	Sync                  SyncStructure                     `json:"sync,omitempty"`
 	Secrets               SecretsStructure                  `json:"secrets,omitempty"`
 	LogForwarders         []map[string]interface{}          `json:"log_forwarders,omitempty"`
+	Schemas               []map[string]interface{}          `json:"schemas,omitempty"`
 }
 
 // AuthStructure represents the v2 Realm app auth structure
@@ -163,11 +164,12 @@ func (a *AppDataV2) LoadData(rootDir string) error {
 	}
 	a.Services = services
 
-	dataSources, err := parseDataSources(rootDir)
+	dataSources, schemas, err := parseDataSources(rootDir)
 	if err != nil {
 		return err
 	}
 	a.DataSources = dataSources
+	a.Schemas = schemas
 
 	httpServices, err := parseHTTPServices(rootDir)
 	if err != nil {
@@ -272,8 +274,9 @@ func parseEndpointsV2(rootDir string) (EndpointStructure, error) {
 	return EndpointStructure{configs}, nil
 }
 
-func parseDataSources(rootDir string) ([]DataSourceStructure, error) {
+func parseDataSources(rootDir string) ([]DataSourceStructure, []map[string]interface{}, error) {
 	var out []DataSourceStructure
+	var schemas []map[string]interface{}
 
 	dw := directoryWalker{
 		path:     filepath.Join(rootDir, NameDataSources),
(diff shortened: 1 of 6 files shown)


internal/local/structure_v2.go
```diff
@@ -292,33 +295,55 @@ func parseDataSources(rootDir string) ([]DataSourceStructure, error) {
...
-				rule[NameSchema] = schema
 
 				relationships, err := parseJSON(filepath.Join(collPath, FileRelationships.String()))
 				if err != nil {
 					return err
 				}
-				rule[NameRelationships] = relationships
 
-				rules = append(rules, rule)
+				if rule == nil && (schemaBody == nil || relationships == nil) {
```
makesitgo, line 322: at the risk of extra indentation, we might be able to consolidate some of these conditionals

```go
if rule == nil {
  if schema == nil || relationships == nil { return fmt.Errorf(...) }
  schemas = append(schemas, map[string]interface{}{
    NameSchema: schema,
    NameRelationships: relationships,
    NameMetadata: ...,
  })
} else {
  if len(rule) == 0 { return errors.New(...) }
  rule[NameSchema] = schema
  rule[NameRelationshpis] = relationships,
}
```

[question] was it intentional that you dropped the `rule` assignments for schema and relationships when `rule != nil`?

internal/local/structure_v2.go
```diff
@@ -534,6 +562,70 @@ func writeSync(rootDir string, sync SyncStructure) error {
...
+		relationships := schema[NameRelationships]
+		if relationships == nil {
+			relationships = map[string]interface{}{}
+		}
+		dataRelationships, err := MarshalJSON(relationships)
+		if err != nil {
+			return err
+		}
+
+		metadata := schema["metadata"].(map[string]interface{})
```
makesitgo, line 588: if you're going to use `.(map[string]interface{})` then you should accompany it with the `, ok` syntax to avoid potential panics

internal/local/structure_v2.go
```diff
@@ -534,6 +562,70 @@ func writeSync(rootDir string, sync SyncStructure) error {
...
+		if metadata == nil {
+			metadata = map[string]interface{}{}
+		}
+		var dataSource, database, collection string
+		if ds, ok := metadata["data_source"]; ok {
```
makesitgo, line 593: you could probably simplify all of this down to:

```go
dataSource, _ := metadata["data_source"].(string)
database, _ := metadata["database"].(string)
collection, _ := metadata["collection"].(string)
```

internal/local/structure_v2.go
```diff
@@ -534,6 +562,70 @@ func writeSync(rootDir string, sync SyncStructure) error {
...
+				database = db
+			}
+		}
+		if coll, ok := metadata["collection"]; ok {
+			if coll, ok := coll.(string); ok {
+				collection = coll
+			}
+		}
+
+		schemaDir := filepath.Join(dir, dataSource, database, collection)
```
makesitgo, line 609: you probably need to `os.MkdirAll` on this path

internal/local/structure_v2.go
```diff
@@ -556,29 +648,7 @@ func writeDataSources(rootDir string, dataSources []DataSourceStructure) error {
 			return err
 		}
 		for _, rule := range ds.Rules {
-			schema := rule[NameSchema]
```
makesitgo, line 559: I think we still want to keep this code path here, if the `rule` happens to have these things, `Write` should account for it

makesitgo (commented):
initial batch of thoughts/comments from me.

my main concern is I think I'd have expected these changes to be more "additive" than they currently are, and I want to make sure we're not breaking the CLI for any older apps out there

internal/local/structure_v2.go
```diff
@@ -534,6 +569,64 @@ func writeSync(rootDir string, sync SyncStructure) error {
...
+		relationships := schema[NameRelationships]
+		if relationships == nil {
+			relationships = map[string]interface{}{}
+		}
+		dataRelationships, err := MarshalJSON(relationships)
+		if err != nil {
+			return err
+		}
+
+		metadata, ok := schema["metadata"].(map[string]interface{})
```
jandersongo, line 595: [nit] can we use `NameMetadata` instead of "metadata"

jandersongo (approved):
lgtm!

internal/local/structure_v2.go
```diff
@@ -272,8 +274,9 @@ func parseEndpointsV2(rootDir string) (EndpointStructure, error) {
 	return EndpointStructure{configs}, nil
 }
 
-func parseDataSources(rootDir string) ([]DataSourceStructure, error) {
+func parseDataSources(rootDir string) ([]DataSourceStructure, []map[string]interface{}, error) {
 	var out []DataSourceStructure
```
caiyingshen, line 278: [nit] now that this function returns two values, could this be changed to `dataSources` or something describing the object?

internal/local/structure_v2.go
```diff
@@ -292,33 +295,62 @@ func parseDataSources(rootDir string) ([]DataSourceStructure, error) {
...
 			colls := directoryWalker{path: dbPath, onlyDirs: true}
 			if err := colls.walk(func(coll os.FileInfo, collPath string) error {
+				// A valid data sources folder contains at least one of:
+				// - a rules.json file
+				// - a pair of files, schema.json and relationships.json
+				// If neither of these conditions are true (e.g. there is only a schema.json
+				// file, or there are no files in the directory at all), we should error
 
-				rulePath := filepath.Join(collPath, FileRules.String())
-				if _, err := os.Stat(rulePath); err != nil {
```
caiyingshen, line 297: [opt] could we use `Stat()` to check the file paths to see if the conditions for a valid data source folder are met before parsing them? up to you, I'm not sure how much of a performance improvement this would be

internal/local/structure_v2.go
```diff
@@ -534,6 +569,64 @@ func writeSync(rootDir string, sync SyncStructure) error {
...
+
+		metadata, ok := schema["metadata"].(map[string]interface{})
+		if !ok {
+			return errors.New("error writing schemas")
+		}
+		if metadata == nil {
+			metadata = map[string]interface{}{}
+		}
+
+		dataSource, _ := metadata["data_source"].(string)
```
caiyingshen, line 603: [nit] are there constants we could use for `data_source`, `database`, and `collection`?

caiyingshen (approved):
LGTM 😎  just a couple questions and nits from me!

makesitgo (commented):
just a last few questions from me here before this is ready to go... nice work!

internal/local/structure_v2.go
```diff
@@ -601,19 +607,35 @@ func writeDataSources(rootDir string, dataSources []DataSourceStructure) error {
...
 			}
-			if err := WriteFile(
-				filepath.Join(ruleDir, FileSchema.String()),
-				0666,
-				bytes.NewReader(dataSchema),
-			); err != nil {
-				return err
+
+			schema := rule[NameSchema]
+			if schema != nil {
```
makesitgo, line 612: let's use the `if schema, ok := rule[NameSchema]; ok {` syntax here (we avoided it previously b/c we always wanted to replace `nil` with `map[string]interface{}{}` which i suppose is no longer the case)

internal/local/structure_v2.go
```diff
@@ -601,19 +607,35 @@ func writeDataSources(rootDir string, dataSources []DataSourceStructure) error {
...
+				dataSchema, err := MarshalJSON(schema)
```
makesitgo, line 613: [meganit/naming] could in theory tighten the scope of this named `var` to just `data` now that we're in this `okSchema` closure

makesitgo (approved):
one last question about our ability within `writeDataSources` to not have to write a `rules.json` file (which at least currently, shouldn't come up)

otherwise, nothing blocking on this from me, nice work!

🆒  beans

PR #279, opened 2022-03-21
REALMC-11800: Use root directory to create app on nested push command

NBSquare:
[Ticket here](https://jira.mongodb.org/browse/REALMC-11800)

I added one test to verify that the `realm_config` was being created properly when running the command. I couldn't figure out a clean way to teardown the test, though, since the push handler tests all use the existing `testdata/project` directory for the environment.

Mocking out the `createNewApp` method might be a nice way to do this, since we could just verify that the method's being called with the right params, but it didn't seem like that's something we support in this repo.

### DIFF
--- a/internal/commands/push/command_test.go
+++ b/internal/commands/push/command_test.go
@@ -1027,6 +1027,47 @@ Try instead: realm-cli push --local testdata/dependencies --remote appID --inclu
 
 		assert.Nil(t, err)
 	})
+
+	t.Run("with local path in nested directory", func(t *testing.T) {
+		var realmClient mock.RealmClient
+		realmClient.FindAppsFn = func(filter realm.AppFilter) ([]realm.App, error) {
+			return []realm.App{{GroupID: "groupID"}}, nil
+		}
+		realmClient.CreateAppFn = func(groupID, name string, meta realm.AppMeta) (realm.App, error) {
+			return realm.App{
+				GroupID: groupID,
+				ID:      primitive.NewObjectID().Hex(),
+				Name:    name,
+				AppMeta: meta,
+			}, nil
+		}
+		realmClient.DiffFn = func(groupID, appID string, appData interface{}) ([]string, error) {
+			return []string{}, nil
+		}
+
+		out := new(bytes.Buffer)
+		ui := mock.NewUIWithOptions(mock.UIOptions{AutoConfirm: true}, out)
+
+		cmd := &Command{inputs{LocalPath: filepath.Join(testApp.RootDir, "functions"), RemoteApp: "appID"}}
+		err := cmd.Handler(nil, ui, cli.Clients{Realm: realmClient})
+		defer os.Remove(filepath.Join(testApp.RootDir, local.FileRealmConfig.String()))
+
+		assert.Nil(t, err)
+		configData, readErr := ioutil.ReadFile(filepath.Join(testApp.RootDir, local.FileRealmConfig.String()))
+
+		assert.Nil(t, readErr)
+		assert.Equal(t, `{
+    "config_version": 20210101,
+    "name": "eggcorn",
+    "location": "US-VA",
+    "deployment_model": "GLOBAL"
+}
+`, string(configData))
+
+		// Verify that a config file was not created in the nested directory, either.
+		_, fileErr := os.Stat(filepath.Join(testApp.RootDir, "functions", local.FileRealmConfig.String()))
+		assert.NotNil(t, fileErr)
+	})
 }
 
 func TestPushCommandCreateNewApp(t *testing.T) {
--- a/internal/commands/push/command_test.go
+++ b/internal/commands/push/command_test.go
@@ -1035,11 +1035,11 @@ Try instead: realm-cli push --local testdata/dependencies --remote appID --inclu
 		}
 		realmClient.CreateAppFn = func(groupID, name string, meta realm.AppMeta) (realm.App, error) {
 			return realm.App{
-				GroupID: groupID,
+				GroupID:     groupID,
 				ClientAppID: "eggcorn-abcde",
-				ID:      primitive.NewObjectID().Hex(),
-				Name:    name,
-				AppMeta: meta,
+				ID:          primitive.NewObjectID().Hex(),
+				Name:        name,
+				AppMeta:     meta,
 			}, nil
 		}
 		realmClient.DiffFn = func(groupID, appID string, appData interface{}) ([]string, error) {
--- a/internal/commands/push/command_test.go
+++ b/internal/commands/push/command_test.go
@@ -1036,6 +1036,7 @@ Try instead: realm-cli push --local testdata/dependencies --remote appID --inclu
 		realmClient.CreateAppFn = func(groupID, name string, meta realm.AppMeta) (realm.App, error) {
 			return realm.App{
 				GroupID: groupID,
+				ClientAppID: "eggcorn-abcde",
 				ID:      primitive.NewObjectID().Hex(),
 				Name:    name,
 				AppMeta: meta,
@@ -1058,6 +1059,7 @@ Try instead: realm-cli push --local testdata/dependencies --remote appID --inclu
 		assert.Nil(t, readErr)
 		assert.Equal(t, `{
     "config_version": 20210101,
+    "app_id": "eggcorn-abcde",
     "name": "eggcorn",
     "location": "US-VA",
     "deployment_model": "GLOBAL"
--- a/internal/commands/push/command_test.go
+++ b/internal/commands/push/command_test.go
@@ -1066,7 +1066,7 @@ Try instead: realm-cli push --local testdata/dependencies --remote appID --inclu
 
 		// Verify that a config file was not created in the nested directory, either.
 		_, fileErr := os.Stat(filepath.Join(testApp.RootDir, "functions", local.FileRealmConfig.String()))
-		assert.NotNil(t, fileErr)
+		assert.True(t, os.IsNotExist(fileErr), "expected nested config path to not exist, but fileErr was: %s", fileErr)
 	})
 }
 
--- a/internal/commands/push/command_test.go
+++ b/internal/commands/push/command_test.go
@@ -1127,7 +1127,7 @@ func TestPushHandlerCreateNewApp(t *testing.T) {
 `,
 		},
 	} {
-		t.Run(fmt.Sprintf("with a %s config file", tc.appConfig.String()), func(t *testing.T) {
+		t.Run(fmt.Sprintf("with a %s config file and nested local path", tc.appConfig.String()), func(t *testing.T) {
 			tmpDir, teardown, tmpDirErr := u.NewTempDir("push_handler")
 			assert.Nil(t, tmpDirErr)
 			defer teardown()
@@ -1148,6 +1148,9 @@ func TestPushHandlerCreateNewApp(t *testing.T) {
 			configData, readErr := ioutil.ReadFile(filepath.Join(tmpDir, tc.appConfig.String()))
 			assert.Nil(t, readErr)
 			assert.Equal(t, tc.expectedConfig, string(configData))
+
+			_, fileErr := os.Stat(filepath.Join(tmpDir, "nested", tc.appConfig.String()))
+			assert.True(t, os.IsNotExist(fileErr), "expected nested config path to not exist, but err was: %s", err)
 		})
 	}
 }
--- a/internal/commands/push/command_test.go
+++ b/internal/commands/push/command_test.go
@@ -1050,35 +1050,22 @@ func TestPushHandlerUseCreateNewApp(t *testing.T) {
 		return []string{}, nil
 	}
 
-	t.Run("with a realm_config.json", func(t *testing.T) {
-		tmpDir, teardown, tmpDirErr := u.NewTempDir("push_handler")
-		assert.Nil(t, tmpDirErr)
-		defer teardown()
-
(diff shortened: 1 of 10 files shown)


jandersongo (changes requested):
a solid start! a few comments from me...

internal/commands/push/command_test.go
```diff
@@ -1027,6 +1027,47 @@ Try instead: realm-cli push --local testdata/dependencies --remote appID --inclu
...
+		realmClient.DiffFn = func(groupID, appID string, appData interface{}) ([]string, error) {
+			return []string{}, nil
+		}
+
+		out := new(bytes.Buffer)
+		ui := mock.NewUIWithOptions(mock.UIOptions{AutoConfirm: true}, out)
+
+		cmd := &Command{inputs{LocalPath: filepath.Join(testApp.RootDir, "functions"), RemoteApp: "appID"}}
+		err := cmd.Handler(nil, ui, cli.Clients{Realm: realmClient})
+		defer os.Remove(filepath.Join(testApp.RootDir, local.FileRealmConfig.String()))
```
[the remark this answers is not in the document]
NBSquare: Yeah, I saw this... The problem we have here is a problem in `createNewApp`. [This line](https://github.com/10gen/realm-cli/blob/master/internal/commands/push/command.go#L478) creates a new config file based on the default app config version, which means if the CLI can't resolve an app it's always going to create a new config file.

Options I can think of here are:
- Updating `createNewApp` to write to the "correct" config file (i.e. if local has `stitch.json`, write to `stitch.json`). I'm guessing this might lead to compatibility issues but I'm not sure.
- Bump the project version for this test suite and use `realm_config.json`. I don't think the suite is too dependent on the old project verison so this oculd be a quick and easy way to set this up.
- Set up this test in a temp directory which we can clean up after the test finishes. Also a good solution but would involve dynamically generating a `realm_config.json`
- Mock the call to `command.createNewApp` altogether. This is the cleanest, best solution IMO, but I'm not sure the CLI repo has a library for mocking/spying yet.

internal/commands/push/command_test.go
```diff
@@ -1027,6 +1027,47 @@ Try instead: realm-cli push --local testdata/dependencies --remote appID --inclu
...
+
+		assert.Nil(t, err)
+		configData, readErr := ioutil.ReadFile(filepath.Join(testApp.RootDir, local.FileRealmConfig.String()))
+
+		assert.Nil(t, readErr)
+		assert.Equal(t, `{
+    "config_version": 20210101,
```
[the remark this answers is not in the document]
NBSquare: This test uses a mocked `realmClient` and I didn't add the `ClientAppId` - I'll add that real quick.

internal/commands/push/command_test.go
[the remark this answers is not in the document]
jandersongo: > Bump the project version for this test suite and use realm_config.json. I don't think the suite is too dependent on the old project version so this could be a quick and easy fix.

We still want to support old cli versions -- we should actually add tests for each 

> Updating createNewApp to write to the "correct" config file (i.e. if local has stitch.json, write to stitch.json). I'm guessing this might lead to compatibility issues but I'm not sure.

This seems like the appropriate solution to me. Ideally, if the user has a config file they are trying to push, we would keep the version they're using. 
Maybe @makesitgo can chime in here.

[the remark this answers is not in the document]
makesitgo: so my first though is: perhaps we make a new top-level test func for this: `TestPushCommandCreateNewAppNested(t *testing.T)` (there's a lot in this test file already... there's no harm in cornering off a little section of our own for this fix)

as far as cleanup/teardown goes, check out the test table's implementation found in `TestPushCommandCreateNewApp` (around L#1150 for my local `master`).  something tells me we can test the push command handler for your fix here by using `mock.NewProfileFromTmpDir(...)` (the usages of that function may prove useful to look at as well)

utlimately, I don't think we need to do what `TestPushCommandCreateNewApp` is doing and make our new tests "interactive"... but I do think it'd be worthwhile to sort of follow what `TestPushHandler` and run this flow in a more "end-to-end" fashion by executing `*Command.Handler` (and specify `AutoConfirm`)

I like jamie's thoughts on running this test for all config versions, that'd be a good thing to try and achieve here

---

also as a complete aside, we don't really need any library or package for mocking/spying things in `go`... we can just write some oddly-syntaxed things:

```go
// instead of
func createNewApp(...) {
  // func impl here
}

// you could write
var createNewApp = func(...) {
  // same func impl here
}

// the funcs can then be consumed/used the same way

// but in a test, you have a change to replace it
func TestSomething(t *testing.T) {
  origCreateNewApp := createNewApp
  createNewApp = func(...) {
    // some mock or spy impl
  }
  // and in this test/mock pattern, it's always good to remember to cleanup afterwards
  defer func() { createNewApp = origCreateNewApp }()
}
```

internal/commands/push/command_test.go
```diff
@@ -1029,11 +1029,170 @@ Try instead: realm-cli push --local testdata/dependencies --remote appID --inclu
...
+
+		app.WriteConfig()
+
+		out := new(bytes.Buffer)
+		console, ui, consoleErr := mock.NewConsoleWithOptions(mock.UIOptions{AutoConfirm: true}, out)
+		assert.Nil(t, consoleErr)
+		defer console.Close()
+
+		// Run handler method on both root directory and nested directory.
+		for _, localPath := range []string{tmpDir, filepath.Join(tmpDir, "nested_directory")} {
```
jandersongo, line 1074: instead of this sort of for loop, can we make this a test table with proper descriptions?

```
for _, tc := range []struct {
	description     string
	localPath          string
}{
	{
		description: "....",
		localPath:    tmpDir,
	},
	{
		description:       "...",
		localPath:    filepath.Join(tmpDir, "nested_directory"),
	},
},
```

jandersongo: on second thought, I don't think it's necessary for us to test creating a new app from the root dir. (you can remove that for loop on line 1074)

But we could condense all three of these cases ("with a realm_config.json", "with a config.json", and "with a stitch.json") into a test table 

```
for _, tc := range []struct {
	description          string
	configVersion        realm.AppConfigVersion
        configFile           File
        configData           string 
}{
	...
},
```

jandersongo (changes requested):

internal/commands/push/command_test.go
```diff
@@ -1029,11 +1029,170 @@ Try instead: realm-cli push --local testdata/dependencies --remote appID --inclu
 	})
 }
 
+func TestPushHandlerUseCreateNewApp(t *testing.T) {
```
jandersongo, line 1032: this func name could be a bit more descriptive for what we're testing -- pushing from nested dirs.
I like nick's previous suggestion `TestPushCommandCreateNewAppNested`

jandersongo (changes requested):
nice progress! a few more comments from me ...

internal/commands/push/command_test.go
```diff
@@ -1180,13 +1339,30 @@ func TestPushCommandCreateNewApp(t *testing.T) {
...
-					description: "should use the package name when present and zero values for app meta",
-					appData:     local.AppConfigJSON{local.AppDataV1{local.AppStructureV1{Name: "name"}}},
+					description:       "should use the package name when present and zero values for app meta",
+					appData:           local.AppConfigJSON{local.AppDataV1{local.AppStructureV1{ConfigVersion: realm.AppConfigVersion20200603, Name: "name"}}},
+					expectedAppConfig: local.FileConfig,
+				},
+				{
+					description:       "should use realm_config when config version is not specified in appData",
+					appData:           &local.AppConfigJSON{local.AppDataV1{local.AppStructureV1{Name: "name"}}},
+					expectedAppConfig: local.FileRealmConfig,
```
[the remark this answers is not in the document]
jandersongo: this does bring up a bit of a concern. we probably shouldn't be able to pass in a config version that doesn't coincide with the app structure version. 🤔

internal/commands/push/command_test.go
NBSquare: Pushing into a nested directory doesn't seem entirely descriptive to me either... This test feels more like an integration test verifying that the handler calls create new app correctly (i.e. writes to the config file as expected).

internal/commands/push/command_test.go
```diff
@@ -1180,13 +1339,30 @@ func TestPushCommandCreateNewApp(t *testing.T) {
...
+				},
+				{
+					description:       "should downgrade to stitch config when specified in appData",
+					appData:           local.AppRealmConfigJSON{local.AppDataV2{local.AppStructureV2{ConfigVersion: realm.AppConfigVersion20180301, Name: "name"}}},
+					expectedAppConfig: local.FileStitch,
+				},
```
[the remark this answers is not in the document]
NBSquare: Removing this test as cross-version pushing is now unexpected behavior.

internal/commands/push/command_test.go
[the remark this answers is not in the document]
NBSquare: Removing this test as well.

internal/commands/push/command_test.go
makesitgo: I wouldn't go too crazy on names here... the only way these functions are "called" is through `go test...` and for my own laziness, I like keeping the names hierarchical so my `go test` regexes are easy/simple to write (read: putting that "Use" in the middle of the name makes it harder for me to potentially run _just_ the tests related to `CreateNewApp`)

I'd agree `...CreateNewAppNested` sounds funny... but again, for ease of regex-ing it allows for meaningful suites of tests to be run via:
`-run 'TestPushCommand'`
`-run 'TestPushCommandCreateNewApp'`
`-run 'TestPushCommandCreateNewAppNested'`

that middle one is :neat: b/c like you said, this is more or less a test that makes sure `Handler` is invoking `createNewApp` the proper way... so we'd still be able to test all things related to `createNewApp` that way

internal/commands/push/command_test.go
```diff
@@ -1029,11 +1029,136 @@ Try instead: realm-cli push --local testdata/dependencies --remote appID --inclu
...
+			app := local.App{RootDir: tmpDir, Config: tc.appConfig, AppData: tc.appData}
+
+			app.WriteConfig()
+
+			out := new(bytes.Buffer)
+			console, ui, consoleErr := mock.NewConsoleWithOptions(mock.UIOptions{AutoConfirm: true}, out)
+			assert.Nil(t, consoleErr)
+			defer console.Close()
+
+			cmd := &Command{inputs{LocalPath: tmpDir, RemoteApp: "appID"}}
```
jandersongo, line 1145: should be testing with the nested dir here instead of `tmpDir`

internal/commands/push/command_test.go
NBSquare: Using `TestPushHandlerCreateNewApp` since this is verifying the integration of handler and create new app at a high level.

internal/local/app.go
```diff
@@ -49,6 +49,20 @@ func (a App) Option() string {
 	return a.AppData.Name()
 }
 
+// hasValidConfigVersion returns whether or not the config file and version number match
```
jandersongo, line 52: [nit] prob doesn't need a comment anymore since the func is un-exported but feel free to leave it

internal/commands/push/command_test.go
```diff
@@ -1029,11 +1029,135 @@ Try instead: realm-cli push --local testdata/dependencies --remote appID --inclu
 	})
 }
 
+func TestPushHandlerCreateNewApp(t *testing.T) {
```
jandersongo, line 1032: [mega-nit] we are testing the version/appStructure functionality in `internal/local/app_test.go` with `"mismatched version in the config file should fail to find app"` so it might make sense to rename this to `TestPushHandlerCreateNewAppNested` but as @makesitgo likes to say, "I won't die on this hill"

internal/commands/push/command_test.go
```diff
@@ -1029,11 +1029,135 @@ Try instead: realm-cli push --local testdata/dependencies --remote appID --inclu
...
+			out := new(bytes.Buffer)
+			console, ui, consoleErr := mock.NewConsoleWithOptions(mock.UIOptions{AutoConfirm: true}, out)
+			assert.Nil(t, consoleErr)
+			defer console.Close()
+
+			cmd := &Command{inputs{LocalPath: filepath.Join(tmpDir, "nested"), RemoteApp: "appID"}}
+			err := cmd.Handler(nil, ui, cli.Clients{Realm: realmClient})
+
+			assert.Nil(t, err)
+			configData, readErr := ioutil.ReadFile(filepath.Join(tmpDir, tc.appConfig.String()))
```
jandersongo, line 1148: [nit/opt] I know this check tells us that the config file was written in the app's root directory but it might also be worth checking that no config file was written to the nested dir 

before this fix, it was previously possible to have an existing config in the root dir and then on `push --local <nested>`, an additional config would be written to the nested dir

jandersongo (approved):
nothing blocking, just a couple more suggestions. 
really great work on this!

internal/commands/push/command_test.go
NBSquare: Before this PR, even if an app had a `config.json` with the correct version number, `push` would write to a `realm_config.json`.
So this test is still checking that if the user has a correctly configured `config.json` that the command writes to the same `config.json`.

internal/local/app.go
NBSquare: Is the placement of the function in this file OK? I'm not sure if we have any conventions around putting private functions towards the bottom of the file or anything like that.

internal/commands/push/command_test.go
NBSquare: Oh yeah, I forgot that was happening. I'll add that!

PR #281, opened 2022-03-23
REALMC-12163: Avoid creating empty schema/relationships when importing a directory with just a rule

malexandert:
JIRA: https://jira.mongodb.org/browse/REALMC-12163


makesitgo (approved):
code looks right, but there's not much context into the "why" of this change from JIRA. do you have any context for this?

otherwise:

🆒  beans

jandersongo (approved):

PR #282, opened 2022-03-24
REALMC-12348: Add confirmation prompt on createNewApp with push

NBSquare:
When creating a new app using the push command, we should give the user an opportunity to approve the configuration they've specified in the CLI.

At the same time, we don't need to confirm diff changes when creating a new app, since there's no remote app to compare diff against.


jandersongo (approved):

internal/commands/push/command.go
```diff
@@ -259,12 +259,14 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		return nil
 	}
 
-	proceed, err := ui.Confirm("Please confirm the changes shown above")
-	if err != nil {
-		return err
-	}
-	if !proceed {
-		return nil
+	if !isNewApp {
```
makesitgo, line 262: I think this confirmation is still valuable to have.. assuming we're not showing an empty diff

was there a particular reason we chose to skip this now in the event we're creating a new app?

internal/commands/push/command.go
```diff
@@ -468,6 +470,12 @@ func createNewApp(ui terminal.UI, realmClient realm.Client, appDirectory, groupI
 		}
 	}
 
+	if proceed, err := ui.Confirm("Are these settings correct?"); err != nil {
```
makesitgo, line 473: to stay consistent with our previous messaging... lets make this:

`"Please confirm the new app details shown above"`

internal/commands/push/command_test.go
```diff
@@ -1246,6 +1313,38 @@ func TestPushCommandCreateNewApp(t *testing.T) {
...
+						c.ExpectString("App Deployment Model")
+						c.Send(string(terminal.KeyArrowDown))
+						c.SendLine("")
+
+						c.ExpectString("App Environment")
+						c.Send(string(terminal.KeyArrowDown))
+						c.SendLine("")
+
+						c.ExpectString("Are these settings correct?")
+						c.SendLine("")
```
makesitgo, line 1339: remind me, is "blank" here by default saying "No"?  we should update the test case description here to reflect what we're actually doing

internal/commands/push/command_test.go
```diff
@@ -1153,6 +1153,70 @@ func TestPushHandlerCreateNewApp(t *testing.T) {
 			assert.True(t, os.IsNotExist(fileErr), "expected nested config path to not exist, but err was: %s", err)
 		})
 	}
+
+	t.Run("should skip diff when creating a new app", func(t *testing.T) {
```
makesitgo, line 1157: I'm am skeptical about the need for this test, this is probably related to my earlier question about the extra `if !createNewApp` check around the existing confirm diffs prompt

makesitgo (commented):
just a few questions/comments from me first

internal/commands/push/command_test.go
NBSquare: Yeah, it's a no. Just renamed the test, let me know if the new name looks good.

internal/commands/push/command_test.go
NBSquare: Yeah, resolved per our discussion. I've added testing paths for when the user receives a diff and asserting on both the output of the CLI as well as that the config file is created regardless of the diff.

internal/commands/push/command.go
NBSquare: Discussed on Slack - for reference, we've decided instead to log the diff (which we weren't doing when creating a new app), while including the additional prompting to confirm app config.

internal/commands/push/command.go
```diff
@@ -232,7 +232,7 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		return nil
 	}
 
-	if !ui.AutoConfirm() && !isNewApp {
+	if !ui.AutoConfirm() {
 		diffs := make([]string, 0, len(appDiffs)+1+hostingDiffs.Cap())
 
 		diffs = append(diffs, appDiffs...)
```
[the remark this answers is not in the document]
NBSquare: If there aren't any diffs we don't reach this point in the execution--see line 232.

internal/commands/push/command_test.go
```diff
@@ -1125,34 +1132,136 @@ func TestPushHandlerCreateNewApp(t *testing.T) {
...
-			cmd := &Command{inputs{LocalPath: filepath.Join(tmpDir, "nested"), RemoteApp: "appID"}}
-			err := cmd.Handler(nil, ui, cli.Clients{Realm: realmClient})
+				cmd := &Command{inputs{LocalPath: filepath.Join(tmpDir, "nested"), RemoteApp: "appID"}}
+				err := cmd.Handler(nil, ui, cli.Clients{Realm: realmClient})
 
-			assert.Nil(t, err)
-			configData, readErr := ioutil.ReadFile(filepath.Join(tmpDir, tc.appConfig.String()))
-			assert.Nil(t, readErr)
-			assert.Equal(t, tc.expectedConfig, string(configData))
+				assert.Nil(t, err)
```
makesitgo, line 1154: [style/nit] I think this was existing code, but should move this assert up one line along with the code that produces the `err` (alternatively, could inline it all: `assert.Nil(t, cmd.Handler(nil, ui, cli.Clients{Realm: realmClient}))`

internal/commands/push/command_test.go
```diff
@@ -1125,34 +1132,136 @@ func TestPushHandlerCreateNewApp(t *testing.T) {
...
+					console.ExpectEOF()
+				}()
+
+				cmd := &Command{inputs{LocalPath: tmpDir, RemoteApp: "appID"}}
+				err := cmd.Handler(nil, ui, cli.Clients{Realm: realmClient})
+
+				console.Tty().Close() // flush the writers
+				<-doneCh              // wait for procedure to complete
+
+				assert.Nil(t, err)
```
makesitgo, line 1259: likewise here, there are very few times (if ever) in `go` that shouldn't be immediately doing something with the `err` returned from a func, so this can be moved up (or inlined) with the `cmd.Handler` call itself

internal/commands/push/command_test.go
```diff
@@ -1246,6 +1358,38 @@ func TestPushCommandCreateNewApp(t *testing.T) {
 					},
 					expectedProceed: true,
 				},
+				{
+					description: "should return empty data if user rejects app configuration",
+					procedure: func(c *expect.Console) {
+
```
makesitgo, line 1364: [style/meganit] an errant newline here, can remove

makesitgo (approved):
nothing major from me... this looks good. nice work!

🆒  beans

PR #286, opened 2022-04-04
REALMC-12648: Bump version to 3.0.0

dpish:
bumped major version since behavior has changed in https://jira.mongodb.org/browse/REALMC-12348


makesitgo (changes requested):
that prompt is not a breaking change, the behavior of `push` is exactly the same with an extra confirmation prompt (which is bypassed in automated environments anyway with `-y`)

I won't argue much b/w `minor` or `patch` here (though initially I'm "team patch" here), but I would ask we avoid bumping to `3.0` for this.  again, the `push` command results in the same set of actions as before

PR #287, opened 2022-04-05
REALMC-12648: Bump version to 2.4.1

dpish:


makesitgo (approved):
I think you just need to update the commit message to reflect we're moving to `2.4.2` here

🆒  beans

PR #288, opened 2022-04-11
REALMC-12640: Parse app meta config

NBSquare:
[REALMC-12640](https://jira.mongodb.org/browse/REALMC-12640)
### Tasks Completed
- Defined `.mdb/app_meta.json` schema
- Modified `FindApp` to search for `.mdb/app_meta.json` when looking for root directory/config files
- Skipped group ID/app ID resolve calls whenever `AppMeta` is populated by a command
- Stubbed out `/push` command.

Note: I'm not sure what to do beyond just creating the `Push` and `DoPush` methods in `import.go`. We use empty interface to pass the POST body to the endpoint so I can't validate that config version is one of those fields, and because the endpoint isn't live yet I wasn't sure how to go about testing. I'd appreciate some input on this part of the PR.


internal/local/data.go
```diff
@@ -23,15 +23,27 @@ type AppData interface {
 	WriteData(rootDir string) error
 }
 
+// MdbAppMeta represents the .mdb/app_meta.json config
+type MdbAppMeta struct {
+	GroupID       string                 `json:"group_id,omitempty"`
```
jandersongo, line 28: [q] why are these fields optional? 
I think it makes sense for the `MdbAppMeta` struct to be optional for` local.App` but what reason do we have for making the GroupID/AppID optional? 
maybe this is more a q for @christineschen

internal/local/app_test.go
```diff
@@ -725,6 +726,12 @@ exports = function({ query }) {
 	},
 }}}
 
+var fullMeta = MdbAppMeta{
```
jandersongo, line 729: do we have tests for `local.LoadAppMeta`?

internal/local/app.go
```diff
@@ -265,26 +313,25 @@ func FindApp(path string) (App, bool, error) {
...
 	for i := 0; i < maxDirectoryContainSearchDepth; i++ {
-		for _, config := range allConfigFiles {
-			_, err := os.Stat(filepath.Join(wd, config.String()))
-			if err != nil {
-				if os.IsNotExist(err) {
-					continue
+		if _, err := os.Stat(filepath.Join(wd, FileMdbAppMeta.String())); err != nil {
+			if os.IsNotExist(err) {
+				// .mdb/app_meta.json may not exist, so we check whether a regular config file exists.
+				if app, proceed, err := checkForConfigFile(wd); proceed || err != nil {
```
jandersongo, line 319: we should still be "checking for the config file" regardless of whether there is ./mdb/app_meta.json or not. 
We need to `LoadConfig`, regardless.

internal/local/app.go
```diff
@@ -251,6 +271,34 @@ func (a *App) LoadConfig() error {
...
+		a.Config = FileStitch
+	case realm.AppConfigVersion20200603:
+		a.Config = FileConfig
+	case realm.AppConfigVersion20210101:
+		a.Config = FileRealmConfig
+	default:
+		return fmt.Errorf("invalid config version: %s", a.AppMeta.ConfigVersion.String())
+	}
+
+	a.LoadConfig()
```
jandersongo, line 297: this loads the original config file, so it should be kept separate from this `LoadAppMeta` helper

internal/local/app_test.go
jandersongo: We should also add tests for local.FindApp to check 
- app with ./mdb/app_meta.json, loads ./mdb/app_meta.json (verify contents)
- app without ./mdb/app_meta.json doesn't load anything
for both cases, we should still load the original config

internal/local/app.go
```diff
@@ -208,6 +214,20 @@ func (a App) WriteConfig() error {
...
 }
 
+// WriteCliConfig writes the .mdb/app_meta.json config to the specified path given a remote app and config version
+func (a App) WriteCliConfig() error {
+	data, err := MarshalJSON(a.AppMeta)
+	if err != nil {
+		return err
+	}
+
+	if err := WriteFile(filepath.Join(a.RootDir, FileMdbAppMeta.String()), 0666, bytes.NewReader(data)); err != nil {
```
jandersongo, line 224: similarly to WriteConfig(), you could just
`return WriteFile(filepath.Join(a.RootDir, FileMdbAppMeta.String()), 0666, bytes.NewReader(data))`
and skip the `if`

internal/local/app.go
```diff
@@ -208,6 +214,20 @@ func (a App) WriteConfig() error {
 	return WriteFile(filepath.Join(a.RootDir, a.Config.String()), 0666, bytes.NewReader(data))
 }
 
+// WriteCliConfig writes the .mdb/app_meta.json config to the specified path given a remote app and config version
+func (a App) WriteCliConfig() error {
+	data, err := MarshalJSON(a.AppMeta)
```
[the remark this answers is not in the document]
NBSquare: `AppMeta` is a struct so it should never be nil. I could add a check to make sure that `config_version` is not empty but as we discussed it may not matter too much and/or be the responsibility of the caller.

internal/commands/push/command_test.go
```diff
@@ -914,6 +914,37 @@ Deployed app is identical to proposed version, nothing to do
 `, out.String())
 	})
 
+	t.Run("with app meta should skip resolve to", func(t *testing.T) {
```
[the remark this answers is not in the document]
NBSquare: I was seeing this naming convention elsewhere in our tests. See the first test in this file for context and let me know what you think.

internal/commands/app/testdata/meta/hosting/metadata.json
```diff
@@ -0,0 +1,10 @@
+[
```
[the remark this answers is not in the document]
NBSquare: I honestly have no idea what this file is but I think it was created by accident. I'll delete for now.

internal/local/app.go
NBSquare: Agreed on removing the call to `LoadConfig` sinc ethat's loading a separate file. But, I do think it's worth keeping the switch statement/setting `app.Config` in this file, since this information is also being "loaded in" from app_meta.

internal/local/app.go
```diff
@@ -251,6 +266,32 @@ func (a *App) LoadConfig() error {
...
+	data, dataErr := ioutil.ReadFile(path)
+	if dataErr != nil {
+		return errFailedToParseAppConfig(path)
+	}
+
+	if err := json.Unmarshal(data, &a.AppMeta); err != nil {
+		return errFailedToParseAppConfig(path)
+	}
+
+	switch a.AppMeta.ConfigVersion {
```
jandersongo, line 281: not sure how this switch statement is beneficial here. 
The config is being set right before calling `LoadConfig`

```
app := App{RootDir: path, Config: config}
if err := app.LoadConfig(); err != nil {
 	return App{}, false, err
}
```

internal/commands/push/command_test.go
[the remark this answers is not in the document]
NBSquare: Oh yeah I see that the test has a `t.Log("and should properly pass through the expected inputs")`. Not sure if that's what is going on here but I agree that this test name makes little-to-no sense so I'm just going to rename it to Jamie's suggestion.

internal/local/app.go
```diff
@@ -208,6 +214,15 @@ func (a App) WriteConfig() error {
 	return WriteFile(filepath.Join(a.RootDir, a.Config.String()), 0666, bytes.NewReader(data))
 }
 
+// WriteMdbAppMeta writes the .mdb/app_meta.json config to the specified path given a remote app and config version
+func (a App) WriteMdbAppMeta() error {
```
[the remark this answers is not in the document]
NBSquare: Yep. I wrote this method when I was working on writing the app meta. The method works and has utility that I _think_ we're going to use down the line so I figured we could leave it here. If this assumption is incorrect or you don't want to merge it here, let me know... I can stash the changes for later.

internal/cloud/realm/import.go
```diff
@@ -58,3 +59,23 @@ func (c *client) doImport(groupID, appID string, appData interface{}, diff bool)
...
+	}
+	if res.StatusCode != http.StatusNoContent {
+		return api.ErrUnexpectedStatusCode{"push", res.StatusCode}
+	}
+	return nil
+}
+
+func (c *client) doPush(groupID, appID string, appMeta interface{}) (*http.Response, error) {
+	return c.doJSON(
+		http.MethodPost,
```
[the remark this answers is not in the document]
makesitgo: also the scope of this ticket doesn't necessitate us creating/including the `/push` API call just yet, let's remove this for now

internal/local/app.go
```diff
@@ -251,6 +266,32 @@ func (a *App) LoadConfig() error {
 	return nil
 }
 
+// LoadAppMeta will load the local app's data from .mdb/app_meta.json
+func (a *App) LoadAppMeta() error {
+	path := filepath.Join(a.RootDir, FileMdbAppMeta.String())
```
makesitgo, line 271: just a [naming/nit] ... let's remove the `Mdb` prefix from the `AppMeta` struct and file (if we're being honest to with go patterns, it probably should have been `MDB` anyways)

instead, you can add some `NameMDB`/`NameDotMDB`/`NameHiddenMDB` to `local/data.go` to help "const-ify" the `.mdb` the `".mdb"` string, which could then be included in this `filepath.Join(...)` sequence

internal/local/app.go
makesitgo: I would actually prefer to see us avoid this here as well... `LoadAppMeta` should do just that

the logic inside `FindApp`, which should attempt to `LoadAppMeta` first before doing our iterative check of known root config files is where we'll be able to manage this stuff

internal/local/app.go
makesitgo: yeah... this is why I'm not the biggest fan of `app_meta.json` as the filename in the first place... it really doesn't do the purpose of this file justice

however, for all intents an purposes, `local.App` is a different representation of an "app" than `realm.App` is, so I think the point stands here that `local.AppMeta` means different things too (than the `realm` counterpart)

and while I don't think we're adding any other files to `.mdb` anytime soon, that's not a reason to couple directory and filename together like that.  we only have `config.json` in the `sync/` directory but still stick with the pattern of differentiating the directories that the CLI understands and the filenames

internal/cloud/realm/import.go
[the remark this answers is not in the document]
makesitgo: how/why did this work get included here?  the JIRA ticket doesn't mention this and that endpoint isn't yet functional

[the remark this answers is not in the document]
NBSquare: On Friday Christine and I discussed this as part of the work we should do here. The `/push` endpoint takes a config version from the AppMeta/config, so because this seemed related to parsing `app_meta.json` it made sense to include this here.

However, it's the *command*'s responsibility to pass params to our REST APIs, and because the command and endpoint aren't live, there wasn't much I could do here beyond stubbing out the endpoint.

internal/local/app.go
```diff
@@ -265,26 +306,33 @@ func FindApp(path string) (App, bool, error) {
 	}
 
 	for i := 0; i < maxDirectoryContainSearchDepth; i++ {
```
jandersongo, line 308: (haven't thoroughly tested but) what do you guys think about something more like this?
```
func resolveConfig(path string) (App, bool, error) {
	app := App{RootDir: path}
	if _, err := os.Stat(filepath.Join(path, FileMdbAppMeta.String())); err == nil {
		app.LoadAppMeta()
		switch app.AppMeta.ConfigVersion {
		case realm.AppConfigVersion20180301:
			app.Config = FileStitch
		case realm.AppConfigVersion20200603:
			app.Config = FileConfig
		case realm.AppConfigVersion20210101:
			app.Config = FileRealmConfig
		}
		return app, true, nil
	}

	for _, config := range allConfigFiles {
		_, err := os.Stat(filepath.Join(path, config.String()))
		if err != nil {
			if os.IsNotExist(err) {
				continue
			}
			return App{}, false, err
		}
		app.Config = config
		return app, true, nil
	}

	return App{}, false, nil
}

// FindApp searches upwards for the root of a Realm app project and
// returns the local app structure, a boolean indicating if the current
// working directory is part of a Realm app project, and any error that occurs
func FindApp(path string) (App, bool, error) {
	wd, wdErr := filepath.Abs(path)
	if wdErr != nil {
		return App{}, false, wdErr
	}

	for i := 0; i < maxDirectoryContainSearchDepth; i++ {
		app, found, err := resolveConfig(wd)
		if err != nil {
			return App{}, false, err
		}
		if !found {
			continue
		}
		if err := app.LoadConfig(); err != nil {
			return App{}, false, err
		}

		// if no config version is found then continue searching for the app's root directory config
		if app.ConfigVersion() == realm.AppConfigVersionZero {
			continue
		}
		return app, true, nil
		if wd == "/" {
			break
		}
		wd = filepath.Clean(filepath.Join(wd, ".."))
	}

	return App{}, false, nil
}
```

internal/cloud/realm/import.go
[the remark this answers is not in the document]
makesitgo: the `/push` endpoint will only be relevant (at least to start) for config versions > `20210101`... which we don't currently support or have a concept of in the CLI

we should be considering this `app_meta.json` file simply as a "cache" of sorts... a way for us to remember the object id versions of app and group ids.  those are id values we need when issuing _any_ admin api request from the cli.

there's going to be a separate ticket for "add support for `20220301`" in the CLI, and its at that point we should address calling the new `/push` and `/pull` endpoints

internal/local/app.go
NBSquare: I like this approach of resolving the config. There's some error handling edge cases I think you're missing but I'll work on this approach.

jandersongo (approved):
LGTM!

christineschen (approved):
looks pretty good! thanks for going through all the back and forth (though still wait for Nick's review before merging)

internal/local/data.go
```diff
@@ -23,15 +23,25 @@ type AppData interface {
...
+// AppMeta represents the .mdb/app_meta.json config
+type AppMeta struct {
+	GroupID       string                 `json:"group_id,omitempty"`
+	AppID         string                 `json:"app_id,omitempty"`
+	ConfigVersion realm.AppConfigVersion `json:"config_version"`
+}
+
 // set of supported local names
 const (
+	// file extensions
```
makesitgo, line 35: [meganit] unnecessary docstrings for non-exported members (feel free to keep as is too)

makesitgo (changes requested):
I'm leaving a few vague comments here so that comes with the additional comment of "feel free to reach out to me if you want to talk about any of this in more detail"... but I think we can consolidate some of this `AppMeta` vs. `ProjectInputs` logic in a few lower-level helpers (I should have totally pointed you to these helpers earlier on)

curious as to what you think about some of these things, and if there is any possibility for what i'm suggesting here!

internal/local/app.go
```diff
@@ -35,6 +35,7 @@ func errFailedToFindApp(path string) error {
 type App struct {
 	RootDir string
 	Config  File
+	AppMeta AppMeta
```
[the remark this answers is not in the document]
NBSquare: The issue with embedding `AppMeta` is that both `AppMeta` and `AppData` have `ConfigVersion` fields which conflict, so I'll rename this to `Meta` rather than rename the meta config version field.

internal/cli/inputs_test.go
```diff
@@ -106,40 +110,75 @@ func TestResolveApp(t *testing.T) {
...
+			description:  "should call find app using AppMeta when present with no input flags and details requested",
+			appMeta:      local.AppMeta{GroupID: "metaGroup", AppID: "metaID", ConfigVersion: realm.DefaultAppConfigVersion},
+			fetchDetails: true,
+			procedure:    func(c *expect.Console) {},
+			expectedApp:  app,
+		},
+		{
+			description:       "should return the single app found from the client call with input flags overriding AppMeta",
+			groupID:           "groupID",
+			appID:             "app",
```
jandersongo, line 176: [nit] "app" --> "appID"

internal/cli/inputs_test.go
```diff
@@ -106,40 +110,75 @@ func TestResolveApp(t *testing.T) {
...
+			expectedAppFilter: realm.AppFilter{GroupID: "groupID"},
 		},
 		{
-			description: "should return an error when no apps are returned from the client call with an app id specified",
-			appID:       "app",
-			procedure:   func(c *expect.Console) {},
-			expectedErr: cli.ErrAppNotFound{"app"},
+			description:       "should return an error when no apps are returned from the client call with an app id specified",
+			groupID:           "groupID",
+			appID:             "app",
```
jandersongo, line 143: [nit]"app" --> "appID"

internal/commands/push/inputs_test.go
```diff
@@ -82,16 +84,33 @@ func TestPushInputsResolve(t *testing.T) {
...
-		assert.Equal(t, appRemote{}, tt)
+		assert.Nil(t, i.Resolve(profile, nil))
+
+		assert.Equal(t, profile.WorkingDirectory, i.LocalPath)
+		assert.Equal(t, "", i.RemoteApp)
 	})
+}
 
+func TestPushInputsResolveTo(t *testing.T) {
 	t.Run("Should return the app id and group id of specified app if to is set to app", func(t *testing.T) {
```
jandersongo, line 114: [nit] can we lowercase the "Should"s

internal/commands/push/inputs_test.go
```diff
@@ -109,10 +128,57 @@ func TestPushInputsResolveTo(t *testing.T) {
...
+			return []realm.App{}, cli.ErrAppNotFound{App: "app"}
+		}
+
+		i := inputs{Project: "groupID", RemoteApp: "appID"}
+		tt, err := i.resolveRemoteApp(nil, client, local.AppMeta{})
+		assert.Nil(t, err)
+		assert.Equal(t, appRemote{GroupID: "groupID"}, tt)
+	})
+
+	t.Run("Should do nothing if find apps cannot find group", func(t *testing.T) {
```
jandersongo, line 173: "should do nothing" --> "should error"

internal/commands/push/inputs_test.go
```diff
@@ -109,10 +128,57 @@ func TestPushInputsResolveTo(t *testing.T) {
...
+
+		var client mock.RealmClient
+		var findAppsCalled bool
+		client.FindAppsFn = func(filter realm.AppFilter) ([]realm.App, error) {
+			findAppsCalled = true
+			return []realm.App{app}, nil
+		}
+
+		var i inputs
+		f, err := i.resolveRemoteApp(nil, client, local.AppMeta{AppID: app.ID, GroupID: app.GroupID, ConfigVersion: realm.AppConfigVersion20210101})
```
jandersongo, line 154: what's `f`? 😛

internal/commands/push/inputs_test.go
```diff
@@ -109,10 +128,57 @@ func TestPushInputsResolveTo(t *testing.T) {
...
 
-		f, err := i.resolveRemoteApp(nil, client)
+		f, err := i.resolveRemoteApp(nil, client, local.AppMeta{})
 		assert.Nil(t, err)
 
 		assert.Equal(t, appRemote{GroupID: app.GroupID, AppID: app.ID, ClientAppID: app.ClientAppID}, f)
 		assert.Equal(t, realm.AppFilter{GroupID: app.GroupID, App: app.ClientAppID}, appFilter)
 	})
+
+	t.Run("Should return the app id and group id from app meta when possible", func(t *testing.T) {
```
jandersongo, line 138: "when possible" --> "when provided"

internal/commands/push/inputs_test.go
```diff
@@ -109,10 +128,57 @@ func TestPushInputsResolveTo(t *testing.T) {
...
+
+		var i inputs
+		f, err := i.resolveRemoteApp(nil, client, local.AppMeta{AppID: app.ID, GroupID: app.GroupID, ConfigVersion: realm.AppConfigVersion20210101})
+		assert.Nil(t, err)
+
+		assert.Equal(t, appRemote{AppID: app.ID, GroupID: app.GroupID}, f)
+		assert.False(t, findAppsCalled, "expected app to skip resolve")
+	})
+
+	t.Run("Should return minimal app remote if find apps cannot find app", func(t *testing.T) {
```
jandersongo, line 161: "if find apps cannot find app" --> "if cannot find app"

makesitgo:
fyi @jandersongo ... I spoke with nathan offline and I'll be stealing his branch here to get this across the finish line.

I'll do my best to incorporate your latest review before I push up to my fork for another round of reviews!

makesitgo:
closing in favor of #292

PR #292, opened 2022-04-22
REALMC-12640: Leverage app meta file when present to query admin api

makesitgo:


internal/cli/inputs.go
```diff
@@ -26,15 +27,16 @@ func (i ProjectInputs) Filter() realm.AppFilter {
...
+	app, _, err := local.FindApp(wd)
+	if err != nil {
+		return err
 	}
 
 	if i.App == "" {
 		var appOption string
 
 		if app.RootDir != "" {
+			i.AppMeta = app.Meta
```
makesitgo, line 39: `// set AppMeta when --app flag is not provided and FindApp locates an app directory`

internal/cli/inputs_test.go
```diff
@@ -49,28 +51,27 @@ func TestProjectAppInputsResolve(t *testing.T) {
 			},
 			test: func(t *testing.T, i cli.ProjectInputs) {
 				assert.Equal(t, "some-app", i.App)
+				assert.Equal(t, local.AppMeta{}, i.AppMeta)
 			},
 		},
 		{
-			description: "when inside a project directory should prompt for app when not flagged and provide client app id as a default",
+			description: "when inside a project directory should not prompt for app when not flagged and use client app id as the option",
```
makesitgo, line 58: these two tests were falsely passing... `tc.procedure` was never being invoked, "defaults" were set from app config files

internal/commands/push/command.go
```diff
@@ -141,11 +141,6 @@ func (cmd *Command) Inputs() cli.InputResolver {
 
 // Handler is the command handler
 func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.Clients) error {
-	app, err := local.LoadApp(cmd.inputs.LocalPath)
```
makesitgo, line 144: ⚠️

not sure I love this, but removed this as an optimization b/c we also called `local.FindApp` in the `inputs.Resolve` phase.  nothing really changes with the command logic itself, but a side effect of this made testing fairly nasty (see: the changes in `push/command_test.go`)

we can add this back or live with the weird `Handler` testing

internal/commands/pull/inputs_test.go
```diff
@@ -116,6 +132,19 @@ func TestPullInputsResolveRemoteApp(t *testing.T) {
...
+		i := inputs{appMeta: local.AppMeta{GroupID: "some-group", AppID: "some-app", ConfigVersion: realm.DefaultAppConfigVersion}}
+		var realmClient mock.RealmClient
+
+		realmClient.FindAppFn = func(groupID, appID string) (realm.App, error) {
+			return realm.App{GroupID: "group-id", ID: "app-id", Name: "some-name"}, nil
+		}
+
+		app, err := i.resolveRemoteApp(nil, cli.Clients{Realm: realmClient})
+		assert.Nil(t, err)
+		assert.Equal(t, realm.App{GroupID: "group-id", ID: "app-id", Name: "some-name"}, app)
```
jandersongo, line 145: [q] I might be missing sth here but shouldn't the `GroupID = "some-group"` (i.e the one provided by the appMeta)?
aren't the app details being set with the app meta here: https://github.com/makesitgo/realm-cli/blob/2e0c742f99998e75f27bf91e55047d1a07550158/internal/cli/inputs.go#L86

internal/commands/push/command.go
jandersongo: 😬 I am also not fond of having to either `LoadApp` or `Resolve` before setting the `cmd` in the tests

internal/commands/pull/inputs_test.go
makesitgo: good callout... there's a funny `TODO` in the `pull_inputs.resolveRemoteApp` function which, after resolving the app, will specifically use `FindApp` to fetch fuller app details (in particular, the "template id").

tweaked this to have the mock "pass through" app data and assert that the `FindApp` function is only called once, rather than twice

internal/commands/push/command.go
makesitgo: convinced me... reverted that optimization

internal/commands/pull/inputs_test.go
```diff
@@ -116,6 +132,22 @@ func TestPullInputsResolveRemoteApp(t *testing.T) {
 		assert.Equal(t, realm.AppFilter{GroupID: "some-project", App: "some-app"}, appFilter)
 	})
 
+	t.Run("should not resolve group id if app meta exists", func(t *testing.T) {
+		i := inputs{appMeta: local.AppMeta{GroupID: "some-group", AppID: "some-app", ConfigVersion: realm.DefaultAppConfigVersion}}
+		var realmClient mock.RealmClient
+
+		var findAppCalls int
+		realmClient.FindAppFn = func(groupID, appID string) (realm.App, error) {
+			findAppCalls++
```
christineschen, line 141: [q] why use an int here instead of a bool?

christineschen (approved):
looks good, we can probably also close Nathan's PR after this is merged

internal/commands/pull/inputs_test.go
makesitgo: it's actually testing that we only call it once rather than twice, so consider this like a "super bool" :)

PR #295, opened 2022-04-22
REALMC-12927: Clean up Resolve for diff/push inputs

jandersongo:


internal/commands/app/diff.go
```diff
@@ -229,7 +240,7 @@ func (i *diffInputs) Resolve(profile *user.Profile, ui terminal.UI) error {
 		i.LocalPath = app.RootDir
 	}
```
makesitgo, line 241: thoughts on adding an `else` condition here to error out if we _still_ don't have a local path which points to an app?  this appears to be an existing bug but it seems like we can still end up in a situation where `i.LocalPath` and `app.RootDir` is empty, which would end up failing the first line of the diff handler when we load the app

if we do go that route, i'd probably flip the if condition (but am commenting here for letting this comment become "Outdated"):

```go
	if app.RootDir == "" {
		return errProjectExistsNot(i.LocalPath) // i'll let you pick a better name
                // it actually looks like we could modify the errInvalidPath you added to account for both
                // a path and project not existing (similar to how errProjectInvalid works in the push package)
	} else {
                i.LocalPath = app.RootDir
        }
```

internal/commands/app/diff_test.go
```diff
@@ -367,4 +374,14 @@ func TestAppDiffInputs(t *testing.T) {
...
 	}
+
+	t.Run("should return an error when specified local path does not exist", func(t *testing.T) {
+		profile, teardown := mock.NewProfileFromTmpDir(t, "app_init_input_test")
+		defer teardown()
+
+		localPath := "fakePath"
+
+		i := diffInputs{LocalPath: localPath}
+		assert.Equal(t, errInvalidPath(localPath), i.Resolve(profile, nil))
```
makesitgo, line 385: if we do go with the extra returned error I suggest above, let's make sure to add a similar test case for that scenario

makesitgo (commented):
just one thing from me before I think this is ready to be merged

internal/commands/app/diff_test.go
```diff
@@ -283,18 +290,6 @@ func TestAppDiffInputs(t *testing.T) {
 		procedure      func(c *expect.Console)
 		test           func(t *testing.T, i diffInputs, p *user.Profile)
 	}{
-		{
-			description:    "should resolve empty inputs when outside an app directory by prompting for the local path which does not exist",
```
jandersongo, line 287: testing non-existent local path input is handled in new tests

makesitgo (approved):
🆒  beans

PR #297, opened 2022-04-27
REALMC-12113: Add DefaultRule to exported app structure

Amaan95:
## Changes
- Add the Default Rule to the exported app structure for App Structure V1 and V2
- Write and parse the default rule to/from the Data Sources directory for App Structure V2
- Write and parse the default rule to/from the Services directory for App Structure V1

### DIFF
--- a/internal/local/app_test.go
+++ b/internal/local/app_test.go
@@ -720,7 +720,6 @@ exports = function({ query }) {
 				{
 					"database":   "foo",
 					"collection": "bar",
-					"name":       "foo.bar",
 				},
 			},
 		},
--- a/internal/local/app_test.go
+++ b/internal/local/app_test.go
@@ -698,6 +698,32 @@ exports = function({ query }) {
 				},
 			},
 		},
+		{
+			Config: map[string]interface{}{
+				"type":    "mongodb",
+				"name":    "mdbSvc",
+				"config":  map[string]interface{}{},
+				"version": float64(1),
+			},
+			DefaultRule: map[string]interface{}{
+				"roles": []interface{}{
+					map[string]interface{}{
+						"name": "owner",
+						"apply_when": map[string]interface{}{
+							"userId": "%%user.id",
+						},
+						"read": true,
+					},
+				},
+			},
+			Rules: []map[string]interface{}{
+				{
+					"database":   "foo",
+					"collection": "bar",
+					"name":       "foo.bar",
+				},
+			},
+		},
 	},
 	Triggers: []map[string]interface{}{
 		{
--- a/internal/local/structure.go
+++ b/internal/local/structure.go
@@ -26,6 +26,7 @@ type SecretsStructure struct {
 // ServiceStructure represents the Realm app service structure
 type ServiceStructure struct {
 	Config           map[string]interface{}   `json:"config,omitempty"`
+	DefaultRule      map[string]interface{}   `json:"default_rule,omitempty"`
 	IncomingWebhooks []map[string]interface{} `json:"incoming_webhooks"`
 	Rules            []map[string]interface{} `json:"rules"`
 }
@@ -207,18 +208,28 @@ func parseServices(rootDir string) ([]ServiceStructure, error) {
 	if walkErr := dw.walk(func(file os.FileInfo, path string) error {
 		var svc ServiceStructure
 
+		// Config
 		config, err := parseJSON(filepath.Join(path, FileConfig.String()))
 		if err != nil {
 			return err
 		}
 		svc.Config = config
 
+		// Default Rule
+		defaultRule, err := parseJSON(filepath.Join(path, FileDefaultRule.String()))
+		if err != nil {
+			return err
+		}
+		svc.DefaultRule = defaultRule
+
+		// Webhooks
 		webhooks, err := parseFunctions(filepath.Join(path, NameIncomingWebhooks))
 		if err != nil {
 			return err
 		}
 		svc.IncomingWebhooks = webhooks
 
+		// Rules
 		rules, err := parseJSONFiles(filepath.Join(path, NameRules))
 		if err != nil {
 			return err
@@ -329,6 +340,8 @@ func writeServices(rootDir string, services []ServiceStructure) error {
 			return errors.New("error writing services")
 		}
 		dirSvc := filepath.Join(dir, nameSvc)
+
+		// Config
 		data, err := MarshalJSON(svc.Config)
 		if err != nil {
 			return err
@@ -340,6 +353,21 @@ func writeServices(rootDir string, services []ServiceStructure) error {
 		); err != nil {
 			return err
 		}
+
+		// Default Rule
+		defaultRule, err := MarshalJSON(svc.DefaultRule)
+		if err != nil {
+			return err
+		}
+		if err := WriteFile(
+			filepath.Join(dirSvc, FileDefaultRule.String()),
+			0666,
+			bytes.NewReader(defaultRule),
+		); err != nil {
+			return err
+		}
+
+		// Webhooks
 		for _, webhook := range svc.IncomingWebhooks {
 			src, ok := webhook[NameSource].(string)
 			if !ok {
@@ -373,6 +401,8 @@ func writeServices(rootDir string, services []ServiceStructure) error {
 				return err
 			}
 		}
+
+		// Rules
 		for _, rule := range svc.Rules {
 			data, err := MarshalJSON(rule)
 			if err != nil {
--- a/internal/local/structure.go
+++ b/internal/local/structure.go
@@ -355,16 +355,18 @@ func writeServices(rootDir string, services []ServiceStructure) error {
 		}
 
 		// Default Rule
-		defaultRule, err := MarshalJSON(svc.DefaultRule)
-		if err != nil {
-			return err
-		}
-		if err := WriteFile(
-			filepath.Join(dirSvc, FileDefaultRule.String()),
-			0666,
-			bytes.NewReader(defaultRule),
-		); err != nil {
-			return err
+		if svc.DefaultRule != nil {
+			defaultRule, err := MarshalJSON(svc.DefaultRule)
+			if err != nil {
+				return err
+			}
+			if err := WriteFile(
+				filepath.Join(dirSvc, FileDefaultRule.String()),
+				0666,
+				bytes.NewReader(defaultRule),
+			); err != nil {
+				return err
+			}
 		}
 
 		// Webhooks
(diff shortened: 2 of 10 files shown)


internal/local/app_test.go
```diff
@@ -698,6 +698,32 @@ exports = function({ query }) {
...
+						},
+						"read": true,
+					},
+				},
+			},
+			Rules: []map[string]interface{}{
+				{
+					"database":   "foo",
+					"collection": "bar",
+					"name":       "foo.bar",
```
malexandert, line 723: [q] do we need to specify the `"name"` field here. Technically rule names should only be internal - externally we specify the `database` and `collection` which will eventually get concatenated into a `name`

internal/local/structure_test.go
```diff
@@ -336,6 +336,73 @@ exports = function({ query }) {
...
+						"apply_when": map[string]interface{}{
+							"userId": "%%user.id",
+						},
+						"read": true,
+					},
+				},
+			},
+			Rules: []map[string]interface{}{
+				{
+					"name":       "foo.bar",
```
malexandert, line 364: same comment about `name` here

internal/local/structure_v2_test.go
```diff
@@ -44,15 +44,27 @@ func TestParseDataSources(t *testing.T) {
 
 	testRoot := filepath.Join(wd, "testdata/data_sources")
 
+	mdbAtlasStr := "mongodb-atlas"
 	t.Run("should return the parsed data sources directory with nested rules and schema", func(t *testing.T) {
 		dataSources, err := parseDataSources(testRoot)
 		assert.Nil(t, err)
-		assert.Equal(t, []DataSourceStructure{{
+		assert.Match(t, []DataSourceStructure{{
```
malexandert, line 51: [q] why the switch from `Equal` to `Match`?

internal/local/structure_v2_test.go
```diff
@@ -44,15 +44,27 @@ func TestParseDataSources(t *testing.T) {
...
 			Config: map[string]interface{}{
-				"type":   "mongodb-atlas",
-				"name":   "mongodb-atlas",
+				"type":   mdbAtlasStr,
```
malexandert, line 53: nice 💯

internal/local/testdata/full_project/services/mdbSvc/rules/foo.bar.json
```diff
@@ -0,0 +1,5 @@
+{
+  "database": "foo",
+  "collection": "bar",
+  "name": "foo.bar"
```
malexandert, line 4: same comment about `name` here

malexandert (approved):
Mainly questions from me, but this looks good! I'd definitely wait till Jamie or Nick have a look before merging though 👍🏾

internal/local/app_test.go
Amaan95: That's what I thought and it's what we do for data sources in v2 of the app structure with v2 rules, but for legacy services with rules and mongo services in the v1 structure it does appear that we use the `"name"` field here: 
https://github.com/10gen/realm-cli/blob/3795c0740d25d61bed4de823807e7351fcd06bbf/internal/local/structure.go#L382

this seems to be intentional but I'll defer to someone with historical context (perhaps @makesitgo) on whether this is how things should be handled for v2 rules in the v1 app structure. if not, it would be a fairly simple change to construct the name for a v2 rule using the database and collection names when writing the rule file.

internal/local/structure_test.go
Amaan95: addressed above

internal/local/testdata/full_project/services/mdbSvc/rules/foo.bar.json
Amaan95: addressed above

internal/local/structure_v2_test.go
Amaan95: I had encountered a type match error earlier when testing but that seems to be resolved, will change it back!

internal/local/app_test.go
makesitgo: we actually don't seem to ever export `"name"` as a field in any exported rule file for any config version.  see:

* `20200603` and below: https://github.com/10gen/baas/blob/master/importexport/mongodbrules/v2/rule.go#L232-L264
* `20210101` and above: https://github.com/10gen/baas/blob/master/importexport/mongodbrules/v2/rule.go#L312-L328

I think what you're seeing in `local/structure` for the `writeServices` codepath is actually a bug, and we should instead being computing "name" via "database"+"collection" fields.

fortunately (and unfortunately), that codepath is actually never hit in the CLI... I don't love that we needed to go down the route of teaching the CLI how to write _any_ app config files b/c for 98% of the scenarios, the "writing" occurs just by reading the zip response returned from the API.  for context, we went this route so that `app init`/`app create` could link clusters/data sources

internal/local/structure_test.go
makesitgo, line 364: would bump the need to remove this (and other instances of it) so we more accurately reflect how the server exports these payloads

internal/local/structure_v2.go
```diff
@@ -44,8 +44,9 @@ type AuthStructure struct {
 
 // DataSourceStructure represents the v2 Realm app data source structure
 type DataSourceStructure struct {
-	Config map[string]interface{}   `json:"config,omitempty"`
-	Rules  []map[string]interface{} `json:"rules,omitempty"`
+	Config      map[string]interface{}   `json:"config,omitempty"`
+	DefaultRule map[string]interface{}   `json:"default_rule,omitempty"`
```
makesitgo, line 48: [question] I'm guessing it's currently a moot point, but perhaps we want to add `DefaultRule` to `HTTPServiceStructure` and its corresponding parse/write methods as well?

basically, in switch from structure v1 => v2, we split "services" into "services+http services+data sources"

if "default rule" is a concept on our core `ServiceDesc` interface in the backend, we may just want to future-proof ourselves and match that design by including it here as well.  it seems that `master` on `baas` is already equipped to handle the same based on your recent changes (b/c data sources and http services all just embed the core `Service` type)

internal/local/structure_v2_test.go
```diff
@@ -44,15 +44,27 @@ func TestParseDataSources(t *testing.T) {
 
 	testRoot := filepath.Join(wd, "testdata/data_sources")
 
+	mdbAtlasStr := "mongodb-atlas"
```
makesitgo, line 47: [style/meganit] :meh: ... doubtful this really buys all that much at the expense of one extra hop in `git blame`/`git history`

makesitgo (commented):
I wouldn't mind your help with a drive-by bug fix on using `"name"` here plus a few other minor things

internal/local/app_test.go
Amaan95: Thanks for the context! I'll make the change 👍🏽

internal/local/structure_v2_test.go
Amaan95: Fair enough 😅

internal/local/structure_v2.go
Amaan95: Yeah, I don't think there's any value in doing this at the moment given that for all non-mongo services we return a nil value for the default rule and I don't think that's liable to change. 

It's unfortunate that the DefaultRule is a concept on the core `ServiceDesc` but that's a tradeoff we had to make for ease of implementation. Once 3rd party services are deprecated in the future, we may be able to revisit this and have separate implementations for Mongo services and have the DefaultRule be exclusive to that interface.

internal/local/structure.go
```diff
@@ -373,13 +401,25 @@ func writeServices(rootDir string, services []ServiceStructure) error {
...
 		}
+
+		// Rules
 		for _, rule := range svc.Rules {
 			data, err := MarshalJSON(rule)
 			if err != nil {
 				return err
 			}
+
+			// V1 Rules have an exported "name" field
```
makesitgo, line 412: is this actually true?  I'm doubting myself now but I thought I saw in `importexport` that we never exported the "name" field...

makesitgo (commented):
I think we should 100% verify if v1 rules (which is a misleading way to refer to them as, since I think you're saying "v1" in an app structure sense but we also commonly refer to these rules as "v2" rules from the mongodb sense) actually export the name field

internal/local/structure.go
```diff
@@ -340,6 +353,21 @@ func writeServices(rootDir string, services []ServiceStructure) error {
 		); err != nil {
 			return err
 		}
+
+		// Default Rule
+		defaultRule, err := MarshalJSON(svc.DefaultRule)
```
makesitgo, line 358: what gets written here if `DefaultRule` is nil or empty?  do we want the file to exist but be empty/`{}`?

I would have guessed maybe we want a `if len(svc.DefaultRule) != 0` check wrapping this file write

internal/local/structure.go
```diff
@@ -373,13 +401,25 @@ func writeServices(rootDir string, services []ServiceStructure) error {
...
 			data, err := MarshalJSON(rule)
 			if err != nil {
 				return err
 			}
+
+			// Mongo service rules (AKA NamespaceRules) do not have an exported "name" field and should construct the rule name
+			// using the database and collection values
+			ruleName := fmt.Sprintf("%s.%s", rule["database"], rule["collection"])
+			_, isMongoSvcRule := rule["database"]
+			if !isMongoSvcRule {
```
makesitgo, line 416: [nit] always good to keep a tight scope
```suggestion
			if _, isMongoSvcRule := rule["database"]; !isMongoSvcRule {
```

internal/local/structure.go
```diff
@@ -373,13 +401,25 @@ func writeServices(rootDir string, services []ServiceStructure) error {
...
+				// Rules with a type of BuiltinRule have an exported "name" field. These are rules used for non-Mongo services
+				ruleName = rule["name"].(string)
```
makesitgo, line 418: we should also always avoid the panic if possible
```suggestion
				if name, ok := rule["name"].(string); ok {
				        ruleName = name
				}
```

makesitgo (approved):
last few things from me... approving since they're all related to the non-critical "write services" portion of this codebase

internal/local/structure.go
Amaan95: Good catch! Added this check as well as a test case to ensure this is handled properly

Amaan95:
CLI tests appear to be failing on master. Failing tests are unrelated to the changes in this PR. Will go ahead and merge.

Failing tests ticket: https://jira.mongodb.org/browse/REALMC-13220

PR #298, opened 2022-04-28
REALMC-11241: add instructions for template apps

jandersongo:
as per designs: https://www.figma.com/file/dVxHOsEQXF8Jsdth2IKpa3/EXPO-1508-CLI-TEMPLATES?node-id=0%3A1

```
jamie.anderson@M-C02F15DGMD6M mongodb % ./realm-cli-next app create  --template web.mql.todo --profile dev_rcymllrh                 
? App Name appName
? Atlas Project Project0 - 60ad6d0cb54e70155ff5ac5a
㊂ Creating template app... Created template app
Successfully created app
{
  "client_app_id": "appname-cwtxq",
  "filepath": "/Users/jamie.anderson/mongodb/appName",
  "url": "https://realm-dev.mongodb.com/groups/60ad6d0cb54e70155ff5ac5a/apps/627883abab706b467656763b/dashboard",
  "backend": "/Users/jamie.anderson/mongodb/appName/backend",
  "clusters": [
    {
      "name": "mongodb-atlas"
    }
  ]
}
Check out your app: cd ./appName && realm-cli app describe
View directions on how to run the template app: appName/frontend/web.mql.todo/README.rst
```
```
jamie.anderson@M-C02F15DGMD6M mongodb % ./realm-cli-next pull --remote appname-cwtxq --template web.mql.todo
? Atlas Project Project0 - 60ad6d0cb54e70155ff5ac5a
Saved app to disk
Successfully pulled app down: appName
Successfully saved template(s) to disk
  web.mql.todo: appName/frontend/web.mql.todo/README.rst
Navigate to the saved directory to view directions on how to run the template app(s)
```


internal/commands/app/create.go
```diff
@@ -21,7 +21,7 @@ import (
 var CommandMetaCreate = cli.CommandMeta{
 	Use:         "create",
 	Display:     "app create",
-	Description: "Create a new app from your current working directory and deploy it to the Realm server",
+	Description: "Create a new app or a template app from your current working directory and deploy it to the Realm server",
```
makesitgo, line 24: [question] did this copy change come down as a requirement from design?  I know @kpham1231 was originally involved with crafting these messages and perhaps ighodalo has thoughts on what to put here

my gut?  probably would have expected `"Create a new app (or template app) from..."`

makesitgo: cc: @Ighodaloijagbone@yahoo.com

internal/commands/app/create.go
```diff
@@ -481,6 +481,11 @@ func (cmd CommandCreate) handleCreateTemplateApp(
 	// TODO(REALMC-9460): Add better template-app-specific directions for checking out the newly created template app
 	ui.Print(terminal.NewJSONLog("Successfully created app", output))
 	ui.Print(terminal.NewFollowupLog("Check out your app", fmt.Sprintf("cd ./%s && %s app describe", cmd.inputs.LocalPath, cli.Name)))
+
+	if matches, err := filepath.Glob(filepath.Join(frontendDir, cmd.inputs.Template, "README.*")); err == nil && len(matches) == 1 {
+		ui.Print(terminal.NewFollowupLog(fmt.Sprintf("View directions on how to run the template app: %s", matches[0])))
```
makesitgo, line 486: also related [question] ... seems like "template" snuck into this line compared to the figma... just want to confirm this is desirable (with design)

internal/commands/pull/command.go
```diff
@@ -258,17 +258,22 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 		ui.Print(terminal.NewDebugLog("Fetched hosting assets"))
 	}
 
-	successfulTemplateWrites := make([]string, 0, len(clientTemplates))
```
makesitgo, line 261: [question] i'm wondering if there was some magic previously that allowed this render properly, or was it just simply always rendering the list of templates horizontally and all within a single set of `[...]` brackets?

related to that, i'm surprised we do not have any tests for `pull/command_test.go` which verifies the output in the scenario where a template was written, can we add those?

internal/commands/pull/command.go
```diff
@@ -258,17 +258,22 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+	successfulTemplateWrites := make([]interface{}, 0, len(clientTemplates))
```
makesitgo, line 261: rather than manage the `"  "` yourself in the `"  Navigate..."` message, you could initialize this to be:
```suggestion
	successfulTemplateWrites := make([]interface{}, 0, len(clientTemplates)+1)
```
and then simply append the "Navigate string" as one last item to get the same effect here

internal/commands/pull/command.go
```diff
@@ -258,17 +258,22 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
 	for _, ct := range clientTemplates {
 		if err := local.WriteZip(filepath.Join(pathFrontend, ct.id), ct.zipPkg); err != nil {
 			return fmt.Errorf("unable to save template '%s' to disk: %s", ct.id, err)
 		}
-		successfulTemplateWrites = append(successfulTemplateWrites, ct.id)
+		readmeFile := ""
```
makesitgo, line 266: [style/meganit/optional] in the theme of "make the zero value meaningful", I tend to prefer declarations like this as `var readmeFile string` instead of assigning `:= ""` to the literal zero value

internal/commands/pull/command.go
```diff
@@ -258,17 +258,22 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+		if matches, err := filepath.Glob(filepath.Join(pathRelative, local.FrontendPath, ct.id, "README.*")); err == nil && len(matches) == 1 {
```
makesitgo, line 267: I hate that I'm suggesting this, but this "readme matcher" _could_ probably live as a helper within the `local` package (where `"README"` could then be its own unexported `const` there) and also consolidate the `err == nil && len == 1` check...

we do a similar thing in `local/dependencies` with:
`filepath.Glob(filepath.Join(rootDir, nameNodeModules+"*"))`

internal/commands/pull/command.go
```diff
@@ -258,17 +258,22 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+			readmeFile = matches[0]
+		}
+		successfulTemplateWrites = append(successfulTemplateWrites, fmt.Sprintf("[%s] %s", ct.id, readmeFile))
 	}
 
 	ui.Print(terminal.NewTextLog("Successfully pulled app down: %s", pathRelative))
 	if len(successfulTemplateWrites) != 0 {
-		ui.Print(terminal.NewListLog("Successfully saved template(s) to disk", successfulTemplateWrites))
+		ui.Print(terminal.NewListLog("Successfully saved template(s) to disk", successfulTemplateWrites...))
+		ui.Print(terminal.NewTextLog("  Navigate to the saved directory to view directions on how to run the template app(s)"))
```
makesitgo, line 276: and as a contradiction to what I mentioned earlier with initializing the list with a capacity+1, do we actually want this "Navigate" message to be indented along with the list of templates saved?  from the looks of it, even with just one template saved it sort of gets buried in the list making it hard to see

---

also just calling out the "template" prefix to `app` here so we can remain consistent with the naming style, however we land on it

internal/commands/pull/command.go
```diff
@@ -258,17 +258,22 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
 	for _, ct := range clientTemplates {
 		if err := local.WriteZip(filepath.Join(pathFrontend, ct.id), ct.zipPkg); err != nil {
 			return fmt.Errorf("unable to save template '%s' to disk: %s", ct.id, err)
 		}
-		successfulTemplateWrites = append(successfulTemplateWrites, ct.id)
+		readmeFile := ""
+		if matches, err := filepath.Glob(filepath.Join(pathRelative, local.FrontendPath, ct.id, "README.*")); err == nil && len(matches) == 1 {
+			readmeFile = matches[0]
+		}
+		successfulTemplateWrites = append(successfulTemplateWrites, fmt.Sprintf("[%s] %s", ct.id, readmeFile))
```
makesitgo, line 270: my guess is the original `[%s]` style formatting was a mistake (of using `[]string` instead of spreading `[]interface{}` as the last arg to `NewListLog`.  do we still want to preserve that?

seems like something a little less distracting could just be:

`fmt.Sprintf("%s: %s", ct.id, readmeFile)`

again, this is just another thing we should run by design... it may help to schedule a quick 15 min screen share with them so they can really see the output in action

makesitgo (commented):
various thoughts and musings here... I sorta went through it quick so apologies if any comments are rambling and meandering 😬

internal/commands/pull/command.go
jandersongo: discussed offline: "un-identing" the "Naviagate" message was approved by design

internal/commands/app/create.go
jandersongo: discussed offline: adding brackets is approved by design

internal/commands/pull/command_test.go
```diff
@@ -626,6 +628,39 @@ Successfully pulled app down: app
...
+				if templateID == xamarinTemplate {
+					return &templateWithReadmeZipPkg.Reader, true, nil
+				}
+				return nil, false, nil
+			}
+
+			out := new(bytes.Buffer)
+			ui := mock.NewUIWithOptions(mock.UIOptions{AutoConfirm: true}, out)
+
+			profile := mock.NewProfileFromWd(t)
```
makesitgo, line 650: this test is pulling data/writing to disk, right?  should we be using `NewProfileFromTmpDir` instead?

is that why there's new files for `testdata/template_app/[backend|frontend]/...`?

makesitgo (commented):
small nitty things and one question about a test setup... should be good to go afterwards though!

internal/local/app.go
```diff
@@ -363,3 +364,12 @@ func loadAppRoot(path string) (App, bool, error) {
...
 	return app, true, nil
 }
+
+// FindReadMe finds the Realm app README file for the specified template
+func FindReadMe(path, templateID string) (string, error) {
+	matches, err := filepath.Glob(filepath.Join(path, FrontendPath, templateID, nameREADME+"*"))
+	if err != nil || len(matches) != 1 {
+		return "", err
+	}
+	return matches[0], nil
```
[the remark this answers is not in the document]
jandersongo: done! (& updated sample output)

internal/commands/pull/command_test.go
jandersongo: removed new template_app files in favour of using `NewProfileFromTmpDir`

makesitgo (approved):
sad we lost the `snowflake` README ☹️ 

but looks good to me!

🆒  beans

PR #299, opened 2022-05-04
REALMC-13094: fetch all available groups from Atlas Public API

jandersongo:
In terms of testing, it looks like we only have access to 1 group. Is it possible to add more groups? at least 1 so we can pass in a `itemsPerPage = 1` query param to force the fetch for more groups: https://github.com/10gen/realm-cli/blob/3795c0740d25d61bed4de823807e7351fcd06bbf/internal/utils/test/test.go#L39

ex
for testing manually, I passed `api.RequestOptions{Query: map[string]string{"itemsPerPage":"1"}}` for profiles in which I had multiple groups and 'printed' the 'next link' to the terminal

```
jamie.anderson@M-C02F15DGMD6M mongodb % ./realm-cli-next --profile dev pull
NEXT LINK: &{https://cloud-dev.mongodb.com/api/public/v1.0/groups?itemsPerPage=1&pageNum=2 next}
NEXT LINK: &{https://cloud-dev.mongodb.com/api/public/v1.0/groups?itemsPerPage=1&pageNum=3 next}
? Atlas Project  [Use arrows to move, type to filter]
> Project - 621e79993ab79a6b79c2b6d9
  Project0 - 60ad6d0cb54e70155ff5ac5a
  Project1 - 60b125e4090b4831f3287555
```

### DIFF
--- a/internal/cloud/atlas/client.go
+++ b/internal/cloud/atlas/client.go
@@ -22,7 +22,7 @@ const (
 
 // Client is a MongoDB Cloud Atlas client
 type Client interface {
-	Groups() ([]Group, error)
+	Groups(url string, useBaseURL bool) (Groups, error)
 
 	Clusters(groupID string) ([]Cluster, error)
 	ServerlessInstances(groupID string) ([]ServerlessInstance, error)
@@ -49,8 +49,12 @@ type client struct {
 	transport *digest.Transport
 }
 
-func (c *client) do(method, path string, options api.RequestOptions) (*http.Response, error) {
-	req, reqErr := http.NewRequest(method, c.baseURL+path, options.Body)
+func (c *client) doWithBaseURL(method, path string, options api.RequestOptions) (*http.Response, error) {
+	return c.doWithURL(method, c.baseURL+path, options)
+}
+
+func (c *client) doWithURL(method, url string, options api.RequestOptions) (*http.Response, error) {
+	req, reqErr := http.NewRequest(method, url, options.Body)
 	if reqErr != nil {
 		return nil, reqErr
 	}
--- a/internal/cloud/atlas/groups.go
+++ b/internal/cloud/atlas/groups.go
@@ -9,6 +9,8 @@ import (
 
 const (
 	groupsPath = publicAPI + "/groups"
+
+	linkRelNext = "next"
 )
 
 // Group is an Atlas group
@@ -17,27 +19,67 @@ type Group struct {
 	Name string `json:"name"`
 }
 
-type groupResponse struct {
-	Results []Group `json:"results"`
+// Link contains additional data related to the response
+type Link struct {
+	Href string `json:"href"`
+	Rel  string `json:"rel"`
 }
 
-func (c *client) Groups() ([]Group, error) {
-	res, resErr := c.do(
+// Groups represents the groups response from Atlas
+type Groups struct {
+	Results []Group `json:"results"`
+	Links   []Link  `json:"links"`
+}
+
+func (c *client) Groups(url string, useBaseURL bool) (Groups, error) {
+	if useBaseURL {
+		url = c.baseURL + url
+	}
+	res, err := c.doWithURL(
 		http.MethodGet,
-		groupsPath,
+		url,
 		api.RequestOptions{},
 	)
-	if resErr != nil {
-		return nil, resErr
+	if err != nil {
+		return Groups{}, err
 	}
 	if res.StatusCode != http.StatusOK {
-		return nil, api.ErrUnexpectedStatusCode{"get groups", res.StatusCode}
+		return Groups{}, api.ErrUnexpectedStatusCode{"get groups", res.StatusCode}
 	}
 	defer res.Body.Close()
 
-	var groupRes groupResponse
-	if err := json.NewDecoder(res.Body).Decode(&groupRes); err != nil {
+	var groups Groups
+	if err := json.NewDecoder(res.Body).Decode(&groups); err != nil {
+		return Groups{}, err
+	}
+	return groups, nil
+}
+
+// AllGroups fetches all atlas groups
+func AllGroups(c Client) ([]Group, error) {
+	groups, err := c.Groups(groupsPath, true)
+	if err != nil {
 		return nil, err
 	}
-	return groupRes.Results, nil
+	return fetchNextGroups(c, groups)
+}
+
+func fetchNextGroups(c Client, groups Groups) ([]Group, error) {
+	allGroups := groups.Results
+	for _, link := range groups.Links {
+		if link.Rel != linkRelNext {
+			continue
+		}
+		res, err := c.Groups(link.Href, false)
+		if err != nil {
+			return nil, err
+		}
+		nextGroups, err := fetchNextGroups(c, res)
+		if err != nil {
+			return nil, err
+		}
+		allGroups = append(allGroups, nextGroups...)
+		break
+	}
+	return allGroups, nil
 }
--- a/internal/cloud/atlas/groups_test.go
+++ b/internal/cloud/atlas/groups_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/10gen/realm-cli/internal/cloud/atlas"
 	u "github.com/10gen/realm-cli/internal/utils/test"
 	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/10gen/realm-cli/internal/utils/test/mock"
 )
 
 func TestAtlasGroups(t *testing.T) {
@@ -29,7 +30,7 @@ func TestAtlasGroups(t *testing.T) {
 		},
 	} {
 		t.Run(tc.description, func(t *testing.T) {
-			_, err := tc.client.Groups()
+			_, err := atlas.AllGroups(tc.client)
 			assert.Equal(t, tc.expectedErr, err)
 		})
 	}
@@ -37,7 +38,7 @@ func TestAtlasGroups(t *testing.T) {
 	t.Run("With an authenticated client should return the list of groups", func(t *testing.T) {
 		client := newAuthClient(t)
 
-		groups, err := client.Groups()
+		groups, err := atlas.AllGroups(client)
 		assert.Nil(t, err)
 		assert.Equal(t, u.CloudGroupCount(), len(groups))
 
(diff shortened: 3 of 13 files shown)


internal/cloud/atlas/groups.go
```diff
@@ -17,27 +17,58 @@ type Group struct {
...
-	if resErr != nil {
-		return nil, resErr
+	groupRes, err := c.fetchGroups(groupsPath, false)
+	if err != nil {
+		return nil, err
+	}
+
+	allGroups := groupRes.Results
+	next := groupRes.nextLink()
+	for next != nil {
```
makesitgo, line 38: we may need to tweak this logic a bit, as _each_ groups response has the potential for a `"next"` link in it, not just that the initial response would have one

I think we can get a little sleek with it here:

```go
func (c *client) Groups() ([]Group, error) {
	var fetcher func(url string) ([]Group, error)
	fetcher = func(url string) ([]Group, error) {
		res, resErr := c.doURL(
			http.MethodGet,
			groupsPath,
			api.RequestOptions{},
		)
		if resErr != nil {
			return nil, resErr
		}
			return nil, api.ErrUnexpectedStatusCode{"get groups", res.StatusCode}
		}
		defer res.Body.Close()

		var groupRes groupResponse
		if err := json.NewDecoder(res.Body).Decode(&groupRes); err != nil {
			return nil, err
		}

		allGroups := groupRes.Results
		for _, link := range groupRes.Links {
			if link.Rel == "next" {
				nextGroups, err := fetcher(link.Href)
				if err != nil {
					return nil, err
				}
				allGroups = append(allGroups, nextGroups...)
			}
		}
		return allGroups, nil
	}

	return fetcher(groupsPath)
}```

the main thing I'd tweak above there is it'd be nice if we could order the returned array in the same way we got the results (whereas I think how it is currently, we end up "reverse ordering" the pages)

Comment (jandersongo) [reply internal/cloud/atlas/groups.go:38]: My understanding was that the group response (first & only group response) will contain the "links" field that is essentially a chain linking ****all** of the other groups** with the "next" field. 
https://www.mongodb.com/docs/atlas/api/atlas-admin-api/#std-label-atlas-api-linking
image
Does that sound right? or are you fairly certain that each group response could have its own chain link of projects?

Comment (jandersongo) [reply internal/cloud/atlas/groups.go:38]: I like this! 
and in regards to the ordering, it looks like it is being kept in the proper order (since we are appending the new groups to the first). 
based off a walk through this is how the allGroups is being filled:
```
get group 1
get group 2 
get group 3
append group 3 to group 2
append groups [2,3] to group 1
```
allGroups is [1,2,3] ✅

File: internal/cloud/atlas/groups.go [hunk 2]
```diff
@@ -17,15 +17,25 @@ type Group struct {
...
+}
+
 type groupResponse struct {
 	Results []Group `json:"results"`
+	Links   []*link `json:"links"`
 }
 
 func (c *client) Groups() ([]Group, error) {
-	res, resErr := c.do(
+	return c.FetchGroups(c.baseURL+GroupsPath, api.RequestOptions{})
```
jandersongo, line 31: (for testing purposes) broke this out into its own func instead of keeping as a nested func

internal/cloud/atlas/client.go
```diff
@@ -49,8 +50,16 @@ type client struct {
 	transport *digest.Transport
 }
 
-func (c *client) do(method, path string, options api.RequestOptions) (*http.Response, error) {
-	req, reqErr := http.NewRequest(method, c.baseURL+path, options.Body)
+func (c *client) doURL(method, path string, options api.RequestOptions) (*http.Response, error) {
```
makesitgo, line 53: I'm not sure this extra private method needs to exist if its solely passing through to `c.do`...

internal/cloud/atlas/client.go
```diff
@@ -49,8 +50,16 @@ type client struct {
...
+	return c.do(method, path, options)
+}
+
+func (c *client) doBaseURL(method, path string, options api.RequestOptions) (*http.Response, error) {
```
makesitgo, line 57: ...instead you could keep `c.do` as is and then make this `c.doWithBaseURL` (although the diff of this will be unfortunately larger than probably need be, if we were ok with less-than-optimal names)

open to ideas here, i'd like to:
a) keep the names obvious
b) keep the diff intentional

but there's going to be some amount of compromise you need with this, somewhere

internal/cloud/atlas/groups.go
```diff
@@ -7,8 +7,9 @@ import (
 	"github.com/10gen/realm-cli/internal/utils/api"
 )
 
+// GroupsPath is the route for group API handlers
 const (
-	groupsPath = publicAPI + "/groups"
+	GroupsPath = publicAPI + "/groups"
```
makesitgo, line 12: would keep this unexported if its otherwise solely for testing purposes... the test can just use a raw string

internal/cloud/atlas/groups.go
```diff
@@ -17,15 +18,25 @@ type Group struct {
...
 }
 
+type link struct {
+	Href string `json:"href"`
+	Rel  string `json:"rel"`
+}
+
 type groupResponse struct {
 	Results []Group `json:"results"`
+	Links   []*link `json:"links"`
```
makesitgo, line 28: I don't think this needs to be an array of _pointers_

internal/cloud/atlas/groups_test.go
```diff
@@ -49,6 +51,28 @@ func TestAtlasGroups(t *testing.T) {
...
+			numGroups int
+		}{
+			{numGroups: 1},
+			{numGroups: 3},
+			{numGroups: 5},
+		} {
+			t.Run(fmt.Sprintf("when num items allowed per page is %d", tc.numGroups), func(t *testing.T) {
+				client := newAuthClient(t)
+				options := api.RequestOptions{Query: map[string]string{"itemsPerPage": fmt.Sprint(tc.numGroups)}}
+				groups, err := client.FetchGroups(u.AtlasServerURL()+atlas.GroupsPath, options)
```
makesitgo, line 68: why does this need to use `FetchGroups` instead of just `Groups`?

makesitgo (commented):
just a few last questions and cleanup from me here

internal/cloud/atlas/groups_test.go
jandersongo: specifically for testing, we pass in the `itemsPerPage` as an option (`Groups` doesn't take in any params).
setting `itemsPerPage = 1` in the query param to forces the fetch for more groups via links. (otherwise, without the param, all groups would be returned in the first request)

internal/cloud/atlas/groups_test.go
```diff
@@ -49,6 +51,28 @@ func TestAtlasGroups(t *testing.T) {
...
+			numGroups int
+		}{
+			{numGroups: 1},
+			{numGroups: 3},
+			{numGroups: 5},
+		} {
+			t.Run(fmt.Sprintf("when num items allowed per page is %d", tc.numGroups), func(t *testing.T) {
+				client := newAuthClient(t)
+				options := api.RequestOptions{Query: map[string]string{"itemsPerPage": fmt.Sprint(tc.numGroups)}}
+				groups, err := client.FetchGroups(u.AtlasServerURL()+"/api/public/v1.0/groups", options)
```
makesitgo, line 68: I think you could theoretically get around this "exporting a function for testing purposes" by mocking `cloud.Client` here and returning a chain of responses (per call) that force you to iteratively fetch the expected number of groups

it doesn't have to be "live data" to prove that your recursion is working here (which is ultimately what this test wants to do)

makesitgo: ignore me... doing this sort of approach would either retweak the entire `Atlas` client interface or doing the same "export something for testing purposes" here.  not worth it, this (indirectly) proves our recursion is working as expected as well

internal/cloud/atlas/client.go
```diff
@@ -23,6 +23,7 @@ const (
 // Client is a MongoDB Cloud Atlas client
 type Client interface {
 	Groups() ([]Group, error)
+	FetchGroups(path string, options api.RequestOptions) ([]Group, error)
```
makesitgo, line 26: [question] I'll relent that this needs to be its own function and exported for the time being, but does it also need to be part of the interface?  i'd like to limit the impact of "coding for testing purposes" here

internal/cloud/atlas/groups.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/10gen/realm-cli/internal/utils/api"
 )
 
+// GroupsPath is the route for group API handlers
```
makesitgo, line 10: [nit] can remove this docstring now

internal/cloud/atlas/groups.go
```diff
@@ -39,5 +50,16 @@ func (c *client) Groups() ([]Group, error) {
 	if err := json.NewDecoder(res.Body).Decode(&groupRes); err != nil {
 		return nil, err
 	}
-	return groupRes.Results, nil
+
+	allGroups := groupRes.Results
+	for _, link := range groupRes.Links {
+		if link.Rel == "next" {
```
makesitgo, line 56: [micro-opt] we end up iterating through _all_ links here no matter what... I might suggest doing something like:

```go
for _, link := range groupRes.Links {
  if link.Rel != "next" {
    continue
  }
  nextGroups, err := ...
  // ...
  allGroups = append(allGroups, nextGroups...)
  break
}
```

---

also, can we move `"next"` to a const here: `linkRelNext = "next"`?

makesitgo (approved):
final few bits of cleanup but should be good to go afterwards... nice work!

🆒  beans

internal/cloud/atlas/client.go
jandersongo: I believe it does need to be part of the interface ... in order to be able to access *client

makesitgo: yeah you're right... back to the drawing board on this one.  I'd hold off on the merge for now, but I really want to get this into a less hacky-state

internal/cloud/atlas/groups.go
```diff
@@ -17,27 +19,67 @@ type Group struct {
 	Name string `json:"name"`
 }
 
-type groupResponse struct {
+// Link contains the url to the next for a set of groups and its current position in the chain
```
makesitgo, line 22: [meganit] "link" is a much more general concept than what we're specifically using it for here
```suggestion
// Link contains additional data related to the response
```
useful article: https://restfulapi.net/hateoas/

internal/cloud/atlas/groups.go
```diff
@@ -17,27 +19,67 @@ type Group struct {
...
+type Link struct {
+	Href string `json:"href"`
+	Rel  string `json:"rel"`
+}
+
+// Groups represent the list of groups retrieved along with the links to retrieve more
```
makesitgo, line 28: [meganit/optional] I usually go dumb/naive with these types of docstrings
```suggestion
// Groups represents the groups response from Atlas
```

internal/cloud/atlas/groups.go
```diff
@@ -17,27 +19,67 @@ type Group struct {
...
+	allGroups := groups.Results
+	for _, link := range groups.Links {
+		if link.Rel != linkRelNext {
+			continue
+		}
+		res, err := c.Groups(link.Href, false)
+		if err != nil {
+			return nil, err
+		}
+		nextGroups, err := nextGroups(c, res)
```
makesitgo, line 77: [optional] nothing's wrong here, but we are shadowing the `nextGroups` function name with a `var` here... it's probably worth adding an extra/unique name here (maybe rename the function to `fetchNextGroups`?)

makesitgo (approved):
last few bits and questions from me

🆒  beans

PR #300, opened 2022-05-06
BAAS-13231: automatically open browser for ‘realm-cli login’ command

jandersongo:


jandersongo:
found a nifty mock browser that was used for testing https://pkg.go.dev/github.com/langered/gonedrive/fakes/mock_browser
but it seems pretty invasive to the existing code. 

wondering if there is a better way to test that browser...

internal/utils/browser.go
```diff
@@ -0,0 +1,18 @@
+package utils
+
+import (
+	"github.com/pkg/browser"
+)
+
+// BrowserInterface provides access to browser
+type BrowserInterface interface {
```
makesitgo, line 8: [question] what was the motivation for placing this in a `utils` package and as its own interface?

I'm wondering if it makes sense to somehow merge in the `OpenBrowser(url string)` method onto our existing `terminal.UI` construct?

internal/utils/browser.go
```diff
@@ -0,0 +1,18 @@
...
+	Open(url string) error
+}
+
+// Browser is a mocked browser
+type Browser struct{}
+
+// Open opens the default browser.
+func (*Browser) Open(url string) error {
+	return browser.OpenURL(url)
```
makesitgo, line 17: [question] I need to look into this package more closely myself, but are there any options/configurations to provide when calling `OpenURL`?  if, at the end of the day, we're just leveraging a package's static function call, having this be it's own "thing" (with interface and all) may be overkill

internal/utils/browser.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/pkg/browser"
+)
+
+// BrowserInterface provides access to browser
+type BrowserInterface interface {
+	Open(url string) error
+}
+
+// Browser is a mocked browser
+type Browser struct{}
```
makesitgo, line 13: [style/nit]

naming an interface with `Interface` as the suffix (or `I` as the prefix... ugh!) is something worth avoiding (it never adds any _real_ meaning to the name other than heft).

while the caveat of "its always best to just start with the `struct` and not an `interface`" holds true everywhere, I think I see where you're going with the interface here for testing/mocking purposes

with that in mind, a fairly typical pattern for this (to help avoid the `Interface` naming suffix would be:

```go
type MyThing interface { ... }

func NewMyThing(...) MyThing {
  return myThing{...}
}

// myThing doesn't need to be exported b/c it inherently implements MyThing
// which is the only type we're really exporting from this package (and the constructor)
type myThing struct { ... }
```

makesitgo (commented):
with a small tweak to the `UI` interface I think we can drastically reduce the footprint of this PR's changes

curious as to your thoughts on all of this though (and if we have any options for configuring the "open browser" part of this e.g separate tab/window, size of browser, bring to front, etc)

internal/utils/browser.go
jandersongo: this is the package if you're interested in checking it out: https://github.com/pkg/browser
to my knowledge it's a very basic package that doesn't offer options/configurations

go.mod
```diff
@@ -11,12 +11,14 @@ require (
 	github.com/edaniels/digest v0.0.0-20170923160545-b81e9c4ee11c
 	github.com/edaniels/golinters v0.0.3 // indirect
 	github.com/fatih/color v1.10.0
+	github.com/golang/mock v1.6.0 // indirect
```
makesitgo, line 14: I'm going to guess the `// indirect` isn't actually true here and will use this moment to point out: it's been a minute since we ran `go mod tidy` in this codebase.  probably worth doing so now and cleaning things up a bit

internal/terminal/browser.go
```diff
@@ -0,0 +1,22 @@
+package terminal
+
+import "github.com/pkg/browser"
+
+// go:generate mockgen -destination=internal/utils/test/mock/browser.go -package=mock github.com/10gen/realm-cli/internal/terminal Browser
```
makesitgo, line 5: this is super :neat: but probably overkill...  since we already have the `terminal.UI` interface and its own (manual) mock for it, is it possible for us to just:

1) keep `OpenBrowser(url string) error` on the `UI` interface
2) remove the `browser Browser` field from the `terminal.ui` struct
3) call `browser.OpenURL(url)` in the impl of the `(*terminal.ui).OpenBrowser(string)` impl
4) add a `OpenURLFn func(url string) error` to the `mock.ui` definition in `internal/utils/test/mock/terminal_ui` (and leverage it the way we usually leverage "mock fns")

it seems like currently, the mock exists to determine if the browser was opened to a certain URL, and I think that manual mock fn could serve a similar purpose (without the addition of a new package and a `//go:generate`)

(btw, I **love** `go:generate` for a lot of the things, it hurts me here to suggest ripping it out... but I do think we can end up a bit simpler without it)

---

this likely has the added affect of not changing the constructor for a `terminal.NewUI` as well

internal/utils/test/mock/terminal_ui.go
```diff
@@ -88,6 +89,26 @@ func NewConsoleWithOptions(options UIOptions, writers ...io.Writer) (*expect.Con
...
 		console.Tty(),
 		console.Tty(),
+		terminal.NewUIBrowser(),
+	)}
+
+	return console, ui, nil
+}
+
+// NewVT10XConsoleWithBrowser returns a *expect.Console and mock terminal UI that write to the buffer
+func NewVT10XConsoleWithBrowser(browser terminal.Browser) (*expect.Console, terminal.UI, error) {
```
makesitgo, line 99: I can't comment on it, but if you don't want to have to add more helpers to this file for "with mocked browser" then I might suggest adding this to `UIOptions`:

type UIOptions struct {
	AutoConfirm bool
	UseColors   bool
	UseJSON     bool
	OpenBrowserFn     func(url string) error
}
```

and then just setting that on the returned `ui` value from each of these helpers

**Review by makesitgo — COMMENTED** [review]
one larger thought about the organization of this new Browser/UI concept... and I then I think it'll just be minor cleanup stuff and otherwise ready to go

File: go.mod [hunk 7]
```diff
@@ -9,23 +9,26 @@ require (
 	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869 // indirect
```
jandersongo, line 9: ran `go mod tidy` 😬

internal/commands/login/command.go
```diff
@@ -76,6 +76,16 @@ func (cmd *Command) Flags() []flags.Flag {
 				},
 			},
 		},
+		flags.BoolFlag{
+			Value: &cmd.inputs.NoBrowser,
+			Meta: flags.Meta{
+				Name:   "no-browser",
+				Hidden: true,
+				Usage: flags.Usage{
+					Description: "Don't try to open a browser session",
```
makesitgo, line 85: we should probably (always) get design's feedback on the wording for these ... but I think in general try to avoid contractions: `don't` = `do not`

also I imagine we wait to hear back from sumedha about the flags for this command in general, but are we sure this wants to be hidden?

internal/commands/login/inputs.go
```diff
@@ -27,11 +30,18 @@ type inputs struct {
 	PrivateAPIKey string
 	Username      string
 	Password      string
+	NoBrowser     bool
 }
 
 func (i *inputs) Resolve(profile *user.Profile, ui terminal.UI) error {
 	u := profile.Credentials()
 
+	if u == (user.Credentials{}) && !i.NoBrowser {
```
makesitgo, line 39: the parentheses look weird, but I'll have learned something today if this works?

I think @andrewzlchen and I ran into this with metrics exporter previously and fell back to declaring:

```go
var (
  emptyThing typeThing
)
```

and then comparing equality against that: `if u == emptyThing { ... }`.  i've never seen this syntax before

internal/commands/login/inputs.go
```diff
@@ -27,11 +30,18 @@ type inputs struct {
...
+		if errBrowser := ui.OpenBrowser(apiKeysPage); errBrowser != nil {
+			_, _ = fmt.Fprintf(os.Stderr, "There was an issue opening your browser\n")
```
makesitgo, line 41: [question] why are we calling `fmt.Fprintf` here?  I'd think our options were:

* return an error, let the `command_factory` handle printing back the message (and terminate command flow)
* use the `terminal.UI` to some debug/warning/error log but let the command continue

(note: there are reasons to choose any of the above approaches based on what behavior we actually want)

internal/commands/login/inputs_test.go
```diff
@@ -16,6 +17,7 @@ func TestLoginInputs(t *testing.T) {
 		prepareProfile func(p *user.Profile)
 		procedure      func(c *expect.Console)
 		test           func(t *testing.T, i inputs)
+		expectBrowser  bool
```
makesitgo, line 20: [meganit] seems like you could have saved yourself some key strokes if you flipped this to `expectNoBrowser bool` 😬

internal/commands/login/inputs_test.go
```diff
@@ -167,9 +177,36 @@ func TestLoginInputs(t *testing.T) {
...
+				c.SendLine("username")
+				c.ExpectString("Private API Key")
+				c.SendLine("password")
+				c.ExpectEOF()
+			},
+			test: func(t *testing.T, i inputs) {
+				assert.Equal(t, "username", i.PublicAPIKey)
+				assert.Equal(t, "password", i.PrivateAPIKey)
+			},
+			expectBrowser: false,
```
makesitgo, line 195: [style/nit] I tend to leave zero values out of the `tc` declarations for brevity/ease-of-reading

internal/commands/login/inputs_test.go
```diff
@@ -184,6 +221,10 @@ func TestLoginInputs(t *testing.T) {
 
 			assert.Nil(t, tc.inputs.Resolve(profile, ui))
 
+			if tc.expectBrowser {
```
makesitgo, line 224: I don't love wrapping test asserts within conditionals like this (I know its easy/full-proof here, but it's a slippery slope)

rather than listen to my `expectNoBrowser bool` comment above, seems like what you could actually do is have `expectedURL string` and set that to `apiKeysPage` everywhere you currently have `expectBrowser: true`

then this assert simply becomes: `assert.Equal(t, tc.expectedURL, expectedURL)`
(note: the flip of args there, the CLI likes having `expected` first then `actual` ... I have very strong feelings both ways depending on the day on whether I like that pattern or not 😬 )

makesitgo (commented):
another batch of comments here for you while we wait on product/design's input on things

depending on where all that lands, we might want to file the appropriate TODO for `--launch-access-page` (with the included stretch functionality of allowing `--project` to be specified)

internal/commands/login/inputs.go
andrewzlchen: oo that's cool. I could try to drive-by this change in the next store PR!

internal/commands/login/inputs.go
jandersongo: Assuming we want the login to continue even if the browser doesn't work, I went with opt #2 --> using `terminal.UI `to output the error

internal/commands/login/inputs_test.go
```diff
@@ -184,10 +205,38 @@ func TestLoginInputs(t *testing.T) {
 
 			assert.Nil(t, tc.inputs.Resolve(profile, ui))
 
+			assert.Equal(t, tc.expectedURL, expectedURL)
```
makesitgo, line 208: [question] I think it matters not, but maybe we want to move this assert to after the `<-doneCh` receive?  that's technically the point when "inputs are done resolving" (even if the browser portion of input resolution is synchronous currently)

internal/commands/login/inputs_test.go
```diff
@@ -184,10 +205,38 @@ func TestLoginInputs(t *testing.T) {
...
+
 			console.Tty().Close() // flush the writers
 			<-doneCh              // wait for procedure to complete
 
 			tc.test(t, tc.inputs)
 		})
 	}
+
+	t.Run("should output error message if browser cannot be opened", func(t *testing.T) {
+		errMessage := "there was an issue opening your browser\n"
```
makesitgo, line 218: [optional] also feel [meh] about needing to DRY this error message for assert purposes (also.. so formal with the newline at the end!)

makesitgo (approved):
nothing major from me... it all looks good

🆒  beans

internal/commands/login/inputs_test.go
jandersongo: - will remove `errMessage` var
- new line is needed for assert :/ 

```
=== RUN   TestLoginInputs/should_output_error_message_if_browser_cannot_be_opened
    inputs_test.go:236: 
        failed to assert equals ( actual, expected )
        	"there was an issue opening your browser\n"
        	"there was an issue opening your browser"
--- FAIL: TestLoginInputs (0.00s)
    --- FAIL: TestLoginInputs/should_output_error_message_if_browser_cannot_be_opened (0.00s)
```

makesitgo: yeah, the newline for the assert makes sense, but the error message in `errors.New` didnt need it, right?

jandersongo: correct, I removed it

PR #302, opened 2022-05-10
REALMC-12577: Add Data API Config support to the CLI

malexandert:
- add support for parsing and writing data_api_config.json
- add data api product


malexandert:
With the import support PR in, these tests should pass on evergreen now, so I'm opening this up for review

internal/local/structure_v2.go
```diff
@@ -34,6 +34,7 @@ type AppStructureV2 struct {
 	Sync                  SyncStructure                     `json:"sync,omitempty"`
 	Secrets               SecretsStructure                  `json:"secrets,omitempty"`
 	LogForwarders         []map[string]interface{}          `json:"log_forwarders,omitempty"`
+	DataAPIConfig         map[string]interface{}            `json:"data_api_config,omitempty"`
```
caiyingshen, line 37: [q] as a more general question, when do we need to create a struct and when can we just use `map[string]interface{}`?

caiyingshen (approved):
LGTM 😎  looks pretty 🧼

gabbyasuncion (approved):
Looks good!!

PR #306, opened 2022-06-03
BAAS-14024: add project names and ids to output of 'whoami' command

jandersongo:


internal/commands/whoami/command.go
```diff
@@ -10,7 +12,7 @@ import (
 var CommandMeta = cli.CommandMeta{
 	Use:         "whoami",
 	Description: "Display information about the current user",
-	// TODO(REALMC-8832): this is an example of where standardizing cli, comamnd and flag names
+	// TODO(REALMC-8832): this is an example of where standardizing cli, command and flag names
 	// into a shared package would be helpful, to reduce coupling the command packages to each other
 	// (since this HelptText creates a "whoami depends on login" package cycle)
```
makesitgo, line 17: missed one...
```suggestion
	// (since this HelpText creates a "whoami depends on login" package cycle)
```

on a more serious note, this ticket has since been worked on and closed, so we should figure out what to do here.  the intent of this comment was remove the potential for this help text to ever get stale, if we possibly changed the `login` command name to something else

while it feels like overkill, it's also not wrong.  can we file a new ticket under the `BAAS-13516` epic:

`Create centralized file for all CLI command and flag names`

and in the description, just add that this file could/should probably live in the `cli/cli.go` package file.  also, we should update this TODO with the new ticket number

internal/commands/whoami/command.go
```diff
@@ -51,5 +53,24 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 	}
 
 	ui.Print(terminal.NewTextLog("Currently logged in user: %s (%s)", userDisplay, userSecret))
+
+	groups, groupsErr := clients.Atlas.Groups()
```
makesitgo, line 57: [style/nit] you've been in a "named err" type of mood lately, eh?

also, b/c this is a `== nil` case, you could inline this with the `if` itself (and keep a tighter scope)

internal/commands/whoami/command.go
```diff
@@ -51,5 +53,24 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+	if groupsErr == nil {
+		headerProjectID := "Project ID"
```
makesitgo, line 59: [question] should these just be consts in the file, rather than vars within the command handler?

internal/commands/whoami/command.go
```diff
@@ -51,5 +53,24 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
...
+		headerProjectName := "Project Name"
+		tableRows := make([]map[string]interface{}, 0, len(groups))
+		for _, group := range groups {
+			tableRows = append(tableRows, map[string]interface{}{
+				headerProjectID:   group.ID,
+				headerProjectName: group.Name,
+			})
+		}
+		ui.Print(terminal.NewTableLog(
+			fmt.Sprintf("Projects available (%d)", len(groups)),
```
makesitgo, line 69: is this potentially a **very large** list of projects?  wondering if it'd be desirable to make this output "opt-in" via a flag so that `realm-cli whoami` doesn't just start blasting users with more information than they're used to

makesitgo (commented):
just a few small thoughts/questions here before giving this green light

internal/commands/whoami/command.go
jandersongo: https://jira.mongodb.org/browse/BAAS-14174

internal/commands/whoami/command.go
```diff
@@ -51,5 +76,23 @@ func (cmd *Command) Handler(profile *user.Profile, ui terminal.UI, clients cli.C
 	}
 
 	ui.Print(terminal.NewTextLog("Currently logged in user: %s (%s)", userDisplay, userSecret))
+
+	if cmd.inputs.ShowProjects {
+		if groups, err := clients.Atlas.Groups(); err == nil {
```
makesitgo, line 81: I think we want to merge in #299 first and then update this to use the new `atlas.AllGroups` function

internal/commands/whoami/inputs.go
```diff
@@ -0,0 +1,14 @@
+package whoami
+
+import (
+	"github.com/10gen/realm-cli/internal/cli/user"
+	"github.com/10gen/realm-cli/internal/terminal"
+)
+
+type inputs struct {
```
makesitgo, line 8: [style/nit] for such a simple struct, we don't need a separate file for this.  just define it all in the `command.go` file

internal/commands/whoami/inputs.go
```diff
@@ -0,0 +1,14 @@
...
+	ShowProjects bool
+}
+
+func (i *inputs) Resolve(profile *user.Profile, ui terminal.UI) error {
```
makesitgo, line 12: for the no-op `Resolve` function, it's actually not necessary (hooray go and its interfaces!)

also, fwiw, if you _did_ need to implement `Resolve` for something here, it would only ever get called if you also then implemented the `CommandInputs` interface on the `Command` struct itself

makesitgo (commented):
just a last few things from me, but i'm going to hold off on the ✅  until we get the "fetch all groups" pr merged first

internal/commands/whoami/command.go
```diff
@@ -1,27 +1,57 @@
...
+	// (since this HelpText creates a "whoami depends on login" package cycle)
 	HelpText: `Displays a table that includes your Public and redacted Private Atlas
 programmatic API Key (e.g. ********-****-****-****-3ba985aa367a). No session
 data will be surfaced if you are not logged in.
 
 NOTE: To log in and authenticate your session, use "realm-cli login"`,
 }
 
+type inputs struct {
+	showProjects bool
```
makesitgo, line 28: no need to change anything here, but I feel like it's worth bringing attention to:

_IF_ you needed to go the route (during input resolution) to ask multiple questions (as opposed to just individual calls to `ui.AskOne`) then you'd want this field to be capitalized/exported

makesitgo (approved):
easy like Monday morning

🆒  beans

PR #307, opened 2022-06-06
BAAS-14096: add hidden 'profiles list' command

jandersongo:
```
jamie.anderson@M-C02F15DGMD6M mongodb % ./realm-cli-next profiles list
Found 6 profiles
  default.yaml
  dev.yaml
  local.yaml
  local_dev.yaml
  prod.yaml
  qa.yaml
```

### DIFF
--- a/internal/cli/user/auth.go
+++ b/internal/cli/user/auth.go
@@ -18,20 +18,20 @@ type Credentials struct {
 	Password      string
 }
 
-// RedactedPrivateAPIKey returns the user's private API key with sensitive information redacted
-func (creds Credentials) RedactedPrivateAPIKey() string {
-	parts := strings.Split(creds.PrivateAPIKey, "-")
+// RedactKey returns the user's private API key with sensitive information redacted
+func RedactKey(privateAPIKey string) string {
+	parts := strings.Split(privateAPIKey, "-")
 	switch len(parts) {
 	case 0:
 		return ""
 	case 1:
-		return redact(parts[0])
+		return Redact(parts[0])
 	default:
 		lastIdx := len(parts) - 1
 
 		out := make([]string, len(parts))
 		for i := 0; i < lastIdx; i++ {
-			out[i] = redact(parts[i])
+			out[i] = Redact(parts[i])
 		}
 		out[lastIdx] = parts[lastIdx]
 
@@ -39,11 +39,7 @@ func (creds Credentials) RedactedPrivateAPIKey() string {
 	}
 }
 
-// RedactedPassword returns the user's password with sensitive information redacted
-func (creds Credentials) RedactedPassword() string {
-	return redact(creds.Password)
-}
-
-func redact(s string) string {
+//Redact returns sensitive information redacted
+func Redact(s string) string {
 	return strings.Repeat("*", len(s))
 }
--- a/internal/commands/commands.go
+++ b/internal/commands/commands.go
@@ -8,6 +8,7 @@ import (
 	"github.com/10gen/realm-cli/internal/commands/login"
 	"github.com/10gen/realm-cli/internal/commands/logout"
 	"github.com/10gen/realm-cli/internal/commands/logs"
+	"github.com/10gen/realm-cli/internal/commands/profile"
 	"github.com/10gen/realm-cli/internal/commands/pull"
 	"github.com/10gen/realm-cli/internal/commands/push"
 	"github.com/10gen/realm-cli/internal/commands/schema"
@@ -201,6 +202,21 @@ var (
 			{
 				Command:     &accesslist.CommandDelete{},
 				CommandMeta: accesslist.CommandMetaDelete,
+			},
+		},
+	}
+
+	Profiles = cli.CommandDefinition{
+		CommandMeta: cli.CommandMeta{
+			Use:         "profiles",
+			Aliases:     []string{"profile"},
+			Description: "Manage the profiles of your local CLI environment",
+			Hidden:      true,
+		},
+		SubCommands: []cli.CommandDefinition{
+			{
+				Command:     &profile.CommandList{},
+				CommandMeta: profile.CommandMetaList,
 			},
 		},
 	}
--- a/internal/commands/profile/list.go
+++ b/internal/commands/profile/list.go
@@ -1,0 +1,54 @@
+package profile
+
+import (
+	"fmt"
+
+	"github.com/10gen/realm-cli/internal/cli"
+	"github.com/10gen/realm-cli/internal/cli/user"
+	"github.com/10gen/realm-cli/internal/local"
+	"github.com/10gen/realm-cli/internal/terminal"
+)
+
+const (
+	headerName = "Profile"
+	headerKey  = "API Key"
+)
+
+// CommandMetaList is the command meta for the `profile list` command
+var CommandMetaList = cli.CommandMeta{
+	Use:         "list",
+	Aliases:     []string{"ls"},
+	Display:     "profiles list",
+	Description: "List the profiles of your local CLI environment",
+}
+
+// CommandList is the `profile list` command
+type CommandList struct{}
+
+// Handler is the command handler
+func (cmd *CommandList) Handler(profile *user.Profile, ui terminal.UI, clients cli.Clients) error {
+	environments, err := local.LoadEnvironments()
+	if err != nil {
+		return err
+	}
+
+	rows := make([]map[string]interface{}, len(environments))
+	for _, v := range environments {
+		credentials := "(logged out)"
+		if v.Credentials.PublicAPIKey != "" && v.Credentials.PrivateAPIKey != "" {
+			credentials = fmt.Sprintf("%s (%s)", v.Credentials.PublicAPIKey, user.RedactKey(v.Credentials.PrivateAPIKey))
+		}
+		rows = append(rows, map[string]interface{}{
+			headerName: v.Name,
+			headerKey:  credentials,
+		})
+	}
+
+	ui.Print(terminal.NewTableLog(
+		fmt.Sprintf("Found %d profile(s)", len(environments)),
+		[]string{headerName, headerKey},
+		rows...,
+	))
+
+	return nil
+}
--- a/internal/local/environment.go
+++ b/internal/local/environment.go
@@ -1,0 +1,67 @@
+package local
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
+	"github.com/10gen/realm-cli/internal/cli/user"
+
+	"gopkg.in/yaml.v2"
+)
+
+// Environment contains the name, full filepath, and credentials of a profile
+type Environment struct {
+	Name        string
+	Filepath    string
+	Credentials Credentials
+}
+
+// Credentials are the apikeys associated with the profiles
+type Credentials struct {
+	PublicAPIKey  string `yaml:"public_api_key"`
+	PrivateAPIKey string `yaml:"private_api_key"`
+}
+
+// LoadEnvironments returns a list of each profile's environment containing name, filepath, and api keys
+func LoadEnvironments() ([]Environment, error) {
+	dir, err := user.HomeDir()
+	if err != nil {
+		return nil, err
+	}
+
+	dirEntryList, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return nil, err
+	}
+
+	environments := make([]Environment, 0, len(dirEntryList))
+	for _, v := range dirEntryList {
+		if v.IsDir() {
+			continue
+		}
+		filename := v.Name()
+		ext := filepath.Ext(filename)
+		if ext != "."+user.ProfileType {
+			continue
+		}
+		name := filename[0 : len(filename)-len(ext)]
+
+		data, err := ioutil.ReadFile(filepath.Join(dir, filename))
+		if err != nil {
+			return nil, fmt.Errorf("failed to read profile: %s", name)
+		}
+
+		var profile map[string]Credentials
+		if err := yaml.Unmarshal(data, &profile); err != nil {
+			return nil, fmt.Errorf("failed to read profile: %s", name)
+		}
+
+		environments = append(environments, Environment{
+			Name:        name,
+			Filepath:    filepath.Join(dir, v.Name()),
+			Credentials: profile[name],
+		})
+	}
+	return environments, nil
+}
--- a/internal/local/environment_test.go
+++ b/internal/local/environment_test.go
@@ -1,0 +1,33 @@
+package local
+
+import (
+	"testing"
+
+	"github.com/10gen/realm-cli/internal/cli/user"
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/10gen/realm-cli/internal/utils/test/mock"
+)
+
+func TestLoadProfiles(t *testing.T) {
+	t.Run("should return a list of profile names along with file paths and api keys", func(t *testing.T) {
+		profile, _ := mock.NewProfileFromTmpDir(t, "temp")
+
+		profile.SetCredentials(user.Credentials{PublicAPIKey: "abcde", PrivateAPIKey: "my-private-key"})
+		profile.Save()
+
+		environments, err := LoadEnvironments()
+		assert.Nil(t, err)
+
+		assert.Equal(t,
+			[]Environment{{
+				Name:     profile.Name,
+				Filepath: profile.Path(),
+				Credentials: Credentials{
+					PublicAPIKey:  "abcde",
+					PrivateAPIKey: "my-private-key",
+				},
+			}},
+			environments,
+		)
+	})
+}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -47,6 +47,7 @@ func Run() {
 	cmd.AddCommand(factory.Build(commands.Function))
 	cmd.AddCommand(factory.Build(commands.Schema))
 	cmd.AddCommand(factory.Build(commands.AccessList))
+	cmd.AddCommand(factory.Build(commands.Profiles))
 
 	os.Exit(factory.Run(cmd))
 }
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,7 @@ require (
 	github.com/xtgo/uuid v0.0.0-20140804021211-a0b114877d4c // indirect
 	go.mongodb.org/mongo-driver v1.5.1
 	gopkg.in/segmentio/analytics-go.v3 v3.1.0
+	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
 
 replace github.com/edaniels/golinters => github.com/mongodb-forks/golinters v0.0.4
--- a/internal/cli/user/profile.go
+++ b/internal/cli/user/profile.go
@@ -19,8 +19,10 @@ const (
 	// HostingAssetCacheDir is the hosting asset cache dir
 	HostingAssetCacheDir = ".asset-cache"
 
-	envPrefix   = "realm"
-	profileType = "yaml"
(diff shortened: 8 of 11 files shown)


jandersongo:
thoughts on making this a table instead of a list output and showing the **public api key next to the profile name**?
I often find myself opening the profile.yaml just to check which api keys are used (so I can log in to the ui with them) 

@makesitgo

internal/cli/user/profile_meta.go
```diff
@@ -0,0 +1,40 @@
+package user
```
makesitgo, line 1: [question] thoughts on pushing this logic to the `local` package?

my thinking: most of our logic around `ioutil.ReadDir` and parsing files lives there already.

the potentially confusing bit: in a "naive-naming sense", this leaves us with both a `user.Profile` and a `local.Profile` type (Tim's got me on the crusade against "Meta"-like names now... 😬 )

if we _do_ go this route, I think the eventual `local.LoadProfiles` func should take in a `dir string` arg, which can be provided via the `profile.Dir()` which is available on the `user.Profile` type (which all commands are passed)

fwiw, I think what you have here is fine too... but the "meta" name got me thinking about alternatives here

internal/cli/user/profile_meta.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	dirEntryList, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get CLI profiles: %w", dirErr)
+	}
+
+	profileMetas := make([]ProfileMeta, 0, len(dirEntryList))
+	for _, v := range dirEntryList {
+		if !strings.Contains(v.Name(), profileType) {
```
makesitgo, line 30: obviously wouldn't be able to use the `profileType = "yaml"` const if we moved to the `local` package.  lazily I'd say duplicating that is really fine, but we also already export `HostingAssetCacheDir` from this `user` package to re-use in the `local` package, so exporting `ProfileType` probably isn't the worst thing either

internal/commands/commands.go
```diff
@@ -204,4 +205,19 @@ var (
 			},
 		},
 	}
+
+	Profiles = cli.CommandDefinition{
+		CommandMeta: cli.CommandMeta{
+			Use:         "profiles",
+			Aliases:     []string{"profile"},
+			Description: "Manage your profiles",
```
makesitgo, line 213: thoughts?
```suggestion
			Description: "Manage the profiles of your local CLI environment",
```

internal/commands/profile/list.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/10gen/realm-cli/internal/terminal"
+	"github.com/10gen/realm-cli/internal/utils/flags"
+)
+
+// CommandMetaList is the command meta for the `profile list` command
+var CommandMetaList = cli.CommandMeta{
+	Use:         "list",
+	Aliases:     []string{"ls"},
+	Display:     "profiles list",
+	Description: "List your profiles",
```
makesitgo, line 17: thoughts?
```suggestion
	Description: "List the profiles of your local CLI environment",
```

internal/commands/profile/list.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// CommandList is the `profile list` command
+type CommandList struct {
+	inputs cli.ProjectInputs
```
makesitgo, line 22: are these inputs necessary?

internal/commands/profile/list.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// Flags is the command flags
+func (cmd *CommandList) Flags() []flags.Flag {
+	return []flags.Flag{
+		cli.ProjectFlag(&cmd.inputs.Project),
```
makesitgo, line 28: are these flags necessary?

makesitgo (commented):
I'd be curious to see what the table output looks like compared to this list output

the other idea would be to just list like you have it, but also include ` (<public_api_key>)` appended to the string output (I think we do something similar when listing out projects in the prompt when the user needs to select one)

jandersongo:
> I'd be curious to see what the table output looks like compared to this list output


here's what the table output could look like:
```
mongodb ./realm-cli-next profile list
Found 13 profile(s)
  Profile        API Key                                        
  -------------  -----------------------------------------------
  default        fdnxgyfs (********-****-****-****-3af5d01d2a93)
  dev            bovvwwyq (********-****-****-****-51d2293df943)
  dev_bovvwwyq   bovvwwyq (********-****-****-****-51d2293df943)
  dev_hsxtnfab   hsxtnfab (********-****-****-****-d9f860aae509)
  dev_jypmqyzr   jypmqyzr (********-****-****-****-dd1f4f0bac88)
  dev_rcymllrh   rcymllrh (********-****-****-****-1001af33d5a2)
  local          bovvwwyq (********-****-****-****-51d2293df943)
  local_dev      -                                              
  p1             -                                              
  prod           fdnxgyfs (********-****-****-****-3af5d01d2a93)
  prod2          -                                              
  prod_nrkgkfyl  nrkgkfyl (********-****-****-****-24a8a260a7cb)
  qa             kfqyyhlf (********-****-****-****-3a0830881cd9)
```

we could also add a flag for showing api keys (similar to what we did for the whoami `--show-projects` output)

internal/cli/user/profile_meta.go
jandersongo: I think I agree with [May-you](https://mongodb.slack.com/archives/D022Z4CEDPC/p1652194428174429): 

> also wondering if it makes sense to actually put this into package local?
> you could then get away with the sleek-but-potentially-confusing type Profile struct (so that we're left with both a user.Profile and local.Profile)
> saying that out loud, i actually don't like it, but at least it was mentioned
> 

unless we're really striving for consistency in keeping the `ioutil.ReadDir` functionality all in the `local` package, I think it makes sense to keep the profile parsing logic together

makesitgo:
I don't think I have a strong opinion on one output format over the other, what do you think?

my only other thought would be to maybe have two separate sections of output: logged in profiles (read: profiles with api key data associated with them) and logged out profiles (read: profiles with the api key data cleared from them).  in that sense, you could likely do a best of both and table the logged in profiles while listing out the logged out profiles (no extra data to show anyways)

jandersongo:
> I don't think I have a strong opinion on one output format over the other, what do you think?
> 
> my only other thought would be to maybe have two separate sections of output: logged in profiles (read: profiles with api key data associated with them) and logged out profiles (read: profiles with the api key data cleared from them). in that sense, you could likely do a best of both and table the logged in profiles while listing out the logged out profiles (no extra data to show anyways)


hhmmm im not sure I like two separate tables/lists. how about something like this, indicating the profile is logged out?
```
➜  mongodb ./realm-cli-next profile list
Found 13 profile(s)
  Profile        API Key                                        
  -------------  -----------------------------------------------
  default        fdnxgyfs (********-****-****-****-3af5d01d2a93)
  dev            bovvwwyq (********-****-****-****-51d2293df943)
  dev_bovvwwyq   bovvwwyq (********-****-****-****-51d2293df943)
  dev_hsxtnfab   hsxtnfab (********-****-****-****-d9f860aae509)
  dev_jypmqyzr   jypmqyzr (********-****-****-****-dd1f4f0bac88)
  dev_rcymllrh   rcymllrh (********-****-****-****-1001af33d5a2)
  local          bovvwwyq (********-****-****-****-51d2293df943)
  local_dev      (logged out)                                   
  p1             (logged out)                                   
  prod           fdnxgyfs (********-****-****-****-3af5d01d2a93)
  prod2          (logged out)                                   
  prod_nrkgkfyl  nrkgkfyl (********-****-****-****-24a8a260a7cb)
  qa             kfqyyhlf (********-****-****-****-3a0830881cd9)
```

internal/cli/user/profile_meta.go
jandersongo: sold! 
I've made the following changes in the latest commit:

- moved profile_meta.go from `user` to `local` package and renamed it to environment.go
- renamed 
`ProfileMeta` to `Environment`
`user.Profiles` to `local.LoadProfileEnvs()`

makesitgo:
I think `(logged out)` looks fine!

internal/local/environment.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Environment contains the name, full filepath, and api keys of a profile
+type Environment struct {
+	Name        string
+	Filepath    string
+	Credentials user.Credentials
+}
+
+// LoadProfileEnvs returns a list of each profile's environment containing name, filepath, and api keys
+func LoadProfileEnvs() ([]Environment, error) {
```
makesitgo, line 20: [naming/nit] `LoadEnvironments` to match the array of the type this func returns?  I like that [nit] more than suggesting changing the `struct` name to `ProfileEnv`/`ProfileEnvironment`...

internal/local/environment.go
```diff
@@ -0,0 +1,56 @@
...
+	dir, dirErr := user.HomeDir()
```
makesitgo, line 21: [naming/meganit] would keep consistent with either the `namedErr` or not

internal/local/environment.go
```diff
@@ -0,0 +1,56 @@
...
+	if dirErr != nil {
+		return nil, fmt.Errorf("failed to get CLI profiles: %w", dirErr)
```
makesitgo, line 23: this returned `err` is ultimately formatted with `"profiles list failed: %w"`, so the wrapped error here is not entirely necessary (it ends up sounding a bit redundant)

same comment goes for below

internal/local/environment.go
```diff
@@ -0,0 +1,56 @@
...
+		if err := profile.Load(); err != nil {
+			return nil, fmt.Errorf("failed to load profile: %s", profile.Name)
+		}
+
+		environments = append(environments, Environment{
+			Name:        name,
+			Filepath:    filepath.Join(dir, v.Name()),
+			Credentials: profile.Credentials(),
+		})
+
```
makesitgo, line 52: [style/nit] extra newline here, and arguably two lines below as well on L#54

internal/local/environment.go
```diff
@@ -0,0 +1,56 @@
...
+			continue
+		}
+
+		name := strings.TrimSuffix(v.Name(), filepath.Ext(v.Name()))
+
+		profile, err := user.NewProfile(name)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load profile: %s", profile.Name)
+		}
+		if err := profile.Load(); err != nil {
```
makesitgo, line 43: there might be unforeseen ramifications using `viper` to `Load` each profile individually (there might be unforeseen ramifications to _not_ using `viper` to `Load` each profile 😬 )

due to the command we're supporting here, I'd rather us take the simple/naive approach first before we get too clever with it

as such, we should just define `yaml` tags on w/e `Credentials` struct we end up using and get that data via `yaml.Unmarshal`

i'll leave it to you if we end up adding `yaml` tags to `user.Credentials` type or define an entirely new type or set of fields here in or alongside `Environment`

internal/local/environment.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	dirEntryList, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get CLI profiles: %w", dirErr)
+	}
+
+	environments := make([]Environment, 0, len(dirEntryList))
+	for _, v := range dirEntryList {
+		if !strings.Contains(v.Name(), user.ProfileType) {
```
makesitgo, line 33: would `HasSuffix` would here as a check?  or perhaps even better: `if filepath.Ext(v.Name()) == "."+user.ProfileType`

internal/local/environment.go
```diff
@@ -0,0 +1,56 @@
...
+			continue
+		}
+
+		name := strings.TrimSuffix(v.Name(), filepath.Ext(v.Name()))
```
makesitgo, line 37: if we go with the extension style check above (and keep the extension in its own `var`), we could make this:

```go
name := v.Name()
ext := filepath.Ext(name)
if ext == "."+user.ProfileType { continue }
name = name[0:len(name)-len(ext)]
```

internal/local/environment_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/10gen/realm-cli/internal/cli/user"
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/10gen/realm-cli/internal/utils/test/mock"
+)
+
+func TestLoadProfiles(t *testing.T) {
+	t.Run("should return a list of profile names, file paths and api keys", func(t *testing.T) {
+		profile, _ := mock.NewProfileFromTmpDir(t, "temp")
+
+		profile.SetCredentials(user.Credentials{PublicAPIKey: "abcde", PrivateAPIKey: "*****"})
```
makesitgo, line 15: less meaningful here but an equal sentiment about using `some-super-secret-key` value instead of simply asterisks for the value here

makesitgo (commented):
last big batch of comments I'm guessing here and then should be good to go

internal/commands/profile/list.go
```diff
@@ -0,0 +1,54 @@
...
+func (cmd *CommandList) Handler(profile *user.Profile, ui terminal.UI, clients cli.Clients) error {
+	profileEnvs, err := local.LoadProfileEnvs()
+	if err != nil {
+		return err
+	}
+
+	rows := make([]map[string]interface{}, 0)
+	for _, v := range profileEnvs {
+		credentials := "(logged out)"
+		if v.Credentials != (user.Credentials{}) {
```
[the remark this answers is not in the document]
makesitgo: when in doubt, use `git blame`/`git history` ... which in a roundabout way here would direct you to a PR titled `"Support local-userpass auth for local development and testing"` (which unfortunately has no JIRA associated with it)

you might be able to find that JIRA, but I think that title gets us our answer: we don't need to worry about supporting the display of this for any normal user

internal/local/environment.go
```diff
@@ -0,0 +1,62 @@
...
+		profile := make(map[string]user.Credentials)
+		err = yaml.Unmarshal(data, &profile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read profile: %s", name)
+		}
+
+		environments = append(environments, Environment{
+			Name:     name,
+			Filepath: filepath.Join(dir, v.Name()),
+			Credentials: user.Credentials{
```
makesitgo, line 55: this can be simplified:

```go
// humor me and let's reference a local `Credentials` type here
type Credentials struct {
  PublicAPIKey string `yaml:"public_api_key"`
  PrivateAPIKey string `yaml:"private_api_key"`
}

type Environment struct {
  Name string
  Filepath string
  Credentials Credentials
}

// then in LoadEnvironments impl:
var profile map[string]Credentials
if err := yaml.Unmarshal(data, &profile); err != nil { ... }
environments = append(environments, Environment{
  Name: name,
  Filepath: filepath.Join(dir, v.Name()),
  Credentials: profile[name],
})
```

internal/local/environment_test.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"testing"
+
+	"github.com/10gen/realm-cli/internal/cli/user"
+	"github.com/10gen/realm-cli/internal/utils/test/assert"
+	"github.com/10gen/realm-cli/internal/utils/test/mock"
+)
+
+func TestLoadProfiles(t *testing.T) {
+	t.Run("should return a list of profile names, file paths and api keys", func(t *testing.T) {
```
makesitgo, line 12: [nit] ditch the commas in the test description

internal/commands/profile/list.go
```diff
@@ -0,0 +1,54 @@
...
+	Display:     "profiles list",
+	Description: "List the profiles of your local CLI environment",
+}
+
+// CommandList is the `profile list` command
+type CommandList struct{}
+
+// Handler is the command handler
+func (cmd *CommandList) Handler(profile *user.Profile, ui terminal.UI, clients cli.Clients) error {
+	profileEnvs, err := local.LoadEnvironments()
```
makesitgo, line 30: [naming/nit] use `envs` or `environments` here to match the function name

internal/commands/profile/list.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return err
+	}
+
+	rows := make([]map[string]interface{}, 0)
```
makesitgo, line 35: missing a `len(profileEnvs)` as the 2nd arg here... this is currently equivalent to `rows := []map[string]interface{}{}`

internal/commands/profile/list.go
```diff
@@ -0,0 +1,54 @@
...
+	for _, v := range profileEnvs {
+		credentials := "(logged out)"
+		if v.Credentials.PublicAPIKey != "" && v.Credentials.PrivateAPIKey != "" {
+			credentials = fmt.Sprintf("%s (%s)", v.Credentials.PublicAPIKey, v.Credentials.RedactedPrivateAPIKey())
```
makesitgo, line 39: i'm guessing the `RedactedPrivateAPIKey()` function is what led you to re-use the `user.Credentials` type and add `yaml` tags to it...

we can/should just refactor the `RedactedPrivateAPIKey` logic into a static function (read: not a receiver function on the `user.Credentials` type).  something like `user.RedactAPIKey(value string) string` (and `user.RedactPassword(value string) string`

internal/local/environment_test.go
jandersongo: is this just a styling preference or does it actually mess with the tests?

makesitgo: styling preference... the `,` right before the `_` (which replaces the space) in `go test...` output makes it a bit clunkier to read

internal/cli/user/auth.go
```diff
@@ -12,15 +12,15 @@ type Session struct {
 
 // Credentials are the user credentials
 type Credentials struct {
-	PublicAPIKey  string
-	PrivateAPIKey string
+	PublicAPIKey  string `yaml:"public_api_key"`
+	PrivateAPIKey string `yaml:"private_api_key"`
```
makesitgo, line 16: the yaml tags here and above should be removed, right?

internal/cli/user/auth.go
```diff
@@ -40,8 +40,8 @@ func (creds Credentials) RedactedPrivateAPIKey() string {
 }
 
 // RedactedPassword returns the user's password with sensitive information redacted
-func (creds Credentials) RedactedPassword() string {
-	return redact(creds.Password)
+func RedactedPassword(password string) string {
+	return redact(password)
```
makesitgo, line 44: [nit] this function isn't doing much anymore, we can probably just delete `RedactPassword` and export `Redact` to do the same thing... maybe we rename `RedactPrivateAPIKey` to something more general like `RedactKey`?

internal/local/environment_test.go
```diff
@@ -0,0 +1,29 @@
...
+		profile, _ := mock.NewProfileFromTmpDir(t, "temp")
+
+		profile.SetCredentials(user.Credentials{PublicAPIKey: "abcde", PrivateAPIKey: "my-private-key"})
+		profile.Save()
+
+		environments, err := LoadEnvironments()
+		assert.Nil(t, err)
+
+		assert.Equal(t,
+			[]Environment{{Name: profile.Name, Filepath: profile.Path(), Credentials: Credentials{
```
makesitgo, line 22: [style/nit] would put the entry for the one `Environment` into multi-line syntax as its own thing (rather than psuedo-inline everything

```go
assert.Equal(t,
  []Environment{{
    Name: profile.Name,
    Filepath: profile.Path(),
    Credentials: Credentials{
      PublicAPIKey: "abcde",
      PrivateAPIKey: "my-private-key",
    },
  }},
  environments,
```

makesitgo (approved):
just a few small things left from me

🆒  beans

PR #308, opened 2022-06-07
BAAS-14085: add 'browser' flag for cli login

jandersongo:
depends on https://github.com/10gen/realm-cli/pull/300

### DIFF
--- a/internal/commands/login/inputs.go
+++ b/internal/commands/login/inputs.go
@@ -30,23 +30,39 @@ type inputs struct {
 	PrivateAPIKey string
 	Username      string
 	Password      string
+	Browser       bool
 }
 
 func (i *inputs) Resolve(profile *user.Profile, ui terminal.UI) error {
 	u := profile.Credentials()
 
-	if u == (user.Credentials{}) {
+	if i.Browser && (i.Username != "" || i.Password != "" || i.PublicAPIKey != "" || i.PrivateAPIKey != "") {
+		return errors.New("credentials will not be authenticated while using browser flag, please login with one or the other")
+	}
+
+	if u == (user.Credentials{}) || i.Browser {
 		if err := ui.OpenBrowser(apiKeysPage); err != nil {
-			ui.Print(terminal.NewErrorLog(errors.New("there was an issue opening your browser")))
+			ui.Print(terminal.NewWarningLog("there was an issue opening your browser"))
 		}
 	}
 
 	var questions []*survey.Question
 
-	switch i.AuthType {
-	case authTypeCloud:
+	switch {
+	case i.Browser:
+		questions = []*survey.Question{
+			{
+				Name:   inputFieldPublicAPIKey,
+				Prompt: &survey.Input{Message: "Public API Key"},
+			},
+			{
+				Name:   inputFieldPrivateAPIKey,
+				Prompt: &survey.Password{Message: "Private API Key"},
+			},
+		}
+	case i.AuthType == authTypeCloud:
 		questions = i.resolveCloudCredentials(u)
-	case authTypeLocal:
+	case i.AuthType == authTypeLocal:
 		questions = i.resolveLocalCredentials(u)
 	default:
 		return fmt.Errorf(
--- a/internal/commands/login/inputs_test.go
+++ b/internal/commands/login/inputs_test.go
@@ -21,7 +21,7 @@ func TestLoginInputs(t *testing.T) {
 		expectedURL    string
 	}{
 		{
-			description: "Should prompt for public api key when not provided for cloud type",
+			description: "should prompt for public api key when not provided for cloud type",
 			inputs: inputs{
 				AuthType:      authTypeCloud,
 				PrivateAPIKey: "password",
@@ -37,7 +37,7 @@ func TestLoginInputs(t *testing.T) {
 			expectedURL: apiKeysPage,
 		},
 		{
-			description: "Should prompt for private api key when not provided for cloud type",
+			description: "should prompt for private api key when not provided for cloud type",
 			inputs: inputs{
 				AuthType:     authTypeCloud,
 				PublicAPIKey: "username",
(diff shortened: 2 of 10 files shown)


internal/commands/login/inputs.go
```diff
@@ -66,19 +71,23 @@ func (i *inputs) Resolve(profile *user.Profile, ui terminal.UI) error {
 func (i *inputs) resolveCloudCredentials(u user.Credentials) []*survey.Question {
```
jandersongo, line 71: note: only prompting for `CloudCredentials` (not `LocalCredentials`) when browser is launched. 
This is because we want the prompts to act as a guide for the user after they create api keys through the browser.

internal/commands/login/inputs.go
```diff
@@ -66,19 +71,23 @@ func (i *inputs) Resolve(profile *user.Profile, ui terminal.UI) error {
...
 	var questions []*survey.Question
 
-	if i.PublicAPIKey == "" {
-		if u.PublicAPIKey == "" {
+	if i.PublicAPIKey == "" || i.ForceBrowser {
```
makesitgo, line 74: spent a bit of time thinking about this and I could very well be missing some aspect of the desired logic here, but it _seems_ to me that `ForceBrowser` could simply be its own, static set of questions.

if we switched the `select` above to not be based purely off `i.AuthType` but rather just a generic one:

```go
	switch {
	case i.ForceBrowser:
		questions = []*survey.Question{
			&survey.Question{
				Name:   inputFieldPublicAPIKey,
				Prompt: &survey.Input{Message: "Public API Key", Default: u.PublicAPIKey},
			},
			&survey.Question{
				Name:   inputFieldPrivateAPIKey,
				Prompt: &survey.Password{Message: "Private API Key"},
			},
		}
	case i.AuthType == authTypeCloud:
		questions = i.resolveCloudCredentials(u)
	case i.AuthType == authTypeLocal:
		questions = i.resolveLocalCredentials(u)
	default: // ...
    }
```

then we don't need any extra wrinkles in the `resolveCloudCredentials` impl

internal/commands/login/inputs_test.go
```diff
@@ -178,6 +178,22 @@ func TestLoginInputs(t *testing.T) {
 				assert.Equal(t, "password", i.Password)
 			},
 		},
+		{
+			description: "Should open browser when flag is set even if a profile exists",
```
makesitgo, line 182:
```suggestion
			description: "should open browser when flag is set even if a profile exists",
```

makesitgo (commented):
first pass

internal/commands/login/inputs_test.go
```diff
@@ -237,4 +253,19 @@ func TestLoginInputs(t *testing.T) {
 		assert.Equal(t, "username", i.Username)
 		assert.Equal(t, "password", i.Password)
 	})
+
+	t.Run("should error if browser flag is enabled and credentials are input", func(t *testing.T) {
+		profile := mock.NewProfile(t)
+		_, ui := mock.NewUI()
```
makesitgo, line 259: probably don't even need the mock UI here (and can pass in `nil` to the `Resolve` func)

internal/commands/user/create.go
```diff
@@ -92,11 +92,11 @@ func (cmd *CommandCreate) Handler(profile *user.Profile, ui terminal.UI, clients
 	case userTypeAPIKey:
 		apiKey, err := clients.Realm.CreateAPIKey(app.GroupID, app.ID, cmd.inputs.APIKeyName)
 		if err != nil {
-			return fmt.Errorf("failed to create api key: %s", err)
+			return fmt.Errorf("failed to create API Keys: %s", err)
```
makesitgo, line 95: would keep this singular (as well as on L#99)

internal/commands/whoami/command.go
```diff
@@ -18,7 +18,7 @@ var CommandMeta = cli.CommandMeta{
 	// into a shared package would be helpful, to reduce coupling the command packages to each other
 	// (since this HelpText creates a "whoami depends on login" package cycle)
 	HelpText: `Displays a table that includes your Public and redacted Private Atlas
-programmatic API Key (e.g. ********-****-****-****-3ba985aa367a). No session
+programmatic API Keys (e.g. ********-****-****-****-3ba985aa367a). No session
```
makesitgo, line 21: would keep this singular (as well as L#44)

makesitgo (commented):
apologies for the mix-up on singular vs plural api keys!  that's on me.

last few things from me but afterwards this will be good to go

internal/commands/login/inputs.go
```diff
@@ -30,23 +30,39 @@ type inputs struct {
...
 	var questions []*survey.Question
 
-	switch i.AuthType {
-	case authTypeCloud:
+	switch {
+	case i.Browser:
+		questions = []*survey.Question{
+			{
+				Name:   inputFieldPublicAPIKey,
+				Prompt: &survey.Input{Message: "Public API Key"},
```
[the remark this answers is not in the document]
jandersongo: I intentionally left it out since we're assuming the purpose of the browser is to **create new keys**, in which case, we do not want to populate it with a default key from a profile.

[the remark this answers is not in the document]
makesitgo: I'd agree that's the purpose of the browser in the scenario where it organically happens, however if we're forcing the browser to launch and if the `--profile` set happens to already exist, providing this as a default value at least gives the user an idea of what they're overriding (should they use a new key).

i mostly asked this b/c i'm fairly certain that was the behavior you had coded before when we had a bunch of `|| i.Browser` conditions.  considering it's just a default value/display, i don't think it _really_ matters either way, the chances of the user being able to re-type the same private api key for the existing profile is **very slim**

i imagine regardless of the default value, we'll still be displaying the prompt/confirm to verify the user has an existing session and they need to confirm if they want to log that existing user out?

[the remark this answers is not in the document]
jandersongo: I'm afraid that if it's provided as a default, the user will be inclined to just press enter thus defeating the prompt for the new keys created in the browser.

side note:  this is what I had before with the bunch of `|| i.Browser` conditions 🙃 
```
var defaultInput string
if !i.ForceBrowser {
  defaultInput = u.PublicAPIKey
}
```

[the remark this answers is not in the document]
jandersongo: 🤝 
will add the confirmation.

makesitgo (approved):
🆒  beans

PR #310, opened 2022-06-10
BAAS-13288 add provider region to app create flow

kpatel71716:
This PR adds the provider_region field to the app create and push flows. Commands that are changed:

App Create changes: 

- app create supports the provider-region flag. Location is automatically determined if provider-region is passed in but location is not
- 
- we support passing in a location flag without provider region to support users using the command as is. Stopping support might warrant a major version bump

App init changes:

- The provider-region flag is part of the output if the provider region is not the default value of `aws-us-east-1`

App Push Changes:

- App Location is replaced with Cloud Provider and App Region selects. The App Region will filter results based on the selected cloud provider

- Deployment model is moved to above where App Location was. If a user selects GLOBAL as their deployment model, we will not show azure regions because azure is not supported for global apps. This also means that we skip the Cloud Provider select if deployment model is GLOBAL (not sure if it is considered a breaking change if we change the order of the selects)

### DIFF
--- a/internal/cloud/realm/realm.go
+++ b/internal/cloud/realm/realm.go
@@ -256,11 +256,12 @@ func isValidLocation(l Location) bool {
 
 // Set of supported cloud providers
 const (
-	CloudProviderAWS = "aws"
-	// TODOO CloudProviderGCP   = "gcp"
+	CloudProviderAWS   = "aws"
 	CloudProviderAzure = "azure"
+	CloudProviderGCP   = "gcp"
 )
 
+// TODO(BAAS-xxxx) add ticket number here if we want to do gcp separately
 var CloudProviderValues = []string{CloudProviderAWS, CloudProviderAzure}
 
 // The provider regions that we currently support
@@ -325,6 +326,7 @@ type ProviderRegion string
 // String returns the ProviderRegion display
 func (p ProviderRegion) String() string { return string(p) }
 
+// Label returns the ProviderRegion display without the cloud provider
 func (p ProviderRegion) Label() string {
 	return p.String()[strings.IndexByte(p.String(), '-')+1:]
 }
--- a/internal/cloud/realm/realm.go
+++ b/internal/cloud/realm/realm.go
@@ -59,7 +59,7 @@ func isValidConfigVersion(cv AppConfigVersion) bool {
 	return false
 }
 
-// set of known config version, location and deployment values
+// set of known config version, location, provider region and deployment values
 var (
 	ConfigVersionValues = []string{
 		AppConfigVersion20180301.String(),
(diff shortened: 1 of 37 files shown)


internal/cloud/realm/realm.go
```diff
@@ -79,6 +79,45 @@ var (
...
+	ProviderRegionValuesByCloudProvider = map[string][]string{
+		CloudProviderAWS: {
+			AWSProviderRegionUSEast1.Label(),
+			AWSProviderRegionUSWest2.Label(),
+			AWSProviderRegionEUCentral1.Label(),
+			AWSProviderRegionEUWest1.Label(),
+			AWSProviderRegionAPSoutheast1.Label(),
+			AWSProviderRegionAPSoutheast2.Label(),
+			AWSProviderRegionAPSouth1.Label(),
+			AWSProviderRegionUSEast2.Label(),
```
nickpoindexter, line 108: These are new and currently unsupported. However given the timelines, I think we should just add GCP in this mapping as well. I think by the time we release this, we should be ready in the newest regions.

internal/cloud/realm/realm.go
```diff
@@ -212,6 +254,142 @@ func isValidLocation(l Location) bool {
...
+
+// String returns the ProviderRegion display
+func (p ProviderRegion) String() string { return string(p) }
+
+func (p ProviderRegion) Label() string {
+	return p.String()[strings.IndexByte(p.String(), '-')+1:]
+}
+
+// Type returns the ProviderRegion type
+func (p ProviderRegion) Type() string { return flags.TypeString }
```
nickpoindexter, line 333: [nit] we should try to avoid having a mix of pointer and non pointer receiver methods.

internal/cloud/realm/realm.go
```diff
@@ -212,6 +254,142 @@ func isValidLocation(l Location) bool {
...
+		return errInvalidProviderRegion
+	}
+	*p = newProviderRegion
+	return nil
+}
+
+func isValidProviderRegion(p ProviderRegion) bool {
+	switch p {
+	case
+		ProviderRegionEmpty,
```
nickpoindexter, line 366: [q] we allow this because it's the same as using the default through the api

kpatel71716: Yeah, also for creating a 'minimal' command, if it isn't required, then I think the cli should support it like it does for location. No strong feelings either way

internal/cloud/realm/realm.go
kpatel71716: I agree, but for now I left a TODO I can fill out/remove once we sync up on when we want the cli released

internal/cloud/realm/realm.go
kpatel71716: I was thinking it's fine here in the same style of the other receivers because these aren't updated the value. But open to discuss

kpatel71716:
@makesitgo feel free to add anyone else to review if you would like, I can also provide more context if needed for this PR since a lot of files were touched

internal/cloud/realm/realm.go
makesitgo: I'll take the "blame" for the intent here, it was definitely my strategy (when it came to these "custom flag types") to only have the receivers be pointers when they _needed_ to, rather than worry about method set consistency nor the cost of a value-copy

basically, letting the last point of your linked article trump the rest of reasoning:

> For types such as basic types, slices, and small structs, a value receiver is very cheap so unless the semantics of the method requires a pointer, a value receiver is efficient and clear.

open to alternative approaches here as truthfully I don't think the code impact is significant one way or another.  these are likely more meaningful conversations around very complex structs and/or ones with fairly large method sets

nickpoindexter: It's fine the way it is. I spoke with Kush offline about the pointers and non pointer methods breaking if we made them consistent. I would say that last segment is about using pointer receivers in the case of complex structures regardless of "set/assignment" methods rather than mixing inconsistencies because the type is simple.

nickpoindexter (approved):
LGTM! But we should get Nick M or someone else with more CLI knowledge than me to sign off as well.

makesitgo:
hey there!

I would hold off on this merge for the time being (but I'd be remiss if I didn't mention these changes would be better served in a series of PRs/tickets)

from my understanding of how `location` works now, we really shouldn't be supporting it as an option for users anymore.  which goes to your point about dropping that flag altogether and making the breaking change to these 3 commands.

in that sense, we're already in the process of making `atlas-app-services-cli` which is going to start fresh (version-wise).  I'm going to reach out to drew now to hear his thoughts on how we want to proceed, but in the end it's a bit clunky/awkward to tell users: "hey, move over to this new major release" and then two weeks later say "hey, move over to this new package entirely"

kpatel71716:
Closing this PR since we aren't committing to this repository anymore. Will follow up determining how to best move these changes over

PR #312, opened 2022-06-27
BAAS 14547: Update request dependency in realm cli

auddin431:


auddin431:
Decided to use Axios here because it's the top rated HTTP request library in the developer community. We only used the `request` to download a file from S3 so in this context, Axios made the most sense. It's the rated to be the most lightweight and performant library, so perfect for this situation. Alternatively, Node's `http` package could've been used as well, but the code to make a simple GET request is far more verbose, and needed a lot more changes to get the download to work.

makesitgo (approved):
the code looks good to me... I'd just verify we're not breaking anything by making sure we can still run `./testInstall.js` successfully

if so, then:

🆒  beans

PR #323, opened 2022-09-30
BAAS-17135: Fix axios usage in installer

kmorkos:


install.js
```diff
@@ -73,50 +69,49 @@ function requstBinary(downloadURL, baseName) {
     const filePath = path.join(process.cwd(), binaryName);
     const outFile = fs.openSync(filePath, 'w');
 
-    const requestOptions = {
-      uri: downloadURL,
-      method: 'GET',
-    };
-    const client = request(requestOptions);
```
kmorkos, line 80: the old `request` library was still being used here, replaced this with `axios` as well

install.js
```diff
@@ -10,14 +10,10 @@ const manifestURL =
...
 function fetchManifest() {
-  return new Promise(function(resolve, reject) {
-    axios.get(manifestURL, function(err, _, body) {
-      if (err) {
-        reject(err);
-        return;
-      }
-      resolve(JSON.parse(body));
-    });
+  return new Promise((resolve, reject) => {
```
makesitgo, line 13: so `() => {}` style functions don't mesh well with particularly old versions of node (we've been burned by this before)... can we revert these types of changes (the intent still remains, just no "new style" function declarations)

makesitgo (commented):
just one tweak to support older node versions first

install.js
kmorkos: ahh good callout. will update 👍🏼

tsedgwick (approved):
LGTM

